PR #5, opened 2017-08-30
Amazeeio 857 bitbucket hook handler

dasrecht:


Schnitzel:
@twardnw 
need you help here, I updated the bitbucket tests to work with the new make build system, but unfortunately it doesn't work :(

especially I found a variable `git_repo_full_name` that has not been defined, I looked through all your code and also couldn't find it. I tried to replace it with `git_repo_name` (see https://github.com/amazeeio/lagoon/pull/5/commits/7fbef4853a23f9afc5ee144ca29edd0c7d05a572#diff-b6b345bc2a3fbebd5b6f2e9d5180cc4dR13) but then fails with 
```
Error: Cannot read property '1' of null
```
here: https://github.com/amazeeio/lagoon/blob/AMAZEEIO-857-bitbucket-hook-handler/services/webhook-handler/src/extractWebhookData.js#L70

twardnw:
@Schnitzel that variable used to be in the node.yaml playbook. Bitbucket does not advertise the git-ssh URI of the repo as part of the webhook payload, so the test has to be set up a bit different to succeed. e15e14e29d298ad58a1e371add5d93a2dd73040f removed node.yaml, but the new bitbucket tests weren't created appropriately.

Schnitzel:
@twardnw 
Okay! Thanks for the hint, I added the stuff back. 
Also I think we can do the regex stuff easier: https://github.com/amazeeio/lagoon/pull/5/commits/ac37a577793efc3460ed5514f0cc23dc57cd2b16
So basically what I changed: In the regex where we search for the domain we don't add the name of the repo, because we are only interested in the domain so we can just search for that?
But maybe I'm also missing a point here?

PR #12, opened 2017-08-30
Rename CLI config and add/improve error messages about missing config

karlhorky:
I renamed the CLI config so that it doesn't get confused with [the existing `.amazeeio.yml` files](https://github.com/amazeeio/lagoon/blob/master/.amazeeio.yml) (feedback please @Schnitzel @ryyppy ) and also because the config in the root was getting erroneously read by the CLI (the config resolution algorithm traverses up all parents).

In addition to that I've improved the error messaging for if the config file doesn't exist and added an error message for if the `sitegroup` configuration is null or undefined.

~~This is a pull request against #10 because it is based on those changes.~~ #10 was merged


Schnitzel (changes requested):
Mhh I don't think we can do that. Lagoon is a project hosted on amazeeio and therefore needs an `.amazeeio.yml` to tell the hosting system what to do. And we specifically want to have only one file, which contains all the information (stuff that the CLI needs, stuff that the deployment system needs).

So if we would add `.io.yml` files and launch that to production, lagoon would also end up having an `.io.yml` again :D (recursive hosting platforms FTW!)

Can we somehow else tell the CLI to be not confused by `.amazeeio.yml` of the git repo root folder?

karlhorky:
Sounds good, I'll revert that commit and find a different way to ignore this for running the CLI in development.

karlhorky:
@Schnitzel Ok, I've reverted the configuration file name change. The new error message that I added for missing sitegroup configuration actually already provides a bit of feedback for the case with the `.amazeeio.yml` in the root:

```bash
# This finds the .amazeeio.yml file in the root of lagoon, but it doesn't contain
# configuration for sitegroup, so it fails with an error.
➜   ../runCli.sh client
yarn run v0.27.5
$ node -r babel-core/register src/cli.js "client"
No sitegroup configured. Please create a .amazeeio.yml config file with "io init" or pass a sitegroup to this command via the --sitegroup option.
Online documentation: https://github.com/amazeeio/lagoon/blob/master/cli/README.md#io-init
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Maybe this is good enough for development.

Schnitzel (approved):

Schnitzel:
Looks good!

PR #18, opened 2017-08-31
Typo on README.md

joshmiller83:
> can take from seconds got maybe even hours

to

> can take from seconds to hours


Schnitzel:
thank you @joshmiller83 ! :)

PR #25, opened 2017-09-05
refactor local development and testing infrastructure

Schnitzel:


Schnitzel:
unfortunately CircleCI cannot build pull requests, so we're switching to Travis

Issue #31, opened 2017-09-13
statefulsets are checked that they successfully deployed


rtprio:
After the initial deployment of a stateful set, nothing is done on subsequentness deployments; we don't want to fail a build in this case.

PR #32, opened 2017-09-13
ignore services marked persistent when jenkins checks deploys

rtprio:
issue #31 

This code would be used for any statefulset and will remove items from the "verify checklist" if set like
```    amazeeio:
      type: custom
      template: services/rabbitmq/.amazeeio.app.yml
      persistent: true```


Schnitzel:
There has been quite some change as Jenkins does not check the deployments anymore, this probably needs a refactoring

dasrecht:
Do we still need this branch or can we remove it? @rtprio

rtprio:
Given the age, and movement in other branches, I don't think there's any value in this branch.

PR #43, opened 2017-09-22
Add support for Alpine Linux based PHP & Nginx images

skwashd:
* Add alpine images for php 5.6, 7.0 and 7.1. Reduces image size by ~100MB
* Support dynamic dockerfile for PHP images. Makes adding new versions easier
* Update tests to handle new images


images/php/fpm-alpine/Dockerfile
```diff
@@ -0,0 +1,40 @@
+ARG PHP_VERSION
+FROM php:${PHP_VERSION}-fpm-alpine
+
+MAINTAINER amazee.io
+
+COPY check_fcgi /usr/sbin/
+COPY fix-permissions /usr/local/bin
+COPY docker-entrypoint /amazeeio/entrypoints/70-php-entrypoint
```
Schnitzel: I think there is an ENTRYPOINT definition missing. 
On the centos7 we have a small script that is used for the entrypoint:
https://github.com/amazeeio/lagoon/blob/master/images/centos7/amazeeio-entrypoints.sh

and some default entrypoints: https://github.com/amazeeio/lagoon/tree/master/images/centos7/entrypoints that provide some basic functionality.

As we don't have a common upstream Dockerimage anymore, I guess we need to copy that into here as well?

Schnitzel:
Merged into Develop, thanks for your help @skwashd

PR #55, opened 2017-10-10
support drush 9/drupal 8.4. AMAZEEIO-927

rtprio:
support drush 9/drupal 8.4. AMAZEEIO-927

test for drush9 and adapt drupal root detection. also update yaml processtor to composer/yaml.


twardnw (approved):
Looks good @rtprio

PR #57, opened 2017-10-11
Api next develop

Schnitzel:


Schnitzel:
@ryyppy 
I realized that targeting the `master` branch for https://github.com/amazeeio/lagoon/pull/53 does not make sense, there are too many changes to `develop` so for right now it makes sense to target that branch.
I rebased `api-next` on top of `develop` and created this branch.

I also added an `api-data.sql` and `add api-data-watcher-pusher` that will automatically update the data in the `mariadb` DB whenever `api-data.sql` is updated (like we do it for hieradata).

`api-data.sql` contains all the data that the hiera also contains, except the information of environments and sql keys for environments, as we don't have the understanding of `environments` yet.

PR #105, opened 2017-11-20
76 - make it easier to inject nginx customizations

twardnw:


Schnitzel:
looks good, but let's finish the discussion about the naming :) https://github.com/amazeeio/lagoon/issues/76

Schnitzel (changes requested):
did you test that locally? because I'm worried that if 
somebody creates a file called `/etc/nginx/conf.d/server_prepend_bla.conf`
that this will be also included by the main `nginx.conf`:

https://github.com/amazeeio/lagoon/blob/95cb1781d058448db9897f231bcb683499c3c400/images/nginx/nginx.conf#L99

twardnw:
I did test a bit, but I guess the things I had in there could exist multiple times. I see two ways to solve that bit:
- specifically include the .conf files in the main nginx config (app, files, fastcgi)
- create a subdir to then glob against.

The directory nginx is trying to glob from here must exist, but doing `server_prepend*.conf` allows the specific file to not exist when it tries to inclue

twardnw:
Includes have been moved to a subdirectory that the Dockerfile adds to the base nginx to avoid globbing issues.

Schnitzel (approved):

PR #116, opened 2017-12-01
alpine varnish 5

twardnw:
Changes to add varnish alpine images and openshift template for deploying pod


tests/tests/drupal/check-deployed.yaml
```diff
@@ -2,7 +2,7 @@
   hosts: localhost
   serial: 1
   vars:
-    url: "http://nginx-php.{{ project | regex_replace('_', '-') }}.{{ branch | regex_replace('/', '-') }}.{{ lookup('env','OPENSHIFT_ROUTE_SUFFIX') }}"
```
Schnitzel: can we keep the nginx-php test and just test varnish additionally?

images/varnish/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY --from=commons /lagoon /lagoon
+COPY --from=commons /bin/fix-permissions /bin/ep /bin/docker-sleep /bin/
+COPY --from=commons /home/.bashrc /home/.bashrc
+
+# When Bash is invoked via `sh` it behaves like the old Bourne Shell and sources a file that is given in `ENV`
+# When Bash is invoked as non-interactive (like `bash -c command`) it sources a file that is given in `BASH_ENV`
+ENV TMPDIR=/tmp TMP=/tmp HOME=/home ENV=/home/.bashrc BASH_ENV=/home/.bashrc
+
+RUN echo "@edge http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories
+RUN apk update && apk add varnish@edge
```
Schnitzel: let's run these two `RUN` in one `RUN` (less image layers) and we can use `--no-cache`, see https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache

images/oc-build-deploy-dind/openshift-templates/varnish/template.yml
```diff
@@ -0,0 +1,147 @@
...
+    template:
+      metadata:
+        creationTimestamp: null
+        labels:
+          service: ${SERVICE_NAME}
+          branch: ${SAFE_BRANCH}
+          project: ${SAFE_PROJECT}
+      spec:
+        containers:
+        - image: ${REGISTRY}/${OPENSHIFT_PROJECT}/varnish:latest
```
Schnitzel: let's use `${SERVICE_NAME}` instead of `varnish`, see https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/node/template.yml#L58

(you probably copied from the `nginx-php` where we can't use that variable as we have two containers)

Schnitzel (changes requested):
Nice work! Some small tweaks :)

One other question that came into my mind:
How do we handle the varnish secret? Drupal will need to know it in order to connect to the varnish and purge. Maybe have a default one that can be overwritten via ENV Variable?

tests/tests/drupal/check-deployed.yaml
twardnw: I think having a default varnish secret and then overriding with EP and a secret in the project would work. My only worry is that a client forgets to have that secret set and ends up using the default secret in production, bit of a security risk there.

I'll get going on the other requested changes.

twardnw:
Have another look @Schnitzel , I think I've got all the changes requested in now.

images/oc-build-deploy-dind/openshift-templates/varnish/template.yml
```diff
@@ -0,0 +1,147 @@
...
+      metadata:
+        creationTimestamp: null
+        labels:
+          service: ${SERVICE_NAME}
+          branch: ${SAFE_BRANCH}
+          project: ${SAFE_PROJECT}
+      spec:
+        containers:
+        - image: ${REGISTRY}/${OPENSHIFT_PROJECT}/${SERVICE_NAME}:latest
+          name: varnish
```
Schnitzel: let's also change that one to `${SERVICE_NAME}`

images/oc-build-deploy-dind/openshift-templates/varnish/template.yml
```diff
@@ -0,0 +1,147 @@
...
+    test: false
+    triggers:
+    - type: ConfigChange
+    - imageChangeParams:
+        automatic: true
+        containerNames:
+        - varnish
+        from:
+          kind: ImageStreamTag
+          name: varnish:latest
```
Schnitzel: another candidate for `${SERVICE_NAME}`

images/oc-build-deploy-dind/openshift-templates/varnish/template.yml
```diff
@@ -0,0 +1,147 @@
...
+            requests:
+              cpu: 100m
+              memory: 100Mi
+    test: false
+    triggers:
+    - type: ConfigChange
+    - imageChangeParams:
+        automatic: true
+        containerNames:
+        - varnish
```
Schnitzel: another candidate for `${SERVICE_NAME}`

Schnitzel (changes requested):
nice! looks good, just some small adaptions on the yaml (should have been more clear that we should replace all `varnish` by `${SERVICE_NAME}` sorry).

regarding Varnish Secret: we need to find a way on how to inject that into the Drupal, but that could be part of another ticket.

Issue #70, opened 2017-10-25
X-Debug


Schnitzel:
- install xdebug php module by default into the php-fpm image, disabled by default and enable as soon as `XDEBUG_ENABLE` env variables are existing

Schnitzel:
done, how to use: https://github.com/amazeeio/drupal-example/blob/master/docker-compose.yml#L48

PR #120, opened 2017-12-05
#70 xdebug enable/disable functionality

Schnitzel:
- install xdebug php module
- disable it by default
- enable it on docker entrypoint via XDEBUG_ENABLE


twardnw (approved):
All looks good to me

Issue #107, opened 2017-11-26
Implement Galera Master-Master Replication


Schnitzel:

Schnitzel:
@rtprio 
can you write your current status in here? @desdury would like to help out :)

rtprio:
I was able to get a cluster up in a stateful set following the pattern of: https://github.com/adfinis-sygroup/openshift-mariadb-galera

that's in: https://github.com/amazeeio/lagoon/blob/maria-cluster/docker-images/oc-build-deploy/openshift-templates/mariadb-cluster/template.yml

It had the shortcomings of not being the default password, and based on centos. 
The cluster properly established, but I did not fully test the stack.

To resume this work:

- [ ] adapt config and stateful set options from adfinis-sygroup to our centos7-mariadb10-drupal
- [ ] refactor and reapply #32 .

PR #127, opened 2017-12-06
Extended OpenShift template for the Galera cluster

desdrury:
I've extended the OpenShift template so that it can do the following.

* Define the root password (see #107).  Stored in a `Secret`.  Uses the OpenShift template variable `MYSQL_ROOT_PASSWORD`.
* Define a default user and password.  Stored in the same `Secret`.  Uses the OpenShift template variables `MYSQL_USER` and `MYSQL_PASSWORD`.
* Define a default database.  Use the OpenShift template variable `MYSQL_DATABASE`.
* Added a `Deployment` and related `Service` that runs the Prometheus MySQL Exporter so that we can collect metrics to visualise in Grafana.

This has all been tested on a local MiniShift instance.

### DIFF
--- a/docker-images/oc-build-deploy/openshift-templates/mariadb-cluster/template.yml
+++ b/docker-images/oc-build-deploy/openshift-templates/mariadb-cluster/template.yml
@@ -25,7 +25,37 @@ parameters:
   - name: ROUTER_URL
     description: URL of the Router
     required: true
+  - name: MYSQL_USER
+    description: MariaDB user
+    required: false
+  - name: MYSQL_PASSWORD
+    description: Password for the MariaDB user
+    required: false
+  - name: MYSQL_DATABASE
+    description: Database to create
+    required: false
+  - name: MYSQL_ROOT_PASSWORD
+    description: Password for the root user
+    required: false
 objects:
+- apiVersion: v1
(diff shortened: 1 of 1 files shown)


Schnitzel:
Thanks!

We want to create an alpine mariadb first (https://github.com/amazeeio/lagoon/issues/61), so we don't depend on any external images (like `adfinissygroup/k8s-mariadb-galera-centos`).

After that happy to look into that :)

desdrury:
@Schnitzel do you have an ETA for when the Alpine MariaDB will be done?  And if not before Governors, i.e. 22/12/2017, would Amazee support us running a version of the Galera cluster using the `adfinissygroup/k8s-mariadb-galera-centos` as an interim solution?

Schnitzel:
pinging @rtprio for the question regarding MariaDB Images

rtprio:
mariadb builds ok with the galera flags; I'm having several issues with building the associated wsrep tools on alpine. I do not think I'll have it ready to go by the 22nd.

At this time `adfinissygroup/k8s-mariadb-galera-centos` would be my second choice.

rtprio:
I didn't mean to close this PR so swiftly but I'm tracking galera in #107

PR #128, opened 2017-12-07
109 better roles

ryyppy:
WIP


ryyppy:
Ready for review

Schnitzel:
Awesome! Two things I found

1. One SSH Key can be used in multiple clients - will try to implement the changes myself for that :)
2. are permissions based on projects implemented? I see 
```
${ifNotAdmin(cred.role, 'AND p.customer = :customerId')}
```
But how I understand is that this is based on the customer access, but not the project access?

ryyppy:
@Schnitzel Ah... right. That's why we iterate :-)
Shouldn't be too hard to implement that as well... could we also have some database fixtures for testing as well?

I manually added some table entries after `docker-compose up`, so I don't mess around with the integration test results.

Schnitzel:
@ryyppy 
did some refactoring:
- removing possibility to add `sshKeys` while `addCustomer`/`addProject` and added `addSshKeyToCustomer` `addSshKeyToProject` `removeSshKeyFromProject` `removeSshKeyFromCustomer` instead
- single SSH Key can be part of multiple customers, updated the permission checking in SQL Querries with `IN ()` instead of `=`
- Updated `permission` view to return list of customers and project
- Added Checks for project access via `OR`
- Added some more database fixtures

I'm actually not sure if something like https://github.com/amazeeio/lagoon/pull/128/files#diff-959ccd5c3237bb653a98fe7322129644R144
is really necessary, because `getOpenshiftByProjectId()` is only called after we check that the user has access to that ProjectID in `getAllProjects` or `getProjectByGitUrl` or `getProjectByName`. Of course it provides higher security, but I think for easyness of the code we could remove these checks?

ryyppy:
Missing:

- [x] reintroduce parsing mechanism for permissions (security reasons)
- [x] permission check for environment addOrUpdateEnvironment, getEnvironmentsByProjectId -> does user have permission for given project?
- [x] permission check for addProject & deleteProject -> does user have permission to customer?
- [x] Simplify async calls (get rid of `new Promise()`)

PR #132, opened 2017-12-10
Add env variables to set PHP server name and port.

thom8:
PHPStorm Zero configuration debugging requires a valid server name and port.

`$_SERVER['SERVER_NAME']`
`$_SERVER['SERVER_PORT']`

https://www.jetbrains.com/help/phpstorm/zero-configuration-debugging.html

This PR adds the ability to define a custom server name and port with environment variables for the `nginx-drupal` container --

`AMAZEEIO_PHP_SERVER_NAME`
`AMAZEEIO_PHP_SERVER_PORT`

Happy to update the defaults if required.

### DIFF
--- a/images/nginx-drupal/drupal.conf
+++ b/images/nginx-drupal/drupal.conf
@@ -55,6 +55,8 @@ server {
       include /etc/nginx/fastcgi.conf;
       fastcgi_param  SCRIPT_NAME        /index.php;
       fastcgi_param  SCRIPT_FILENAME    $realpath_root/index.php;
+      fastcgi_param  SERVER_NAME        ${AMAZEEIO_PHP_SERVER_NAME:-lagoon};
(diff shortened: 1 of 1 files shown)


Schnitzel:
mhh but they are set here 
https://github.com/thom8/lagoon/blob/fe521e825a553986f8f035fce5816b6572323ba2/images/nginx/fastcgi.conf#L27-L28

not sure why we would need to overwrite them again? In my tests I was able to run PHPStorm Debugging.

thom8:
Debugging is triggered however you need to manually create the server as `$server_name` doesn't seem to be defined anywhere and the `$server_port` defaults to the listening port.

By default the `$_SERVER['SERVER_NAME']` is empty and `$_SERVER['SERVER_PORT']` is `8080` so how would you configure this?

Schnitzel:
@thom8 
researched this and you where correct, `SERVER_NAME` was empty and phpstorm didn't like that.

I now did #167 which sets these correctly, let me know if it works now :)

thom8:
All working well, thanks :)

Issue #86, opened 2017-10-26
Development Environment Nginx Headers


Schnitzel:
Development Environments should have some special headers (x-robots, etc.)
we need a system that automatically injects them (probably via LUA?)

Schnitzel:
Infortunately there is no nginx:alpine with LUA, so I needed to do it without it, will check with maintainer if they are interested in a LUA version

twardnw:
https://github.com/nginxinc/docker-nginx/issues/213

Schnitzel:
so I found a way to implement this without the need of LUA.
But also in case we ever want LUA, we could use openresty, which is basically nginx bundled with lua (and a couple of other things, which is the reason that we don't wanna switch to it completely as it is probably too much)

Issue #77, opened 2017-10-26
Auto Generate Nginx config based on ENV variables


Schnitzel:
for common tasks (redirects, basic auth) it would be nice to tell the developers to define some ENV variables instead of writing full nginx configs.
These ENV variables are read by an entrypoint script and auto generates nginx configs based on them.

PR #148, opened 2017-12-28
#86 #77 switching to openresty:

Schnitzel:
- this gives us lua in nginx
- dynamically adding X-Robots-Tag header if we are on development
- allowing basic auth via BASIC_AUTH_USERNAME and BASIC_AUTH_PASSWORD


Schnitzel:
this is obsolete with #151 
if we ever need lua we can use openresty, but for now we rather implement it with plain nginx:alpine

PR #149, opened 2017-12-29
better api stability:

Schnitzel:
- readiness and liveness probe check /status
- error handler for sql errors (prevents uncaught exceptions and stuck of sqlclient)
- timeout of sqlclient as it would run forever if mysql is down

### DIFF
--- a/services/api/src/dao.js
+++ b/services/api/src/dao.js
@@ -69,6 +69,9 @@ const query = (sqlClient, sql) =>
       }
       res(rows);
     });
+    setTimeout(() => {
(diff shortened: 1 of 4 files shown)


services/api/src/dao.js
```diff
@@ -357,7 +354,7 @@ const getProjectByName = sqlClient => async (cred, args) => {
     `;
 
   const prep = prepare(sqlClient, str);
-
+  console.log("here")
```
ryyppy: remove this :-)

ryyppy (changes requested):

PR #158, opened 2018-01-01
140 mongo

twardnw:


Schnitzel:
Looks good, some changes:
- for the env variables we use a new feature of kubernetes 1.6 to import all keys from a configmap as env variables:
https://github.com/amazeeio/lagoon/blob/5da1856ba13eabd66df31d45a59c35cfdc4cafd3/images/oc-build-deploy-dind/openshift-templates/redis/template.yml#L73-L75
- is there a reason that we use the centos instead of alpine with mongo installed?

Schnitzel (changes requested):
see comment :)

twardnw:
Well, same story as Varnish, there was a mongodb package hiding in APK Edge, have that added now, also updated to use the env var map too.
 
@Schnitzel please review again? :)

images/mongo/docker-entrypoint
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+exec "$@"
```
Schnitzel: this is already defined in the main entrypoints.sh file:
https://github.com/amazeeio/lagoon/blob/master/images/commons/lagoon/entrypoints.sh#L15

so we don't need that file at all :)

images/mongo/yum.mongodb-org-3.4.repo
```diff
@@ -0,0 +1,6 @@
+[mongodb-org-3.4]
+name=MongoDB Repository
+baseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/3.4/x86_64/
+gpgcheck=1
+enabled=1
+gpgkey=https://www.mongodb.org/static/pgp/server-3.4.asc
```
Schnitzel: yum in alpine? :) 
guess we can remove that file.

images/oc-build-deploy-dind/openshift-templates/mongo/template.yml
```diff
@@ -0,0 +1,117 @@
...
+    description: git hash sha of the current deployment
+    required: true
+  - name: ROUTER_URL
+    description: URL of the Router
+    required: true
+objects:
+- apiVersion: v1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: ${SERVICE_NAME}-persistent
```
Schnitzel: we don't call the PVCs `-persistent ` anymore, as their name already suggests that they are persistent, just the `${SERVICE_NAME}`

images/oc-build-deploy-dind/openshift-templates/mongo/template.yml
```diff
@@ -0,0 +1,117 @@
...
+              tcpSocket:
+                port: 27017
+              initialDelaySeconds: 15
+              timeoutSeconds: 1
+          envFrom:
+          - configMapRef:
+             name: lagoon-env
+          volumeMounts:
+            - name: ${SERVICE_NAME}-persistent
+              mountPath: ${PERSISTENT_STORAGE_PATH}
```
Schnitzel: mongo has a predefined storage path, so it's not likely that a user wants to inject another path, let's define it hardcoded, see https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/mariadb/template.yml#L89 as an example

images/oc-build-deploy-dind/openshift-templates/mongo/template.yml
```diff
@@ -0,0 +1,117 @@
+apiVersion: v1
+kind: Template
+metadata:
+  creationTimestamp: null
+  name: amazeeio-openshift-template-mongo
+parameters:
```
Schnitzel: these seem not up to date, see https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/mariadb/template.yml#L6 as an example

Schnitzel (changes requested):

Schnitzel (approved):

PR #159, opened 2018-01-03
Fix missing npm dependency

ryyppy:


rtprio (approved):
That fixes the build for me.

Issue #210, opened 2018-01-30
xdebug dropped support for php 5.6


twardnw:
Modify the setup of the php 5.6 images to no longer include xdebug

twardnw:
will install old version, see #212

PR #211, opened 2018-01-30
#210 - do not install xdebug on php5.6

twardnw:
xdebug fully removed support for PHP 5 in their release yesterday, this change prevents xdebug from being installed on our php 5.6 images


twardnw:
well that works too 😛

PR #212, opened 2018-01-31
Install older version of xdebug for php 5.6

Schnitzel:


karlhorky:
Just tried this in [`103-refactor-cli-for-new-api`](https://github.com/amazeeio/lagoon/tree/103-refactor-cli-for-new-api) and it fixed the problem for me.

PR #231, opened 2018-02-08
fix for "mariadb image cannot be run as root"

rtprio:
Do not specify a user in docker-compose.
Add shadow package to provide usermod.
usermod mysql user into group 0, run container as mysql.


Schnitzel:
mhh do we really NEED the 
```
RUN usermod -G 0 --append mysql
```
?

can you maybe test without it? so we don't have to install another package?

Schnitzel (approved):
oh much nicer

PR #233, opened 2018-02-09
Add attribute filter for openshift token

ryyppy:
This prevents non-admin roles to access the `Openshift.token` attribute.


Schnitzel (approved):
nice!

PR #245, opened 2018-02-14
Fixes mariadb backups and backups for api-db on alpine

rtprio:
This also requires `MYSQL_USER`  and `MYSQL_PASSWORD` added to the lagoon-env configmap.

Is there any problem with that?


images/mariadb/Dockerfile
```diff
@@ -27,6 +27,7 @@ do mkdir -p $i; chown mysql $i;  /bin/fix-permissions $i; done
 
 COPY docker-entrypoint.sh /lagoon/entrypoints/90-mariadb-entrypoint
 COPY mysql-backup.sh /lagoon/
+RUN chmod 755 /lagoon/mysql-backup.sh
```
twardnw: I have found that just setting executeable and committing a script is enough to make it work in the container, eliminating another step from the Dockerfile.

images/mariadb/mysql-backup.sh
```diff
@@ -55,11 +55,11 @@ function delete_old_backups()
...
 
 function mysql_login() {
-  local mysql_login="-u $MYSQL_USER -h $MYSQL_HOST"
+  cmd="-u $MYSQL_USER -h $MYSQL_HOST"
   if [ -n "$MYSQL_PASSWORD" ]; then
-    local mysql_login+=" -p$MYSQL_PASSWORD"
+    cmd="$cmd -p$MYSQL_PASSWORD"
   fi
-  echo $mysql_login
+  echo $cmd
```
twardnw: might be safer to do this as `` $($cmd) `` or `` `$cmd` ``

twardnw: nvm, per slack convo it's not useful here 😄

Schnitzel (approved):
nice!

PR #239, opened 2018-02-13
153 add updates

ryyppy:
Implementation for #153

Current implementations of update mutations:
- [x] Project
- [x] Customer
- [x] SshKey (admin & normal user can update their own key)
- [x] Openshift (admin only)
- [x] NotificationSlack (admin only)
- [x] Environment (admin only)

PR #247, opened 2018-02-15
Add new graphql to get customer by name

igayoso:
Create new graphql query to get customer by name.

### DIFF
--- a/services/api/src/schema.js
+++ b/services/api/src/schema.js
@@ -93,6 +93,7 @@ const typeDefs = `
 
   type Query {
     projectByName(name: String!): Project
+    customerByName(name: String!): Project
     projectByGitUrl(gitUrl: String!): Project
     allProjects(createdAfter: String, gitUrl: String): [Project]
     allCustomers(createdAfter: String): [Customer]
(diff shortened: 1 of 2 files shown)


services/api/src/schema.js
```diff
@@ -93,6 +93,7 @@ const typeDefs = `
 
   type Query {
     projectByName(name: String!): Project
+    customerByName(name: String!): Project
```
Schnitzel: should be `Customer` no `Project`

Schnitzel (changes requested):
just one tiny thing

services/api/src/schema.js
igayoso: 👍

Schnitzel (approved):
lets see what the CI says :)

ryyppy:
I will merge this in our #239 PR

ryyppy:
Added to #239 .. see https://github.com/amazeeio/lagoon/pull/239/commits/c91c422f80f66bbdd1f0d728de17508b2c60ed64

This can be closed now!

PR #264, opened 2018-02-22
work in progress mariadb-galera

rtprio:


Schnitzel:
I did some refactoring: https://github.com/amazeeio/lagoon/pull/264/commits/8e1c560e0aba0858f0f0df724c86be2a4d0477d0

I think right now it is easer to just check based on the service-type name if we should check for a rollout or not. We can make this more clever in the future

rtprio:
The other thing I noticed is that the cluster fails to establish with a randomized password; It must be set in the environment.

Schnitzel:
merged, epic thanks to @rtprio <3

PR #268, opened 2018-02-27
do not install x-pack by default in `elasticsearch`

Schnitzel:
`elasticsearch` is the oss image by default, create a specific `platinum` image

also update to newest elasticsearch


twardnw (approved):
Thanks for explaining the backstory

Schnitzel:
keeping this on hold, as now elasticsearch has opensourced x-pack, need to check what we are gonna do here

Schnitzel:
x-pack will be part of the new elasticsearch, so we will check again when this comes out

PR #288, opened 2018-03-14
Update the Solr configuration to the latest version

twerthmueller:
We would like to use the ngram text data type form the current version of the search_api_solr module, but the current schema-versions from lagoon does not support ist.

Please update to the latest version of the configuration.


Schnitzel:
@twerthmueller 
thank you! merged :)

PR #299, opened 2018-03-23
add higher loglevel to rsh command

Schnitzel:


Schnitzel:
this should show why we have 
```
error: unable to upgrade connection: container not found ("cli")
```
sometimes

PR #308, opened 2018-03-26
WIP: skip deployment token

rtprio:
scans all commits in a github push for `skip deploy` or `deploy skip` , if so, logs and bails.


services/webhooks2tasks/src/handlers/githubPush.js
```diff
@@ -19,6 +19,15 @@ async function githubPush(webhook: WebhookRequestData, project: Project) {
     const branchName = body.ref.toLowerCase().replace('refs/heads/','')
     const sha = body.after
 
+    const skip_deploy = body.commits.map( commit => Array.isArray(commit.message.match(/skip deploy|deploy skip/i)))
+                                    .reduce( function(a,i,x,z) { return a||i })
+    if (skip_deploy) {
+      sendToLagoonLogs('info', project.name, uuid, `${webhooktype}:${event}:handled`, meta,
```
Schnitzel: can we change the message type from `${webhooktype}:${event}:handled` to `${webhooktype}:${event}:skipped`

services/webhooks2tasks/src/handlers/githubPush.js
```diff
@@ -19,6 +19,15 @@ async function githubPush(webhook: WebhookRequestData, project: Project) {
     const branchName = body.ref.toLowerCase().replace('refs/heads/','')
     const sha = body.after
 
+    const skip_deploy = body.commits.map( commit => Array.isArray(commit.message.match(/skip deploy|deploy skip/i)))
```
Schnitzel: let's add a short comment here what this actually does :)

services/webhooks2tasks/src/handlers/githubPush.js
Schnitzel: also if we do that, we have to update logs2slack that it sends a slack message for these type of events

Schnitzel (changes requested):
nice! Overall I'm not sure if just `deploy skip` is good, as Travis puts them into brackets: 
https://docs.travis-ci.com/user/customizing-the-build#Skipping-a-build

so I would rater use `[deploy skip]`

Schnitzel:
also this needs a test :)

services/webhooks2tasks/src/handlers/githubPush.js
```diff
@@ -19,6 +19,17 @@ async function githubPush(webhook: WebhookRequestData, project: Project) {
     const branchName = body.ref.toLowerCase().replace('refs/heads/','')
     const sha = body.after
 
+    // for each item in the body.commits array, search for our regular expression; this is an arroy of boolean
+    // which is .reduced into a single true if any are true.
+    const skip_deploy = body.commits.map( commit => Array.isArray(commit.message.match(/\[skip deploy\]|\[deploy skip\]/i)))
+                                    .reduce( function(a,i,x,z) { return a||i })
+    if (skip_deploy) {
+      sendToLagoonLogs('info', project.name, uuid, `${webhooktype}:${event}:skipped`, meta,
```
Schnitzel: this now needs an additional entry in logs2slack here: https://github.com/amazeeio/lagoon/blob/master/services/logs2slack/src/readFromRabbitMQ.js#L57
or we would never receive slack notifications for it

Schnitzel (changes requested):
one small thing

tests/tasks/git-skip-empty-commit-push.yaml
```diff
@@ -0,0 +1,32 @@
+- name: "{{ testname }} - make sure we are on the branch {{ branch }}"
+  shell: git checkout -b {{ branch }} || git checkout {{ branch }}
+  args:
+    chdir: /{{git_repo_name}}
+
+- name: "{{ testname }} - create empty commit in {{git_repo_name}}"
+  command: git commit -m "empty commit but [skip deploy]" --allow-empty
```
twardnw: I think it would be a more real test if we run a regular commit/deploy first, then test a 2nd commit which does [skip deploy]

services/webhooks2tasks/src/handlers/githubPush.js
rtprio: Those are in: 7e0f8fc57 d91a3af9c 318b3f02d

services/webhooks2tasks/src/handlers/githubPush.js
rtprio: The revised, simplified version should be more obvious.

PR #310, opened 2018-03-26
93 autoidle exclusions

rtprio:
Adds a field to the project table to ignore select projects from autoidle.


services/auto-idler/idle-clis.sh
```diff
@@ -31,67 +31,71 @@ echo "$ALL_ENVIRONMENTS" | jq -c '.data.developmentEnvironments[] | select((.env
   do
     PROJECT_NAME=$(echo "$project" | jq -r '.name')
     OPENSHIFT_URL=$(echo "$project" | jq -r '.openshift.console_url')
+    AUTOIDLE=$(echo "$project" | jq -r '.auto_idle')
```
Schnitzel, line 35: we would also need to add `auto_idle` to the graphQL request

services/auto-idler/idle-services.sh
```diff
@@ -54,37 +54,41 @@ echo "$DEVELOPMENT_ENVIRONMENTS" | jq -c '.data.developmentEnvironments[] | sele
   do
     PROJECT_NAME=$(echo "$project" | jq -r '.name')
     OPENSHIFT_URL=$(echo "$project" | jq -r '.openshift.console_url')
+    AUTOIDLE=$(echo "$project" | jq -r '.auto_idle')
```
Schnitzel, line 58: same here like in idle-clis.sh

Schnitzel (changes requested):

services/auto-idler/idle-clis.sh
rtprio: ✔️

services/auto-idler/idle-services.sh
rtprio: ✔️

rtprio:
Looks like that fixed the tests. 🎉

services/api/src/dao/project.js
```diff
@@ -145,7 +145,8 @@ const addProject = sqlClient => async (cred, input) => {
         },
         ${input.branches ? ':branches' : '"true"'},
         ${input.pullrequests ? ':pullrequests' : '"true"'},
-        ${input.production_environment ? ':production_environment' : 'NULL'}
+        ${input.production_environment ? ':production_environment' : 'NULL'},
+        ${input.auto_idle ? '"auto"':'"disabled"'}
```
Schnitzel: so whatever we send as a string in `auto_idle` it will never be added to the database?

Schnitzel (changes requested):

rtprio:
That wasn't how I was intending to do it, but blank is an acceptable value
for an enum (?).
Would an `int(1)` make more sense for this field?

services/api/src/dao/project.js
```diff
@@ -145,7 +145,8 @@ const addProject = sqlClient => async (cred, input) => {
         },
         ${input.branches ? ':branches' : '"true"'},
         ${input.pullrequests ? ':pullrequests' : '"true"'},
-        ${input.production_environment ? ':production_environment' : 'NULL'}
+        ${input.production_environment ? ':production_environment' : 'NULL'},
+        :auto_idle
```
Schnitzel: @rtprio this will assume that `auto_idle` is always set via GraphQL, which is not the case, it's for example not marked as mandatory during `ProjectInput` (https://github.com/amazeeio/lagoon/pull/310/files#diff-8f08ef36ccc2a2d39d5a3ac681238b23R128)
so we need something like:

```
${input.auto_idle ? ':auto_idle' : '1'},
```

services/api/src/schema.js
```diff
@@ -67,6 +67,7 @@ const typeDefs = `
     active_systems_remove: String
     branches: String
     production_environment: String
+    auto_idle: String
```
Schnitzel: `auto_idle` is defined as a `String` but inside the DB it's actually an `Int`, let's change that here.

Schnitzel (changes requested):
we're getting there!

services/api/src/schema.js
rtprio: I've changed this to `Int`.

Schnitzel (approved):

Schnitzel:
thank you @rtprio ! <3

PR #326, opened 2018-04-04
Fixed typo and changed usage for adaptation

bomoko:
Found an actual typo while reading the docs (accordint) and also changed "adaptions" to "adaptations" as the latter is the more common form (but not incorrect! Just a small edit).


bomoko:
I'm removing the one edit. Will make a new PR (sorry!)

Issue #335, opened 2018-04-09
Service type override check fails when there is no overrides


RoSk0:
```
Error: invalid path 'environments.master.types.cli', can't query subvalue 'types' of a leaf (leaf value is None)
```

This happens on deploy when followed "Lagoonize" steps from docs.

Schnitzel:
I believe this is fixed

PR #336, opened 2018-04-09
Issue #335: Fixed service type override check.

RoSk0:


Schnitzel:
I believe this is not necessary and was just caused by a missformed .lagoon.yml file

PR #350, opened 2018-04-18
refactor to use parallel with retries

rtprio:


Schnitzel:
tested on voaiv and this successfully worked!

PR #355, opened 2018-04-20
305 deploy log hashes

twardnw:
Swap out the hash built from project/branch info and just use a UUID instead


Schnitzel:
thanks @twardnw <3

PR #356, opened 2018-04-20
Create possibility to overwrite the backend port varnish talks to via…

dasrecht:
… the envrionment variable VARNISH_PORT

This would be needed for having varnish caches in front of decoupled sites. because they run on different ports.

### DIFF
--- a/images/varnish-drupal/drupal.vcl
+++ b/images/varnish-drupal/drupal.vcl
@@ -4,8 +4,8 @@ import std;
 
 # set backend default
 backend default {
-  .host = "${VARNISH_HOST:-nginx}";
-  .port = "${VARNISH_PORT:-8080}";
+  .host = "${VARNISH_BACKEND_HOST:-nginx}";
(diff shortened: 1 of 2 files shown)


Schnitzel:
mhh why? each pod on kubernetes has their full own port range available (the nginx and varnish pod for example both run on port 8080). The same is also through in local development

dasrecht:
Yes but when we want to put varnish in front of the decoupled containers varnish needs to connect to port 3000 instead of 8080. So we would need a way to overwrite the port somehow

Schnitzel:
ah, but then it should be called `VARNISH_BACKEND_PORT` no? because `VARNISH_PORT` suggests to be the port that Varnish listens on.

We probably also should rename `VARNISH_HOST` to `VARNISH_BACKEND_HOST`

dasrecht:
Fullack :) Commit incoming

Schnitzel:
thanks @dasrecht  <3

PR #366, opened 2018-04-25
Make http_resp_hdr_len and http_resp_size configurable

dasrecht:
Make http_resp_hdr_len and http_resp_size configurable for large header sizes. This can happen if we use varnish to cache sites that use a huge amount of cache tags

### DIFF
--- a/images/varnish/Dockerfile
+++ b/images/varnish/Dockerfile
@@ -30,8 +30,8 @@ EXPOSE 8080
 # tells the local development environment on which port we are running
 ENV LAGOON_LOCALDEV_HTTP_PORT=8080
 
-# Make http_resp_hdr_len and http_resp_size configurable
-ENV HTTP_RESP_HDR_LEN=${http_resp_hdr_len:-8k}
-ENV HTTP_RESP_SIZE=${http_resp_size:-32k}
+# Make http_resp_hdr_len and http_resp_size configurable (see in docker-entrypoint)
+ENV HTTP_RESP_HDR_LEN=8k
+ENV HTTP_RESP_SIZE=32k
 
 ENTRYPOINT ["/lagoon/entrypoints.sh"]
--- a/images/varnish/Dockerfile
+++ b/images/varnish/Dockerfile
@@ -19,8 +19,7 @@ RUN apk --no-cache add varnish
 RUN echo "${VARNISH_SECRET:-lagoon_default_secret}" >> /etc/varnish/secret
 
 COPY default.vcl /etc/varnish/default.vcl
-COPY varnish-start.bash /lagoon/varnish-start.bash
-RUN chmod 777 /lagoon/varnish-start.bash
+COPY varnish-start.sh /lagoon/varnish-start.sh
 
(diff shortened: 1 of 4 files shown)


images/varnish/Dockerfile
```diff
@@ -2,7 +2,6 @@ ARG IMAGE_REPO
 FROM ${IMAGE_REPO:-lagoon}/commons as commons
 FROM alpine
 
-MAINTAINER amazee.io
```
Schnitzel: why is this removed?

images/varnish/Dockerfile
```diff
@@ -31,5 +30,8 @@ EXPOSE 8080
 # tells the local development environment on which port we are running
 ENV LAGOON_LOCALDEV_HTTP_PORT=8080
 
+# Make http_resp_hdr_len and http_resp_size configurable
+ENV HTTP_RESP_HDR_LEN=${http_resp_hdr_len:-8k}
```
Schnitzel: this will not allow to overwrite the variables in runtime, but instead in Buildtime which I think it would be good to happen in Runtime

Schnitzel: or actually it does allow to overwrite the variables, but then we don't need the `${http_resp_hdr_len:-8k}` in there, and just use `HTTP_RESP_HDR_LEN=8k`

Schnitzel (changes requested):

images/varnish/Dockerfile
dasrecht: Because the MAINTAINER flags are deprected. We can add it via a label. Can create a seperate PR for that and go trough our images.


https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

dasrecht: i'll reverted that change and we can do that in a seperate pr.

images/varnish/Dockerfile
dasrecht: updated :)

images/varnish/docker-entrypoint
```diff
@@ -2,4 +2,4 @@
 
 ep /etc/varnish/*
 
-exec "$@"
+exec /usr/sbin/varnishd -a :8080 -T :6082 -F -f /etc/varnish/default.vcl -S /etc/varnish/secret -p http_resp_hdr_len=$HTTP_RESP_HDR_LEN -p http_resp_size=$HTTP_RESP_SIZE
```
Schnitzel, line 5: we can't have that inside an entrypoint, or we will never be able to start a container with `docker run --rm -it lagoon/varnish sh`

so we need to create it's own shell script that is then used in `CMD`

Schnitzel (changes requested):

images/varnish/docker-entrypoint
dasrecht: updated and fixed a few smaller things on the way there

dasrecht:
All checks are green :)

Schnitzel (approved):

PR #405, opened 2018-05-24
fail if notification is already added to the project, not if it does …

Schnitzel:
…not exist yet


ryyppy (approved):

Schnitzel:
actually looks like this does work already, but the code is a bit confusing, let's close that for now and we will look at it later

PR #426, opened 2018-05-31
Added rsyslog image and lagoon templates.

thom8:
Adds support for a `rsyslog` service which can be used by [monolog](https://www.drupal.org/project/monolog) to push Drupal logs.

Once monolog is enabled it you need to define a new handler eg.

```
parameters:
  monolog.channel_handlers:
    default: ['rsyslog']

services:
  monolog.handler.rsyslog:
    class: Monolog\Handler\SyslogUdpHandler
    arguments: ['rsyslog']
```

The service also needs to be added to the project:

```
  rsyslog:
    image: amazeeio/rsyslog
    labels:
      lagoon.type: rsyslog
```

Once in place logs will output to the rsyslog pod and piped to ELK.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,7 @@ images :=     centos7 \
 							varnish \
 							varnish-drupal \
 							redis \
+							rsyslog \
 							mongo \
 							elasticsearch \
 							kibana \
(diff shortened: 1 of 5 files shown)


Schnitzel:
we're gonna use a slightly different approach here:
- having a syslog receiver on each node which will forward the syslog to the elasticsearch

PR #440, opened 2018-06-07
Disable newrelic during builds and enable at runtime.

thom8:
There's any issue with having a pre envplated `newrelic.ini` when running build that require valid PHP config eg. `docker-php-ext-enable`.

This basically moves `newrelic.ini` to `newrelic.disable` then will move it back after the file has been envplated.


Schnitzel:
oh nice found. Actually we maybe should only do the move from `newrelic.disable` to `newrelic.ini` if `XDEBUG_ENABLE` env variable is set, with that we don't even have the newrelic extension loaded.
can you add that?

thom8:
updated :)

PR #441, opened 2018-06-07
Expose variable for PHP_MAX_INPUT_VARS.

thom8:
A common setting which needs to be increased particularly on Drupal site with large menus.


kurtfoster (approved):

thom8:
Relates to https://github.com/amazeeio/lagoon/issues/365

PR #446, opened 2018-06-11
Remote logging

rtprio:
* add http handler to elasticsearch
* add `logs-collector`  via secure_forward and nodeport


services/logs2logs-db/pipeline/router-logs.conf
```diff
@@ -5,6 +5,11 @@ input {
     queue_size => 5000
     receive_buffer_bytes => 26214400
   }
+  http {
+    id => "http_input"
+    user => "${LOGSTASH_USERNAME:-username}"
```
Schnitzel: logsstash configs work different for not defined variables, see:
https://www.elastic.co/guide/en/logstash/current/environment-variables.html
`${var:default value}`
but actually, let's use the same system like we use  for `ELASTICSEARCH_URL`:
1. define  `LOGSTASH_USERNAME` and `LOGSTASH_PASSWORD` in the Dockerfile with `logstash-username` and `logstash-password`
2. then we override these env variables in realtime via the kubernetes yaml

services/logs2logs-db/.lagoon.yml
```diff
@@ -82,6 +82,11 @@ objects:
                 secretKeyRef:
                   key: RABBITMQ_PASSWORD
                   name: rabbitmq-password
+            - name: LOGSTASH_PASSWORD
```
Schnitzel: that probably also needs `LOGSTASH_USERNAME`

Schnitzel (changes requested):

services/logs2logs-db/Dockerfile
```diff
@@ -14,5 +16,9 @@ ENV PATH_CONFIG=null
 ENV RABBITMQ_HOST="rabbitmq" \
     RABBITMQ_USER="guest" \
     RABBITMQ_PASSWORD="guest" \
-    ELASTICSEARCH_URL="http://logs-db:9200"
+    ELASTICSEARCH_URL="http://logs-db:9200" \
+    LOGSTASH_USERNAME="$LOGSTASH_USERNAME" \
```
Schnitzel: we want to define the default here, not reuse already existing variables, so:
`LOGSTASH_USERNAME="username"` and `LOGSTASH_PASSWORD="password"`

Schnitzel (changes requested):

Schnitzel (approved):

Issue #461, opened 2018-06-15
PERSISTENT_STORAGE_CLASS   is not defined in  elasticesearch PVM.yml


markxtji:
PERSISTENT_STORAGE_CLASS    is not defined in 
https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml

There should be a option for us to choose different type of storage

PR #463, opened 2018-06-15
this is going to fix lagoon issue #461

markxtji:
this is going to fix this issue:

https://github.com/amazeeio/lagoon/issues/461

### DIFF
--- a/images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
+++ b/images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
@@ -34,9 +34,12 @@ parameters:
   - name: DEPLOYMENT_STRATEGY
     description: Strategy of Deploymentconfig
     value: "Rolling"
+  - name: PERSISTENT_STORAGE_PATH
+    description: Full path to which should be mounted into persistent storage
+    required: true
   - name: PERSISTENT_STORAGE_CLASS
     description: Name of the Storage Class to use
-    value: "bulk"
+    value: ""
   - name: PERSISTENT_STORAGE_SIZE
     description: Size of the Storage to request
     value: "5Gi"
--- a/images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
+++ b/images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
@@ -34,6 +34,12 @@ parameters:
   - name: DEPLOYMENT_STRATEGY
     description: Strategy of Deploymentconfig
     value: "Rolling"
+  - name: PERSISTENT_STORAGE_CLASS
+    description: Name of the Storage Class to use
+    value: "slow"
+  - name: PERSISTENT_STORAGE_SIZE
+    description: Size of the Storage to request
+    value: "5Gi"
 objects:
 - apiVersion: v1
   kind: PersistentVolumeClaim
@@ -42,6 +48,8 @@ objects:
   spec:
     accessModes:
     - ReadWriteOnce
+    storageClassName: "${PERSISTENT_STORAGE_CLASS}"
     resources:
       requests:
-        storage: 5Gi
+        storage: ${PERSISTENT_STORAGE_SIZE}
+
--- a/images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
+++ b/images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
@@ -36,7 +36,7 @@ parameters:
     value: "Rolling"
   - name: PERSISTENT_STORAGE_CLASS
     description: Name of the Storage Class to use
-    value: "slow"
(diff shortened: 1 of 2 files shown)


images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
```diff
@@ -34,6 +34,12 @@ parameters:
   - name: DEPLOYMENT_STRATEGY
     description: Strategy of Deploymentconfig
     value: "Rolling"
+  - name: PERSISTENT_STORAGE_CLASS
+    description: Name of the Storage Class to use
+    value: "slow"
```
amazeeio-deploy: @markxtji `slow` is a custom class setup for DPC, you'll need to use `bulk`

amazeeio-deploy (dismissed):

markxtji:
@thom8 changed has been made according to what you suggested, build was successful but test failed on CI, which is quite hard for me to identify what exactly is the error.

Can you please advise what the possible error is  ? The change I made is just a very simple change.

thom8:
@markxtji can you try to replicate the issue locally? you'll have access to the build logs for debugging.

images/oc-build-deploy-dind/openshift-templates/elasticsearch/pvc.yml
```diff
@@ -34,6 +34,12 @@ parameters:
   - name: DEPLOYMENT_STRATEGY
     description: Strategy of Deploymentconfig
     value: "Rolling"
+  - name: PERSISTENT_STORAGE_CLASS
+    description: Name of the Storage Class to use
+    value: "bulk"
```
Schnitzel: let's use the default value `""` as `bulk` is usually some kind of NFS filesystem on our clusters which we don't want :)

Schnitzel (changes requested):
I don't think this will have the requested result, see:

https://github.com/amazeeio/lagoon/blob/e46f6816ddfc00be24ffe50683443cc145b061f7/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L381-L400

the `PERSISTENT_STORAGE_SIZE` and `PERSISTENT_STORAGE_CLASS` is only added when `PERSISTENT_STORAGE_PATH` is also set.

I'm happy though to move the check for  `PERSISTENT_STORAGE_SIZE` and `PERSISTENT_STORAGE_CLASS` out of the wrapping if in order to allow the usecase here.

markxtji:
Hi @Schnitzel ,

Thank you for the information.
I reference from below file you created for ngninx-php-persistent, added PERSISTENT_STORAGE_PATH and changed the value of PERSISTENT_STORAGE_CLASS to be empty.

https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/nginx-php-persistent/pvc.yml


When I use it from dockercompose file, I should use below lables, am I correct ?

    labels:
      lagoon.type: elasticsearch
      lagoon.persistent: /usr/share/elasticsearch/data
      lagoon.persistent.class: slow
      lagoon.name: elasticsearch

where can I find the mapping between the lables and parameters in the pvc.yml ?

Thank you.

Schnitzel:
@markxtji 
Hi!

So there are two systems in place here:
For the `ngninx-php-persistent` we actually want to give the developers the possibility to define the Path of the persistent storage, as this depends on the application they are deploying (like Drupal needs `/app/web/sites/default/files`, while Wordpress needs something like `/app/files`). So allowing the user to define the path is a good thing.

On the other side, for elasticsearch we DON'T want the developers to define the path, as elasticsearch expects the path of the persistent storage always to be at `/usr/share/elasticsearch/data`. If a developer defines another path (let's say `/foo/data`) then that persistent storage might be mounted, but elasticsearch will still put it's data at `/usr/share/elasticsearch/data` and with that create a lot of confusion.

So that's why I am suggesting to move the adding of the parameters here:
https://github.com/amazeeio/lagoon/blob/e46f6816ddfc00be24ffe50683443cc145b061f7/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L381-L400

outside of the `if [ ! $PERSISTENT_STORAGE_PATH == "false" ];` check in order to have the possibility to define just that in the docker-compose.yaml:
```
labels:
  lagoon.type: elasticsearch
  lagoon.persistent.class: slow
```

makes sense? Happy to do that part in `build-deploy-docker-compose.sh` if it's too confusing :)

> where can I find the mapping between the lables and parameters in the pvc.yml ?

that also happens here: https://github.com/amazeeio/lagoon/blob/e46f6816ddfc00be24ffe50683443cc145b061f7/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L381-L400

markxtji:
HI @Schnitzel  ,
thank you for the detail information.
I have updated according to what you suggested, please review again.
thank you.

Mark

Schnitzel (approved):

PR #490, opened 2018-07-10
#459 - update collation and charset

twardnw:


Schnitzel:
@twardnw can this just be changed on a live site? no migration of data or so necessary?

Schnitzel:
checked myself: existing tables will not be touched, it's just the default for new tables, so this should get in no problem

Schnitzel (approved):

Issue #514, opened 2018-07-18
Document how to update local docker images


dasrecht:
The magic command should be documented

> docker image ls --format "{{.Repository}}:{{.Tag}}" | grep amazeeio/ | grep -v none | xargs -n1 docker pull | cat

dasrecht:
This will be released here : https://github.com/amazeeio/pygmy/pull/21

Issue #477, opened 2018-06-26
Lagu: Add a self-update command


dan2k3k4:
I didn't notice any self-update command, not sure how that would work if people installed via `npm` versus `yarn`, perhaps it's easier to just run a version ping for the latest version and then show a notice message to the user if they're running an older version of `lagu`?

PR #515, opened 2018-07-18
Record lagoon version in containers

rtprio:
Partially assists with #477 insofar as that one can compare `/etc/lagoon-version` to [the latest release](https://github.com/amazeeio/lagoon/releases/latest), and run the script in #514 .


Schnitzel (changes requested):
@rtprio 
Instead of adding the `ARG` to ever image, could we just add it to the `commons` image within `/lagoon` as `/lagoon/version`?
this will then be copied by every images already via
https://github.com/amazeeio/lagoon/pull/515/files#diff-3c80f1e8503e7224690d4288bc69620dR10

Schnitzel:
mhh I realize this breaks now all the layer caching of all images.

can we change it to `git describe --abbrev=0 --tags` ? this will show something like:

```
~/git/amazeeio/lagoon (v0.16.0=)$ git describe --tags
v0.15.1-128-g55982fce
~/git/amazeeio/lagoon (v0.16.0=)$ git describe --abbrev=0 --tags
v0.15.1
```

rtprio:
I was trying to express if the image was a released version or some sort of development/hotfix/manually pushed version. 

`git describe --tags --exact-match 2>/dev/null || echo "development"  `
and that should keep caching working.

rtprio (commented):
I've revised this so that layer caching is improved.

PR #501, opened 2018-07-13
Add E2E Mutation tests for Lagu

karlhorky:
Build on the work done in #430 to write end to end tests for lagu that test the commands that execute mutations.

❗️Should probably merge #430 and then set the base branch of this PR to `master` before merging this one.

**TODO**

- [x] Create `lagu project delete` command
- [x] Tests for `lagu project create`
- [x] Tests for `lagu project delete`
- [x] Fix issue with truncated `lagu projects` output
    Ref: https://github.com/amazeeio/lagoon/blob/52a42dd55145b2e56c4b23e226bc8cd6c27f2f93/cli/src/__tests__/__snapshots__/e2e.test.js.snap#L95
    Ref: https://github.com/amazeeio/lagoon/blob/52a42dd55145b2e56c4b23e226bc8cd6c27f2f93/cli/src/__tests__/__snapshots__/e2e.test.js.snap#L136
    Ref: https://github.com/nodejs/node/issues/19218 
    Ref: https://github.com/sindresorhus/execa/issues/130)

Closes #498

PR #430, opened 2018-06-04
lagu: Add end to end tests

karlhorky:
**TODO**

- [x] Upgrade to latest Jest
- [x] Review other CLI tools to develop testing strategy
- [x] Adapt file structure and executable running to more standard approach
- [x] Choose executable spawning strategy
- [x] `lagu` - failing test (not enough options)
- [x] Write more config values from `lagu init` (`api`, `ssh`)
- [x] Remove reliance on global token for `login`, `logout` and other commands (option for token file path?)
- [x] Make `yarn execute` work again without building first
- [x] Change `cwd` to fixtures directory
- [x] Figure out a way to deal with changing `created` times in snapshots
- [x] Add VS Code extension settings and recommendations
- [x] Update CLI GraphQL schema
- General
	- [x] `lagu init`
	- [x] `lagu login`
        - [x] Configure local token path
        - [x] Test token
	- [x] `lagu logout`
- Queries
	- [x] `lagu customer`
	- [x] `lagu environment` - add data to GraphQL fixtures
	- [x] `lagu project`
	- [x] `lagu projects`

Bonus tasks:

- ~~Figure out how to automate on CI~~ Moved to #495
- ~~Fix binary not being on path for `yarn global add @lagoon/lagu` (`npm i -g @lagoon/lagu` works). Linking problem?~~ Moved to #496
- ~~Consolidate command data and configuration~~ Moved to #497
- ~~Mutations (`project create` / `project delete`)~~ Moved to #498

PR #520, opened 2018-07-23
lagu project create: Autocomplete for customers and openshifts

### LINKED ISSUE #421: lagu: typeahead in `project create`
Currently when creating a new project and we get presented with a long list of selections (like all clients I can create the project in) the way to select the clients only works with up and down arrow keys (see example screenshot below), would be awesome that if you type a client name that the selection list is automatically searched and filtered (check if the library we are using actually supports this).

screen shot 2018-07-16 at 17 09 00

karlhorky:
❗️Should probably merge #430 and #501 and then set the base branch of this PR to `master` before merging this one.

Closes #421


Schnitzel (approved):

Schnitzel:
merged into master!

PR #524, opened 2018-07-24
Cli helpers

twardnw:


images/php/cli/55-cli-helpers.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+dsql () {
+	drush sql-sync $1 default -d -v
```
Schnitzel: let's do them without `-d -v` I think that's too verbose, if somebody needs the verbose, they can still run `dsql -d -v`

images/php/cli/55-cli-helpers.sh
```diff
@@ -0,0 +1,9 @@
...
+}
+
+dfiles () {
+	drush rsync $1:%files default:%files -d -v
```
Schnitzel: let's do them without `-d -v` I think that's too verbose, if somebody needs the verbose, they can still run `dfiles -d -v`

Schnitzel (changes requested):

PR #525, opened 2018-07-24
if running in kubernetes, try to use ssh at  `ssh.lagoon.svc:2020`

Schnitzel:
this should solve the weird SSH MTU issues we're having


Schnitzel:
we decided we're not gonna do that anymore, it's too much risk that it connects to the wrong ssh endpoint

PR #538, opened 2018-08-02
492 &  548 route storage & monitoring urls

twardnw:
This PR adds storage to `environment`s for `lagoon_route` (primary route to an environment), and `lagoon_routes` (all routes for an environment). The `openshiftbuilddeploymonitor` will add these routes upon completion of a successful deploy, and update them with subsequent deploys.


services/openshiftbuilddeploymonitor/src/index.js
```diff
@@ -187,10 +190,21 @@ const messageConsumer = async msg => {
...
-
       sendToLagoonLogs('info', projectName, "", `task:builddeploy-openshift:${buildPhase}`, meta,
         `*[${projectName}]* ${logMessage} Build \`${buildName}\` complete. ${logLink} \n ${routes.join("\n")}`
       )
+      try {
+        const updateEnvironmentResult = await updateEnvironment(
+          branchName,
+          environmentOpenshift.id,
+          `{
+            lagoon_route: "${routes[0]}",
```
Schnitzel: I don't think this will be always the Route that is saved within the ConfigMap as `LAGOON_ROUTE`. This just takes the first route that is found within the Routes object of Kubernetes.
I think we should actually load the configmap and the entries in `LAGOON_ROUTE` and `LAGOON_ROUTES` to fill the data in the API

services/openshiftbuilddeploymonitor/src/index.js
```diff
@@ -44,10 +46,13 @@ const messageConsumer = async msg => {
   } = JSON.parse(msg.content.toString())
 
   logger.verbose(`Received BuildDeployOpenshift monitoring task for project: ${projectName}, buildName: ${buildName}, openshiftProject: ${openshiftProject}, branch: ${branchName}, sha: ${sha}`);
-
-  const result = await getOpenShiftInfoForProject(projectName);
+  let result
+  result = await getOpenShiftInfoForProject(projectName);
   const projectOpenshift = result.project
 
+  result = await getEnvironmentByName(`${branchName}`, projectOpenshift.id)
```
Schnitzel: Let's use another variable than `result` as the variable is already used just above, a little bit safer. Also then we can actually use `const` for both variables. Again safer :)

services/api-db/docker-entrypoint-initdb.d/00-setup.sql
```diff
@@ -71,6 +71,8 @@ CREATE TABLE IF NOT EXISTS environment (
        deploy_type            ENUM('branch', 'pullrequest', 'promote') NOT NULL,
        environment_type       ENUM('production', 'development') NOT NULL,
        openshift_projectname  varchar(100),
+       lagoon_route           varchar(300),
```
Schnitzel: If we are adding new columns in the table, we also need an update procedure for existing tables, see
https://github.com/amazeeio/lagoon/blob/f02a574502307efdcc0db9e1ee7a88f595ec6fa2/services/api-db/docker-entrypoint-initdb.d/00-setup.sql#L161-L179
and
https://github.com/amazeeio/lagoon/blob/f02a574502307efdcc0db9e1ee7a88f595ec6fa2/services/api-db/docker-entrypoint-initdb.d/00-setup.sql#L415
as an example

Schnitzel (changes requested):

Schnitzel (dismissed):

PR #539, opened 2018-08-03
Bitbucket pullrequest support

rtprio:
WIP


Schnitzel:
this is probably missing some code like this:
https://github.com/amazeeio/lagoon/blob/master/services/logs2slack/src/readFromRabbitMQ.js#L48-L61
https://github.com/amazeeio/lagoon/blob/master/services/logs2rocketchat/src/readFromRabbitMQ.js#L49-L59

Schnitzel (changes requested):
logs2slack and logs2rocketchat also need some adaptions :)

Schnitzel (approved):

PR #544, opened 2018-08-07
Allow environment overrides for templates and rollout type.

thom8:
Allows `OVERRIDE_TEMPLATE` & `SERVICE_ROLLOUT_TYPE` to be overridden by environment in `.lagoon.yml`.

Might be easier to define an array of `lagoon.__` that will override labels defined in `docker-compose.yml`


Schnitzel:
closing in favor of https://github.com/amazeeio/lagoon/pull/559

Issue #537, opened 2018-08-02
solr pods do not shut down properly


twardnw:
Shutdown of solr images is not handled well during deployment, leading to files in the persistent storage which appear to the new pod that they are in use. Initial investigation leads this to look like the solr process is not getting the TERM signal.

dasrecht:
Solr starts with following error when we hit that issue:

```
java.util.concurrent.ExecutionException: org.apache.solr.common.SolrException: Unable to create core [drupal]
  | at java.util.concurrent.FutureTask.report(FutureTask.java:122)
  | at java.util.concurrent.FutureTask.get(FutureTask.java:192)
  | at org.apache.solr.core.CoreContainer.lambda$load$6(CoreContainer.java:594)
  | at com.codahale.metrics.InstrumentedExecutorService$InstrumentedRunnable.run(InstrumentedExecutorService.java:176)
  | at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
  | at java.util.concurrent.FutureTask.run(FutureTask.java:266)
  | at org.apache.solr.common.util.ExecutorUtil$MDCAwareThreadPoolExecutor.lambda$execute$0(ExecutorUtil.java:229)
  | at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
  | at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
  | at java.lang.Thread.run(Thread.java:748)
  | Caused by: org.apache.solr.common.SolrException: Unable to create core [drupal]
  | at org.apache.solr.core.CoreContainer.createFromDescriptor(CoreContainer.java:966)
  | at org.apache.solr.core.CoreContainer.lambda$load$5(CoreContainer.java:565)
  | at com.codahale.metrics.InstrumentedExecutorService$InstrumentedCallable.call(InstrumentedExecutorService.java:197)
  | ... 5 more
  | Caused by: org.apache.solr.common.SolrException: Index dir '/opt/solr/server/solr/mycores/drupal/data/index/' of core 'drupal' is already locked. The most likely cause is another Solr server (or another solr core in this server) also configured to use this directory; other possible causes may be specific to lockType: native
  | at org.apache.solr.core.SolrCore.<init>(SolrCore.java:977)
  | at org.apache.solr.core.SolrCore.<init>(SolrCore.java:830)
[...]

PR #552, opened 2018-08-09
#537 - change solr to no lock

twardnw:


Schnitzel:
nice! is this something we need to change on the existing solr containers? of will just a new image fix that also in existing ones?

Schnitzel:
Isn't the config copied during the entrypoint into a persistent config folder and from that point on always loaded from there and not from what is shipped within the image?

Schnitzel:
So I think we can have two approaches here:
A: either we change the solr system that it always uses the config that is provided from the Image and do not actually copy the config (my preference, as all other images also work like this)
B: We implement an entrypoint which can update an already copied config with some changes (like an update system, but this could lead to errors and other weird stuff)

images/solr/00-solr-datadir.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# old volume looks like this
+# $corename/data/-files-
+# new volume looks like this
+# $corename/-files-
+
+for datadir in /var/solr/* ; do
+  if [ -d ${datadir}/data ]; then
+    echo "${datadir}/data is in old format, copying to ${datadir}."
+    cp -Rpv ${datadir}/data/* ${datadir}/
```
Schnitzel: why copy? can't we move it? the chance that something happens is super small, no?

images/solr/00-solr-datadir.sh
```diff
@@ -0,0 +1,35 @@
...
+  fi
+
+  echo "if this core is working properly, you may delete the following files:"
```
Schnitzel: this echo is outputed all the time, even if there are no files do delete, let's refactor this that it's only echoed if actually something is to delete

Schnitzel (changes requested):

images/solr/00-solr-datadir.sh
rtprio: Yes, it's just paranoia. I'll remove it.

Schnitzel (changes requested):

docs/using_lagoon/drupal/services/solr.md
```diff
@@ -22,15 +22,25 @@ FROM amazeeio/solr:6.6
...
-To implement multiple cores, you will also need to ship your own solr schema as above, the only change needed is to the `CMD` of the Dockerfile, repeat the pattern of `precreate corename /solr-conf/ ;` for each core you require.
+To implement multiple cores, you will also need to ship your own solr schema as above, the only change needed is to the `CMD` of the Dockerfile, repeat the pattern of `precreate-core corename /solr-conf/ ;` for each core you require.
 
 ```
-CMD ["sh", "-c", "precreate-core drupal /solr-conf/ ; precreate-core core1 /solr-conf/ ; precreate-core core2 /solr-conf/ ; precreate-core core3 /solr-conf/ ; solr start -f"]
+
+FROM amazeeio/solr:6.6-drupal
+
+RUN precreate-core drupal-index1 /solr-conf
+RUN precreate-core drupal-index2 /solr-conf
```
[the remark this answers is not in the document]
rtprio: I was going for readability over functionality here as it is documentation. 
But I can update.

local-dev/git/Dockerfile
```diff
@@ -25,7 +25,8 @@ RUN mkdir -m 700 /git/.ssh && \
 	git --bare init /git/nginx.git && \
 	git --bare init /git/features.git && \
 	git --bare init /git/features-subfolder.git && \
-    git --bare init /git/elasticsearch.git
+  git --bare init /git/elasticsearch.git && \
+  git --bare init /git/solr.git
```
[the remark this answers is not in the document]
rtprio: I started; but then I realized I wasn't testing the feature that I was developing, only that solr came up and could index and search.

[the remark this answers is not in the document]
Schnitzel: better than nothing :) how far are you? happy to get this change in and add the tests later

Makefile
```diff
@@ -52,7 +52,7 @@ SHELL := /bin/bash
 #######
 
 # Parameter for all `docker build` commands, can be overwritten with `DOCKER_BUILD_PARAMS=` in cli
-DOCKER_BUILD_PARAMS := --quiet
```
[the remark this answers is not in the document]
rtprio: yes, a mismerge of fbf19896b88fda2f.

Schnitzel (approved):

karlhorky:
The API data population in the [`local-dev/api-data/api-data.gql`](https://github.com/amazeeio/lagoon/pull/552/files#diff-153edce4d54b9b3a50e909cdce8058d8) file in this PR will have conflicts with the changes in [`develop`](https://github.com/amazeeio/lagoon/tree/develop) branch (see the `develop` PR in #675). It's mostly file structure changes (different filename) so it shouldn't be too tough to resolve.

images/solr/00-solr-datadir.sh
```diff
@@ -0,0 +1,24 @@
...
+  fi
+
+  mkdir -p /opt/solr/server/solr/${corename}
+  cp -R ${datadir}/conf /opt/solr/server/solr/${corename}/
+  touch /opt/solr/server/solr/${corename}/core.properties
+  sed -ibak 's/<dataDir>.*/<dataDir>\/var\/solr\/${solr.core.name}<\/dataDir>/' /opt/solr/server/solr/${corename}/conf/solrconfig.xml
+
+done
+
+exec solr-foreground
```
Schnitzel: this will cause all other lagoon entrypoints to not run anymore, I don't think that's expected

images/solr/00-solr-datadir.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+set -ex
+
+# old volume looks like this
+# $corename/data/-files-
+# new volume looks like this
+# $corename/-files-
+
+for datadir in `ls -d /var/solr/*` ; do
```
Schnitzel: let's use `$()` instead of backticks, as backticks are not suggested by bash: https://stackoverflow.com/a/33301370

images/solr/00-solr-datadir.sh
```diff
@@ -0,0 +1,24 @@
...
+  corename=$(basename $datadir)
+  if [ -d ${datadir}/data ]; then
+    echo "${datadir}/data is in old format, moving to ${datadir}."
+    mv ${datadir}/data/* ${datadir}/
+    rm -Rf ${datadir}/data
+  fi
+
+  mkdir -p /opt/solr/server/solr/${corename}
```
Schnitzel: please comment what exactly these lines are doing and why

Schnitzel (changes requested):

images/solr/00-solr-datadir.sh
rtprio: updated: 63a7061d8bc9792c21f59e300dda3803f9fab822

rtprio: fixed in 63a7061d8bc9792c

images/solr/00-solr-datadir.sh
rtprio: updated in 31b0190b927d29

Issue #429, opened 2018-06-04
Allow a set limit of environments per project.


thom8:
Would be nice to be able to set a hard limit of new environments per project so older ones can be tidied up before new ones are created.

This could be a validation check before launching a new project and report an error relating to this limit.

PR #556, opened 2018-08-10
add option to configure environment limits per project

rtprio:
will address #429 

(still work in progress; promotion codepath not yet considered)


services/api/src/dao/project.js
```diff
@@ -179,7 +179,8 @@ const addProject = ({ sqlClient }) => async (cred, input) => {
         ${input.pullrequests ? ':pullrequests' : '"true"'},
         ${input.production_environment ? ':production_environment' : 'NULL'},
         ${input.auto_idle ? ':auto_idle' : '1'},
-        ${input.storage_calc ? ':storage_calc' : '1'}
+        ${input.storage_calc ? ':storage_calc' : '1'},
+        ${input.environment_limit ? ':environment_limit' : '2'}
```
Schnitzel: this `2` means that the default is a limit of two environments?
I don't think that's a good default, as we probably will run into issues with that.
Let's do the default at 5  :)

Schnitzel:
@rtprio any chance that you can look at the merge conflicts? Thanks!

Schnitzel:
looks like something is failing with the containers locally, maybe run a test locally to see what exactly is failing.

Schnitzel (changes requested):
sorry to jump into this quite late, but could we call it `developmentEnvironmentsLimit`? as the production environments do not count against the limits I want to make it very clear when people are using the API

services/api/src/dao/project.js
rtprio: default to 5 now.

rtprio:
This _should_ be ready to merge.

Issue #565, opened 2018-08-14
Error 8 While running the lagoon kickstart


amjad1233:
**Describe the bug**
 I was installing lagoon kick-start as per documentation to my Openshift 
http://lagoon.readthedocs.io/en/latest/administering_lagoon/install/

**To Reproduce**
1. in the git clone after adding token and OC Console URL
2. run `make lagoon-kickstart`
3. On step 6/7 it stops with following error
The command '/bin/sh -c apk update     && apk upgrade     && apk add --no-cache         libstdc++     && apk add --no-cache         binutils-gold         curl         g++         gcc         gnupg         libgcc         linux-headers         make         git         file         openssl         python         bash         ca-certificates         wget         libpng-dev     && wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub     && wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.26-r0/glibc-2.26-r0.apk     && apk add glibc-2.26-r0.apk     && rm -rf /var/cache/apk/*' returned a non-zero code: 8
make: *** [build/node__8-builder] Error 8

**Expected behavior**
The installation should finish

**Additional context**
```make lagoon-kickstart
docker build --quiet --build-arg IMAGE_REPO=lagoon --build-arg NODE_VERSION=8 -t lagoon/node:8-builder -f images/node/builder/Dockerfile images/node/builder
Sending build context to Docker daemon   2.56kB
Step 1/7 : ARG NODE_VERSION
Step 2/7 : ARG IMAGE_REPO
Step 3/7 : FROM ${IMAGE_REPO:-lagoon}/node:${NODE_VERSION}
 ---> b81e3b914b56
Step 4/7 : MAINTAINER amazee.io
 ---> Using cache
 ---> a48511cbfd7b
Step 5/7 : ENV LAGOON=node
 ---> Using cache
 ---> d3a779b958f2
[...]

PR #567, opened 2018-08-14
Fix paths to sgerrand.rsa.pub

rocketeerbkw:
The path to a public key used for installing software has changed causing builds of some docker images to fail.

See https://github.com/sgerrand/alpine-pkg-glibc#please-note


rocketeerbkw:
Looks like this will address #565

PR #589, opened 2018-08-24
if the deploymentconfig doesn't exist, don't even try run the command

rtprio:


Schnitzel:
I don't understand the need for this, as there could also be services that run `statefulset` or `deamonsets` and we would like to run commands inside of them?

rtprio:
My logic here was that it fails if the cli pod doesn't exist on pre-deploy;
I'll need to split out the scripts to allow pre-deploy to fail on the first deployment.

Schnitzel:
@rtprio ah good point, yes I would change the scripts that for `pre-deploy` tasks we don't abort when there is no running pod.

PR #592, opened 2018-08-30
Begin building lagoon UI

kristymoon:


services/ui/src/components/ProjectTeaser/index.js
```diff
@@ -0,0 +1,13 @@
...
+export default ({ project }) => (
+  <tr>
+    <td>
+      <Link href={{ pathname: '/project', query: { name: project.name } }}>
+        <a>{project.name}</a>
+      </Link>
+    </td>
+    <td>{project.customer.name}</td>
+  </tr>
+);
```
rocketeerbkw: This doesn't feel like a reusable component to me. Will project teasers always be in a table? It's not very complex, could just as easily be inlined in pages/index.js

services/ui/src/components/Breadcrumbs/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Link from 'next/link';
+
+export default ({project, environment}) => {
+  const projectURL = project ? `/project/${project.replace(/\s+/g, '-').toLowerCase()}` : ';'
+  const environmentURL = environment ? `/project/${project}/${environment}`.replace(/\s+/g, '-').toLowerCase() : '';
```
rocketeerbkw: Query strings are fine for now. Can get rid of these constants since they won't be used.

services/ui/src/components/EnvironmentTeaser/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Link from 'next/link';
+
+export default ({ environment, project }) => {
+  const productionLabel = environment.environmentType == 'production' ? <div>Prod</div> : '';
+  const environmentURL = `/project/${project}/${environment}`.replace(/\s+/g, '-').toLowerCase();
```
rocketeerbkw: Can be removed.

services/ui/src/components/EnvironmentTeaser/index.js
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <div key={environment.id}>
```
rocketeerbkw: `key` is only necessary when the component is inside a loop or you want to force a re-render. I think this situation is neither and can be removed.

services/ui/src/pages/projects.js
```diff
@@ -0,0 +1,41 @@
...
+      if (error) return <p>Error :(</p>;
+
+      const projects = data.allProjects.map(project => <ProjectOld key={project.id} project={project} />);
+
+      return <div>
+        {projects}
+      </div>
+    }}
+  </Query>
+</>
```
rocketeerbkw: This whole page is replaced by stuff in `pages/index.js` and can be deleted. `../components/ProjectOld` can also go away, it'll be in git if it's ever useful in the future

services/ui/src/components/ProjectItem/index.js
```diff
@@ -0,0 +1,31 @@
...
+    <tr>
+      <td>Environments</td>
+      <td>
+        {!project.environments.length && `No Environments`}
+        {project.environments.map(environment => <Environment key={environment.id} environment={environment} />)}
+      </td>
+    </tr>
+    </tbody>
+  </table>
+);
```
rocketeerbkw: It doesn't look like this component is used anywhere, safe to delete?

services/ui/src/components/Breadcrumbs/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Link from 'next/link';
+
+export default ({project, environment}) => {
```
rocketeerbkw: The `project` prop here is actually only the project name. In other places where `project` is used as a prop, an object is passed. The `environment` variable here is also an object. Changing this `projectName` I think would be better for understanding what's actually needed. (And of course, don't forget to change the prop names where this component is used).

services/ui/src/pages/environment.js
```diff
@@ -0,0 +1,49 @@
...
+        if (error) return `Error!: ${error}`;
+        const environment = data.environmentByOpenshiftProjectName;
+        return (
+          <div>
+            <div>
+              <Header />
+              <Breadcrumbs project={environment.project.name} environment={environment}/>
+            </div>
+            <NavTabs activeTab='overview' environment={environment.openshiftProjectName}/>
+            <EnvironmentData key={environment.id} environment={environment} />
```
rocketeerbkw: I wouldn't expect `key` to be necessary here either?

services/ui/src/pages/environment.js
```diff
@@ -0,0 +1,49 @@
...
+    <Query query={query} variables={{openshiftProjectName: props.router.query.name}}>
+      {({ loading, error, data }) => {
+        if (loading) return null;
+        if (error) return `Error!: ${error}`;
+        const environment = data.environmentByOpenshiftProjectName;
+        return (
+          <div>
+            <div>
+              <Header />
+              <Breadcrumbs project={environment.project.name} environment={environment}/>
```
rocketeerbkw: Not now, but when you're styling, you might consider a layout component instead of injecting `Header` `Breadcrumbs` and at some point probably a `Footer` into every page. Next.js has an example in their docs https://github.com/zeit/next.js/wiki/Global-styles-and-layouts

PR #613, opened 2018-09-12
Disable debug in build script.

thom8:
https://github.com/amazeeio/lagoon/pull/600 introduced an issue where the SSH key is output in build logs. Disabling debug flag will hide the `export` command.


Schnitzel:
this will be fixed with https://github.com/amazeeio/lagoon/pull/614

PR #630, opened 2018-09-18
Mariadb alpine 10.2 docker images

seanhamlin:
This change brings in:

* New docker images (based on Alpine Linux latest) (`mariadb` and `mariadb-drupal`). This image has MariaDB 10.2.
* Moving the Galera images to their own namespace (`mariadb-galera` and `mariadb-galera-drupal`). This might be a controversial change.

We might need a beta tag to ensure that all projects are not going to change underneath them.


Schnitzel:
superseeded by https://github.com/amazeeio/lagoon/pull/635

PR #631, opened 2018-09-19
Feature/gov 29 startpage

kristymoon:


Schnitzel:
@kristymoon 
awesome work! Can we also make the search to search through the URL of the projects?

Schnitzel:
merged into master

Issue #636, opened 2018-09-20
add environment type filter to allEnvironments gql query


twardnw:

twardnw:
interesting, this was deployed, but the query returns an empty array with any filter on it against production. However, building lagoon locally and using the query there works as expected 😒

PR #637, opened 2018-09-20
#636 - add filter for env type to allenvironment query

twardnw:
Adds a filter for the environment type to the allEnvironments graphql query

### DIFF
--- a/services/api/src/schema.js
+++ b/services/api/src/schema.js
@@ -245,7 +245,7 @@ const typeDefs = gql`
     allProjects(createdAfter: String, gitUrl: String): [Project]
     allCustomers(createdAfter: String): [Customer]
     allOpenshifts: [Openshift]
-    allEnvironments(createdAfter: String): [Environment]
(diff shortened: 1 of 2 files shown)


services/api/src/schema.js
```diff
@@ -245,7 +245,7 @@ const typeDefs = gql`
     allProjects(createdAfter: String, gitUrl: String): [Project]
     allCustomers(createdAfter: String): [Customer]
     allOpenshifts: [Openshift]
-    allEnvironments(createdAfter: String): [Environment]
+    allEnvironments(createdAfter: String, envType: EnvType): [Environment]
```
Schnitzel: We have another filter within Projects to filter for the type of environments, but there it's just called `type`:
https://github.com/amazeeio/lagoon/blob/d7c190f56b87ad597e74d15f056268353087226f/services/api/src/schema.js#L188
therefore, let's rename this also to `type`

Schnitzel (changes requested):

Schnitzel (approved):

PR #632, opened 2018-09-19
Set up Keycloak and integrate with API

karlhorky:
~~**Merging blocked** on #575 and #627 because this contains commits from those PRs.~~

To use, (re)build containers:

```
docker-compose down
rm build/{api,keycloak,keycloak-db,local-api-data-watcher-pusher}
make build/api
make build/keycloak
make build/keycloak-db
make build/local-api-data-watcher-pusher
```

**TODO:**

- [x] Fork `keycloak-admin-client` to https://github.com/amazeeio/keycloak-admin-client
- [x] Fix security warnings
- [x] Merge [create group upstream](https://github.com/jderuere/keycloak-admin-client/tree/create_find_by_id_group) into fork. Done: amazeeio/keycloak-admin-client#1
- [x] Add `keycloak-admin-client` dependency to API
- [x] Switch to [`keycloak-admin`](https://github.com/Canner/keycloak-admin) instead of using our fork
- [x] Wait on Keycloak startup (~20 seconds) before accepting API requests
- [x] `addProject` mutation creates Keycloak group
- [x] `updateProject` with a new `name` renames Keycloak group
- [x] `updateProject` with a new `customer` deletes Keycloak users with customer-only project access from the Keycloak group
- [x] `updateProject` with a new `customer` adds Keycloak users with customer-only project access to the Keycloak group
- [x] `deleteProject` deletes Keycloak group
- [x] `deleteAllProjects` deletes all corresponding Keycloak groups
- [x] `addUser` mutation creates Keycloak user
- [x] `addUserToProject` adds a Keycloak user to a group
- [x] `removeUserFromProject` deletes a Keycloak user from a group
- [x] `removeAllUsersFromAllProjects` deletes all Keycloak users from all groups
- [x] `addUserToCustomer` adds a Keycloak user to the groups corresponding to the customer's projects
- [x] `removeUserFromCustomer` deletes a Keycloak user from the groups corresponding to the customer's projects
- [x] `removeAllUsersFromAllCustomers` deletes all Keycloak users from the groups corresponding to all customers' projects
- [x] Update of users
- [x] Delete of users
- [x] Move truncation code to each DAO file and add Keycloak deletion commands
- [x] Make `user.email` column unique
- [x] Require email to be passed to user creation command
- [x] Adapt addUser mutations in `api-data.gql` to also specify email
[...]

PR #651, opened 2018-10-01
GOV-27 Authenticate Lagoon UI via Keycloak

rocketeerbkw:
**Depends on #632** 

I added keycloak authentication to the UI. All unauthenticated requests will trigger a redirect to the keycloak login page, and then redirect back to the UI. The UI will take the JWT given by keycloak and use that as the access token for the graphql API.

I added keycloak authentication to the API. All requests will check if the access token is generated by keycloak, validate the token, and extract the user data. If the keycloak token is not passed, or is not valid, the API will fallback to the original JWT token authentication. This ensures that current systems (other servers, cli, etc) can still authenticate.

I added all the necessary keycloak configuration for the ui to function as part of the keycloak setup script.

TODO
- [x] Add admin role mapping to keycloak, so that UI will show all data to admins


fubarhouse:
@Schnitzel,

Just a small remark on authenticating via docker for mac:

> DNS name `host.docker.internal` should be used for host resolution from containers. Older aliases (still valid) are deprecated in favor of this one. (See https://tools.ietf.org/html/draft-west-let-localhost-be-localhost-06).

**Source**:
* https://docs.docker.com/docker-for-mac/release-notes/#docker-community-edition-18030-ce-mac59-2018-03-26

The following may also prove useful to you, which is preventing authentication on my end:
image

rocketeerbkw:
@fubarhouse I guess that will be a problem for local development in the long term then. Keycloak will not validate a JWT token unless the keycloak url configured on the [server side](https://github.com/amazeeio/lagoon/pull/651/files#diff-4be51b3aa4961beb3d088053b369e8a7R19) is exactly the same as the keycloak url configured on the [client](https://github.com/amazeeio/lagoon/pull/651/files#diff-258e0fc4ffeebf480a358615fe9d6c10R2). Currently, the only way that's possible when developing locally is to use `docker.for.mac.localhost` in both places.

When deployed in a full lagoon environment that has real URLs it will all work properly, but locally you won't be able to authenticate if you have mismatched URLs

Schnitzel:
merged into https://github.com/amazeeio/lagoon/pull/632

PR #655, opened 2018-10-02
initial draft of redis documentation.

Schnitzel:


Schnitzel:
@rtprio can you also add the D7 do the example settings within the documentation?

PR #660, opened 2018-10-02
use script rather than sh to invoke; will capture this log elsewhere.

Schnitzel:


Schnitzel:
@rtprio 
found this old gem :)

anything we want to work on?

rtprio:
This branch may need to be discarded:

1. go-crond logs to stdout
1. stdout should be captured by central logging.
1. users could check cli logs on their own with access.

rtprio:
This process will (does?) go to centralized logging; abandon.

PR #664, opened 2018-10-02
pvc reclaim script

rtprio:


Schnitzel:
we're running into some issues with
```
when rebuilding the project, it failed because of this error :The PersistentVolumeClaim "mariadb-data" is invalid: spec: Forbidden: field is immutable after creation
```
during lagoon deployments after the PVC has been migrated.

I think the issue comes from the fact that the PVC after a migration looks like that:

```
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"PersistentVolumeClaim","metadata":{"annotations":{},"creationTimestamp":null,"name":"mariadb-data","namespace":"xxx"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"5Gi"}},"storageClassName":"gluster"}}
```

while a PVC never touched by the migration looks like:

```
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"PersistentVolumeClaim","metadata":{"annotations":{},"name":"mariadb-data","namespace":"dtf-production"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"5Gi"}}}}
```

The `"storageClassName":"gluster"` is now suddenly there, as we need it during the migration to define which storageClassName we use. But Lagoon applies the PVC with `"storageClassName":""` (in order to use the default storageClass, which will be `gluster` again, but at that time the yaml parser doesn't know that yet and decides that it cannot be changed.

So I think the fix would be to use `oc create` instead of `oc apply` for the PVCs as `oc create` will not add any of the  `kubectl.kubernetes.io/last-applied-configuration` annotations

helpers/reclaim-pv.sh
```diff
@@ -0,0 +1,119 @@
...
+    
+    echo "dumping pvc ${PVC} to ${TMP}."
+    ${OC} get -o json pvc/${PVC} --export=true | jq 'del(.metadata.annotations, .metadata.selfLink, .spec.volumeName, .status)' > $TMP
+
+
+    ${OC} rollout pause deploymentconfig/pv-migrator
+
+    ${OC} volume deploymentconfig/pv-migrator --remove --name=${PVC}
+    ${OC} delete pvc/${PVC}
+    ${OC} apply -f $TMP && rm $TMP
```
Schnitzel:
```suggestion
    ${OC} create -f $TMP && rm $TMP
```

Schnitzel:
```suggestion
    ${OC} create -f $TMP && rm $TMP
```

Schnitzel (changes requested):

Schnitzel:
@rtprio 
can you run the updated script and see if the newly created PVC still has some `kubectl.kubernetes.io/last-applied-configuration` annotations in there?

PR #669, opened 2018-10-05
Fix auth errors when running `local-dev/es-kibana-watcher-pusher`

rocketeerbkw:
When running the full stack locally, I would get `logs-db` auth errors and had difficulty viewing indexed data in kibana.

I traced the auth errors to a change in some scripts in `logs-db-ui` which added credential environment variables. These variables were not added to `local-dev/es-kibana-watcher-pusher` which uses the same scripts as `logs-db-ui` so it was failing.


Schnitzel:
`es-kibana-watcher-pusher` is gone for good :)

PR #673, opened 2018-10-09
Add ability to bypass keycloak authentication for local development

rocketeerbkw:
If you have a `.env` file and it includes a `GRAPHQL_API_TOKEN` variable, that will be used instead of attempting to authenticate with keycloak


rocketeerbkw:
Merging as part of #675

PR #677, opened 2018-10-10
WIP: escape json

Leksat:


Schnitzel:
@rtprio I think you mentioned you want to work on this one, so maybe here is a good start :)

Leksat:
I have no time currently to work on this, but by plan was to use https://stedolan.github.io/jq/ to solve the issue. I saw `jq` is already used in a couple of places in Lagoon.

Usage is simple, with `-R` flag, it takes a raw string, and returns it as a JSON:
```
$ echo 'a quote " goes here' | jq -R
"a quote \" goes here"
```

rocketeerbkw:
`jq` approach seemed reasonable to me, addressed in #804

PR #679, opened 2018-10-11
Add setting for HSTS header policy to a route

eskp:
Adds "hsts" key for a route which tells Lagoon to create an appropriate
annotation on that route with the supplied to the key options. Which in
turn will enable and distribute HSTS header for that route.

More info:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
https://docs.openshift.com/dedicated/admin_guide/managing_networking.html#admin-guide-enabling-hsts

### DIFF
--- a/docs/using_lagoon/lagoon_yml.md
+++ b/docs/using_lagoon/lagoon_yml.md
@@ -57,6 +57,7 @@ environments:
         - "www.example.com":
             tls-acme: 'true'
             insecure: Redirect
+            hsts: max-age=31536000
     cronjobs:
      - name: drush cron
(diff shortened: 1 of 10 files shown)


dasrecht:
Hi @eskp - Great work!

Would you mind adding the new configuration part to the documentation too? 
https://github.com/amazeeio/lagoon/blob/master/docs/using_lagoon/lagoon_yml.md

Thanks in advance
/b

eskp:
Done @dasrecht 👍

dasrecht (approved):
Looks good to me. Thanks!

PR #692, opened 2018-10-18
redis service with persistent volume

rtprio:


Makefile
```diff
@@ -159,6 +160,7 @@ build/nginx-drupal: build/nginx images/nginx-drupal/Dockerfile
 build/varnish: build/commons images/varnish/Dockerfile
 build/varnish-drupal: build/varnish images/varnish-drupal/Dockerfile
 build/redis: build/commons images/redis/Dockerfile
+build/redis-persistent: build/commons images/redis-persistent/Dockerfile
```
Schnitzel: as `redis-persistent` depends on `redis` we need to reflect that here.
```suggestion
build/redis-persistent: build/redis images/redis-persistent/Dockerfile
```

images/redis/conf/redis.conf
```diff
@@ -11,6 +11,4 @@ maxmemory-policy allkeys-lru
 protected-mode no
 bind 0.0.0.0
 
-# this disabled the persistent cache
-save ""
-appendonly no
\ No newline at end of file
+include /etc/redis/${flavor:ephemeral}.conf
```
Schnitzel: shouldn't that be:
```suggestion
include /etc/redis/${FLAVOR:ephemeral}.conf
```
maybe nitpicking, but better readable

Schnitzel (changes requested):
just some small things!

images/redis/conf/redis.conf
rtprio: fixed in acf88dbcf152b4b67c33f6e9e46f203f95e78a2a

PR #699, opened 2018-10-19
EnvVars on projects and environments

### LINKED ISSUE #694: Allow ENV variables to be defined for Projects and Enviornments
The idea of this is, that via the Lagoon API there is the possibility to define an arbitrary amount of ENV variables either per Project or per Environment.

AC;
- [x] Env Variables are filled in the `lagoon-env` configmap of the environment.
- [x] If the same env variable is defined in the project and the environment, the one from the environment is used.
- [x] Additionally all defined Env variables are also exposed as `build-arg` during a Docker Build and can be consumed by the Docker Build

rocketeerbkw:
Fixes #694


rocketeerbkw:
Tests passed. I tested quickly some ways to escape the environment variables but had a lot of problems. Currently, users would be allowed to enter data into the lagoon api that will break their builds, but they could already do that with code.

PR #703, opened 2018-10-24
Overflowing of commit hashes on backups route

fubarhouse:
Affects services/ui.

The commit hashes are just unpleasant to work with on responsive devices. This change was intended to be used on media queries, but it is far more efficient and consistent when used outside of a media query as it also impacts other display resolutions.

It's only a couple of lines, but it's hopefully a start towards contributing more to this project in the future.

**What is does**

It prevents overflow of the commit hash comtainer on the backups page on all responsive devices and will shorten the hash where appropriate on the screen and represent any missing characters with an ellipsis.


Schnitzel:
wow thank you! let's see what the testing thinks :)

PR #704, opened 2018-10-24
Feature/gov 30 swappable logos

kristymoon:


rocketeerbkw (approved):
Tested, works as expected. Meets the requirements as I understand them.

Issue #688, opened 2018-10-16
add automatically generated routes to configmap


twardnw:
During the build of autogenerated routes, add a label. Then during the 'project wide env variables' portion of the build, select all routes with that label and add to an env var

Issue #609, opened 2018-09-10
Send `X-Robots-Tag: noindex, nofollow` also for autogenerated urls


Schnitzel:
currently we add `X-Robots-Tag: noindex, nofollow` headers to all development environment, it would be good to also add these headers to autogenerated domains (like https://varnish-drupal-example-master.ch.amazee.io)

In order for this to work, the webserver (nginx) needs to know which domains are the autogenerated domains or which URL patterns are used for the auto generated domains

twardnw:
@Schnitzel , any argument against carrying https://github.com/amazeeio/lagoon/blob/master/services/openshiftbuilddeploy/src/index.js#L170-L171 forward into the nginx container as an env var?

twardnw:
After discussion, we will add the automatically generated routes to the configmap ( #688 ), then nginx can use that list to set the headers appropriately

PR #711, opened 2018-10-25
#609 #687 #688 - env vars in configmap and x-robots-tag on autogenerated routes

twardnw:
This solves tickets #609 #687 #688
#687 is adding the openshift name to the configmap
#688 is adding the autogenerated route names in a configmap
#609 sets the x-robots-tag header on autogenerated routes to prevent crawler indexing


images/nginx/helpers/x-robots-header-development.conf
```diff
@@ -1,13 +1,9 @@
-# Load environment variable LAGOON_ENVIRONMENT_TYPE into nginx variable. LAGOON_ENVIRONMENT_TYPE will be replaced via envplate during docker-entrypoint
-set $lagoon_environment_type '${LAGOON_ENVIRONMENT_TYPE:-development}';
-
-# By default no x-robots-tag
-set $x_robots_tag '';
-
-# If this is an development environment do not index
-if ($lagoon_environment_type = 'development') {
-  set $x_robots_tag 'noindex, nofollow';
+# Load environment variable LAGOON_ENVIRONMENT_TYPE into nginx variable. LAGOON_ENVIRONMENT_TYPE will be replaced via envplate during docker-entrypo
```
Schnitzel: as LUA actually has access to the env variables itself, we could now load the env variables directly instead via the envplate system in the docker-entrypoint which is hard to understand.

be careful though with: https://stackoverflow.com/questions/41800071/unable-to-use-environment-variables-in-lua-code

images/nginx/helpers/x-robots-header-development.conf
```diff
@@ -1,13 +1,9 @@
...
+header_filter_by_lua_block {
+  host = string.gsub(ngx.var.host, "%p", "%%%1")
```
Schnitzel, line 4: can you document what this does?

Schnitzel (changes requested):
overall super awesome work, thanks! Just some improvements :)

images/nginx/helpers/x-robots-header-development.conf
twardnw: 👍🏻

Schnitzel (dismissed):

images/nginx/helpers/x-robots-header-development.conf
```diff
@@ -1,13 +1,12 @@
...
-
-# By default no x-robots-tag
-set $x_robots_tag '';
-
-# If this is an development environment do not index
-if ($lagoon_environment_type = 'development') {
-  set $x_robots_tag 'noindex, nofollow';
+# Set X-Robots-Tag to 'noindex, nofollow' for development environments and Lagoon autogenerated routes
+header_filter_by_lua_block {
+  -- escape characters in the hostname
```
Schnitzel, line 3: can you describe which characters we are escaping?

Schnitzel (changes requested):

Issue #707, opened 2018-10-24
default database in mariadb does not use default character set and collation


Schnitzel:
our my.cnf defines nicely:
https://github.com/amazeeio/lagoon/blob/80dfb2843f785fe47645b4caae2fc9603df3d871/images/mariadb/my.cnf#L12-L13

but the entrypoint which creates the default database is actually defining `CHARACTER SET utf8 COLLATE utf8_general_ci` which means the defaults from above are actually not used
https://github.com/amazeeio/lagoon/blob/80dfb2843f785fe47645b4caae2fc9603df3d871/images/mariadb/docker-entrypoint.bash#L78

the entrypoint should not define it's own character set and collation and instead just use the default (so if I understand correct, just use: `CREATE DATABASE IF NOT EXISTS \`$MARIADB_DATABASE\``

PR #717, opened 2018-10-25
#707 - change default charset and collation on DBs

twardnw:


Schnitzel:
so looks like now the keycloak is not booting, as the keycloak-db is created with `utf8mb4` which will not allow keycloak to setup the database. Means we would need to find a way to setup the keycloak DB with `utf8`, maybe configure an env variable to configure the characterset in the my.cnf and set that to `utf8` for keycloak-db?

PR #725, opened 2018-11-01
Adding a project should link keycloak users associated with the project customer

### LINKED ISSUE #721: Adding new project does not give access to existing Users of Customers in Keycloak
**Describe the bug**
When adding a new project to the Lagoon API it correctly creates the group in Keycloak with the same name, but it does not give the users of the customer that the project belongs too automatically access.

**Expected behavior**
When adding a new project to a customer, the users belonging to the customer have automatically access to the project.

rocketeerbkw:
Fixes #721

### DIFF
--- a/services/api/src/resources/project/resolvers.js
+++ b/services/api/src/resources/project/resolvers.js
@@ -215,24 +215,7 @@ const addProject = async (
   await KeycloakOperations.addGroup(project);
   await SearchguardOperations.addProject(project);
 
-  await Helpers.mapIfNoDirectProjectAccess(
-    R.prop('id', project),
-    cid,
-    async ({
-      keycloakUserId,
-      keycloakUsername,
-      keycloakGroupId,
-      keycloakGroupName,
-    }) => {
-      await keycloakClient.users.addToGroup({
-        id: keycloakUserId,
-        groupId: keycloakGroupId,
-      });
-      logger.debug(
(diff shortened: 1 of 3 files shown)


services/api/src/resources/project/resolvers.js
```diff
@@ -215,6 +215,25 @@ const addProject = async (
...
+  await Helpers.mapIfNoDirectProjectAccess(
+    R.prop('id', project),
+    cid,
+    async ({
+      keycloakUserId,
+      keycloakUsername,
+      keycloakGroupId,
+      keycloakGroupName,
+    }) => {
+      await keycloakClient.users.addToGroup({
```
Schnitzel: can we use `KeycloakOperations.addUserToGroup` (or maybe create another function as this one does quite some searching for userid and groupid)? I would like to not use `keycloakClient` in the `resolvers.js` anymore and abstract all Keycloak operations through `KeycloakOperations`

Schnitzel (changes requested):

services/api/src/resources/project/resolvers.js
rocketeerbkw: I don't see any way to avoid querying a lot for keycloak user/project information, since the keycloak-admin library doesn't return the data that was created (even the [tests](https://github.com/Canner/keycloak-admin/blob/master/test/groups.spec.ts#L26) run a `create` then a `find`). I did abstract away the map function in a Helper.

Schnitzel (approved):

PR #733, opened 2018-11-05
skip deploy take 2

rtprio:
Add ramda functions to make object introspection more robust.


services/webhooks2tasks/src/handlers/githubPush.js
```diff
@@ -3,6 +3,7 @@
 const { logger } = require('@lagoon/commons/src/local-logging');
 const { sendToLagoonLogs } = require('@lagoon/commons/src/logs');
 const { createDeployTask } = require('@lagoon/commons/src/tasks');
+const { R } = require('ramda');
```
Schnitzel:
```suggestion
const R = require('ramda');
```

services/webhooks2tasks/src/handlers/gitlabPush.js
```diff
@@ -3,6 +3,7 @@
 const { logger } = require('@lagoon/commons/src/local-logging');
 const { sendToLagoonLogs } = require('@lagoon/commons/src/logs');
 const { createDeployTask } = require('@lagoon/commons/src/tasks');
+const { R } = require('ramda');
```
Schnitzel:
```suggestion
const R = require('ramda');
```

Schnitzel (changes requested):
small change which should fix the failing tests.
Background: `ramda` exports a whole object that can be used directly and not a group of objects like `@lagoon/commons/src/logs` does

Issue #581, opened 2018-08-21
Persistent Redis Container


dasrecht:
We had the customer request of a redis container that persists the cache across container restarts

this would need a seperate deploy service type and a seperate configuration for redis

PR #734, opened 2018-11-05
Redis persistent, with volume

rtprio:
correct #581


dasrecht:
Hi there, sorry to pitch in at that late stage but i guess we need to think a bit more about how we configure the persistent redis.

We set the `maxmemory-policy` to `allkeys-lru` here - https://github.com/amazeeio/lagoon/blob/master/images/redis/conf/redis.conf#L8

I think we should look into the other eviction policies as when using redis as a persistent storage (e.g. for a queue) we should make sure no data gets lost so the `noeviction` policy would be an idea - but this will lock up redis when we reach the `maxmemory` configured.

Eviction Policies: https://redis.io/topics/lru-cache

Schnitzel:
@dasrecht 
mhh that is a good point, my understanding was that we need the persistent Redis still with `allkeys-lru` but we want it persistent so that a possible redis container restart does not loose any caching.

I guess we could make `maxmemory-policy` configurable via an env variable and by default have it set to  `allkeys-lru`?

PR #735, opened 2018-11-07
Add backup restores to api and openshift

### LINKED ISSUE #685: Backup Restore and monitoring of such Restores
With the new Restic based Backup System we can now allow the users to request a "Restore" of a specific Backup, while Restore means that we get a Webhook after the Restore is done with a Link to an S3 Bucket to download the Restored files as a Tar.

In order to request a Restore the Lagoon API needs to create a new Restore object within the Namespace of the environment:

```
apiVersion: backup.appuio.ch/v1alpha1
kind: Restore
metadata:
  name: restore-e260f07c3
spec:
  snapshot: e260f07c374e4a3319c5d46e688dab6f1eb23c3e61c166a37609d55762d2ee0b
  restoreMethod:
    s3: {}
  backend:
    s3: {}
```
Endpoint: `/apis/backup.appuio.ch/v1alpha1/namespaces/lagoon/restores`

The snapshot ID is the `backupID` from the list of existing backups of this environment (which are added to the LagoonAPI whenever there is a backup created).

After that Restore object is created, the Backup system will go and start the restore into an S3 Bucket, as soon as this is done, the Restic Backup sends another webhook with the following payload:

```
{
  "restore_location": "http://localhost:9000/restore/backup-test-2018-10-08T14:09:55+02:00.tar.gz",
  "snapshot_ID": "e260f07c374e4a3319c5d46e688dab6f1eb23c3e61c166a37609d55762d2ee0b",
  "restored_files": [
    "/Users/simonbeck/Desktop/data/output.tar.gz"
  ]
}
```
This should be consumed by the webhook handler system to add the restore_location into the Lagoon API.

My understanding how the whole thing would work:

1. User requests a restore from an existing backup
2. Lagoon creates `Restore` Object in Kubernetes and creates a Lagoon Object `Restore` as well with the status `Pending`
3. As soon as the backup is done, webhook about successful restore is received and the `restore_location` added to the Lagoon Object `Restore` and sets status to `Successful`
4. User can then download the Restore via visiting the `restore_location` that is in the Lagoon API.

rocketeerbkw:
Fixes #685


services/openshiftmisc/src/handlers/resticRestore.js
```diff
@@ -0,0 +1,97 @@
...
+    url: openshiftConsole,
+    insecureSkipTlsVerify: true,
+    auth: {
+      bearer: openshiftToken
+    }
+  });
+
+  try {
+    const restoreConfigPost = promisify(
+      kubernetes.ns(openshiftProject).restore.post
```
rocketeerbkw: I'm highly confident that this doesn't work as is. Minishift doesn't have this api, and the [appuio docs](http://docs.appuio.ch/en/latest/baas.html#) don't mention api support either.

Need to find out what the api path is, if it exists, or do something altogether different.

rocketeerbkw: Fixed and tested locally!

PR #759, opened 2018-11-14
Dynamic varnish 2

rtprio:


images/varnish/Dockerfile
```diff
@@ -1,7 +1,20 @@
 ARG IMAGE_REPO
 FROM ${IMAGE_REPO:-lagoon}/commons as commons
-FROM alpine:3.7
 
+
+FROM alpine:3.7 as vmod
+ENV VERSION=5.2
```
Schnitzel: to make this a bit more readable and understandeable can we:
- rename `VERSION` to `LIBVMOD_DYNAMIC_VERSION`
- add a version env variable for the bodyaccess libvmod as well

Schnitzel (changes requested):

images/varnish/Dockerfile
rtprio: good idea; fixed in e7d0ca23d38806 and d3ac64d3eb46a1f51d4.

Schnitzel (approved):

PR #762, opened 2018-11-15
Changes to layout of user information

### LINKED ISSUE #761: UI: Sidebar not wide enough to display full email addresses
**Describe the bug**
When on a project page, I can't see full email addresses.

**To Reproduce**
1. Go to 'https://ui-lagoon-master.ch.amazee.io/project?name=lead-2030'
2. See the email addresses are truncated

**Expected behaviour**
I expect to see the full email addresses.

**Screenshots**
Not good 
screenshot 2018-11-15 at 00 06 55

Good, but with inline browser hack
screenshot 2018-11-15 at 00 07 09

**Additional context**
I am using my 15" MacBook Pro and the viewport is showing as 1680px wide

fubarhouse:
This PR resolves #761.

Affects services/ui

* Displays user name and email on separate lines when the fields cannot fit side by side
* In the event a user name or email is exceptionally long, the sidebar will expand to accommodate the text

### DIFF
--- a/services/ui/src/components/Project/index.js
+++ b/services/ui/src/components/Project/index.js
@@ -76,7 +76,7 @@ class Project extends React.Component {
               <div className='field'>
                 {usersList.map(user =>
                   <div key={user.email} className='member'>
-                    {user.firstName ? <div>{user.firstName} {user.lastName} ({user.email})</div> : <div className="email">{user.email}</div>}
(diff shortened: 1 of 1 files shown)


Schnitzel:
wow! this is awesome, unfortunately we actually found a bug with the members list and are hiding it right now, but will be merged anyway and then when we bring the member list back it will be like in here. thanks to @ChandeepKhosa for reporting and @fubarhouse for the fix <3

PR #785, opened 2018-11-27
pre-validate .lagoon.yml

rtprio:


Schnitzel:
@rtprio 
is this still WIP?

images/oc-build-deploy-dind/validator/validate
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+BASE=$(dirname $0)
+
+echo "Linting files:"
+
+yamllint -c ${BASE}/config/yamllint.yaml .lagoon.yml | cat
+yamllint -c ${BASE}/config/yamllint.yaml docker-compose.yml | cat
```
Schnitzel: we can't assume that the file that we are using is named `docker-compose.yml` as it can be overwritten by `docker-compose-yaml:` in `.lagoon.yml`
so we need to parse the `.lagoon.yml` in order to know which docker-compose.yml we need to validate, see:
https://github.com/amazeeio/lagoon/blob/2780ac39f589c8701ed3cd39bbede5608cbf5070/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L19

Schnitzel (changes requested):

seanhamlin:
I think this is a good idea, discussed on the call today, as a first step, perhaps just lint `.lagoon.yml` and even then, only warn if not valid. Probably low value in linting docker files, as these would break locally for developers if invalid.

twardnw:
Lagoon has come a long way since this began, closing this for a fresh start

PR #852, opened 2019-01-16
Add delete button to environment page in UI

### LINKED ISSUE #770: UI: Ability to delete environments from UI
Due to being restrictions in place for the number of environments a project can have it would be useful to be able to delete an environment via the UI without having to close a PR or remove a branch.

It would also be be a good idea to include a confirmation step to ensure environments aren't accidentally removed and require a restore from backup to retrieve any environment specific data.

rocketeerbkw:
Fixes #770.

Only admins are allowed to delete production environments.

image


Schnitzel:
awesome! this will wait on https://github.com/amazeeio/lagoon/issues/838 which will completely deprecate the `rest2tasks` service

PR #879, opened 2019-02-05
841 disable git sha injection

### LINKED ISSUE #841: Setting to enable/disable injecting of LAGOON_GIT_SHA into DeploymentConfig
Currently we inject the GIT SHA of the current build into the deploymentconfig of many services:
https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/node/deployment.yml#L98-L99
https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/node-persistent/deployment.yml#L114-L115
https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/openshift-templates/nginx/deployment.yml#L101-L102

This provides a nice feature that if the application code needs to access the Git Hash of the current build it has it accessible in an environment variable `LAGOON_GIT_SHA`.
Unfortunately though this also means that every deploymentconfig will redeploy it's containers even if the docker image itself has not changed. That can cause much longer deployments and even sometimes service interruptions where they would not be needed.

Therefore the idea is to somehow define within `.lagoon.yml` globally for all services or in `docker-compose.yml` per service if the application needs the `LAGOON_GIT_SHA` env variable and only if it is requested by the service actually inject it into the deploymentconfig.

It probably is possible to have the default to not inject it anymore as I don't think many applications today actually use `LAGOON_GIT_SHA`, definitely the lagoon internal tests are using it, but there we could just adapt the `docker-compose.yml`

twardnw:
# Checklist

# Changelog Entry
## #841 
Change - disable the injection of the deployed SHA in the `LAGOON_GIT_SHA` environment variable. 
Adds - .lagoon.yml option `environment_variables.git_sha` to have the deployed SHA set.

# Closing issues
Closes #841


Schnitzel:
need to check: we would like to have still access to the SHA during build time

twardnw:
I suppose rather than setting/unsetting the SHA in build-deploy.sh, we could set the value back to 0s when each service is deployed in build-deploy-docker-compose.sh?

images/oc-build-deploy-dind/build-deploy.sh
```diff
@@ -28,6 +26,14 @@ if [ ! -f .lagoon.yml ]; then
   echo "no .lagoon.yml file found"; exit 1;
 fi
 
+INJECT_GIT_SHA=$(cat .lagoon.yml | shyaml get-value environment_variables.git_sha false)
+if [ "$INJECT_GIT_SHA" == "true" ]
+then
+  LAGOON_GIT_SHA=`git rev-parse HEAD`
+else
+  LAGOON_GIT_SHA="0000000000000000"
```
Schnitzel:
```suggestion
  LAGOON_GIT_SHA="0000000000000000000000000000000000000000"
```

tests/tests/features/disable-inject-git-sha.yaml
```diff
@@ -0,0 +1,45 @@
...
+    project: "{{ project }}"
+  tasks:
+  - include: ../../tasks/rest/deploy-no-sha.yaml
+
+- name: "{{ testname }} - check that the git sha is 0's"
+  hosts: localhost
+  serial: 1
+  vars:
+    url: "{{ check_url }}"
+    expected_content: "LAGOON_GIT_SHA=0000000000000000"
```
Schnitzel:
```suggestion
    expected_content: "LAGOON_GIT_SHA=0000000000000000000000000000000000000000"
```

Schnitzel (changes requested):

PR #892, opened 2019-02-11
Update mariadb and mariadb-galera images to use common Alpine version

rocketeerbkw:
Currently tests are failing in CI and locally. This was caused by the most recent release of alpine linux (3.9) and it's upgrade to mariadb 10.3.x.

The issues were:
1) Some config options were removed in [10.3.1](https://mariadb.com/kb/en/library/mariadb-1031-changelog/): [innodb_file_format](https://mariadb.com/kb/en/library/innodb-system-variables/#innodb_file_format) and [innodb_large_prefix](https://mariadb.com/kb/en/library/innodb-system-variables/#innodb_large_prefix)
   - Both of the settings lagoon images were using are the new defaults so this should not cause issues just removing the settings
2) Some cli options were removed in [10.3.7](https://mariadb.com/kb/en/library/mariadb-1037-changelog/): `mysql_install_db` `--skip-auth-anonymous-user` was removed in favor of `--skip-test-db`
3) New default config files are present in `/etc` which take precedence over the locations used in docker builds

In addition to fixing the above, the mariadb-galera image has been updated to use `alpine:latest` and mariadb 10.3.x also.
 
# Changelog Entry

Bugfix - Fix mariadb configs and options to be compatible with 10.3.x


rocketeerbkw:
This PR takes the approach of upgrading mariadb (single and cluster). Another option would be to pin `mariadb` to the current version, which is what `mariadb-galera` is currently doing.

smlx:
Should we pin alpine to `3.11` rather than `latest` as per https://github.com/amazeeio/lagoon/pull/1531#issuecomment-582706129?

This PR also needs a rebase onto master.

tobybellwood:
agree - let's rework this to match the other dockerfile formats.

tobybellwood:
ugh - latest release mariadb on alpine is 10.4.12 - which has some changes around the root user which will mean a couple of changes to the init bash script (https://mariadb.com/kb/en/authentication-from-mariadb-104/)

tobybellwood:
Updated PR for Alpine 3.11 / MariaDB 10.4 - testing will show if Galera is ready...

seanhamlin:
Something I spotted was that [Galera is now on v4](https://mariadb.com/kb/en/changes-improvements-in-mariadb-104/#galera-4) with MariaDB 10.4.

I have no idea what the upgrade path looks like, but this will need to be tested.

smlx (approved):
LGTM, though there are now conflicts with master..

tobybellwood:
Closing, merged in #1866

PR #897, opened 2019-02-13
Underpinnings for Python, CKAN projects in Lagoon

stooit:
- base alpine Python images (2.7.15, 3.7.2)
- base CKAN image (based on 2.7.15)
- base Solr + Postgres


images/solr-ckan/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG SOLR_MAJ_MIN_VERSION
+ARG IMAGE_REPO
+FROM ${IMAGE_REPO:-lagoon}/solr:${SOLR_MAJ_MIN_VERSION}
+ARG SOLR_MAJ_MIN_VERSION
```
steveworley, line 4: Do you need a second `SOLR_MAJ_MIN_VERSION ` argument?

Schnitzel: yes, as each `FROM` resets all `ARG`

PR #903, opened 2019-02-18
Replace newlines in the GITLAB_DEFAULT_CUSTOMER_SSH_PRIVATEKEY variable

thom8:
Newlines are not properly replaced when being set via the `GITLAB_DEFAULT_CUSTOMER_SSH_PRIVATEKEY` environment variable.


thom8:
Closing as we will no longer be using a default key.

PR #908, opened 2019-02-21
Ensure the searchguard init script is only run on initial deployment

thom8:
We have had a number of cases where the init script is run on a restart of the `logs-db` pod and resets the tenants to the default configuration.

This PR ensures that the searchguard init script `sgadmin_demo.sh` is only run on the initial deployment of `logs-db`.

You can also delete the state file and redeploy if you need to reset an existing deployment.

### DIFF
--- a/docs/administering_lagoon/install.md
+++ b/docs/administering_lagoon/install.md
@@ -62,6 +62,13 @@ In order to use a local Lagoon to deploy itself on an OpenShift, we need a subse
 
 4.  As soon as the build is done, go to the `Application > Deployments` section of the OpenShift Project and you should see all the Lagoon Deployment Configs deployed and running. Also go to `Application > Routes` and click on the generated route for `rest2tasks` (for a local OpenShift this will be http://rest2tasks-lagoon-develop.192.168.99.100.xip.io/), if you get `welcome to rest2tasks` as result, you did everything correct, bravo!
 
+### Searchguard
+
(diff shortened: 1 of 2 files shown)


Schnitzel:
nice idea, but we need to make sure that this only runs on the very first node of an elasticsearch cluster. As in a cluster you have multiple nodes and all of them will run the same code, aka we would initialize the cluster the same amount that we have nodes. If we add a node to a cluster, it will also cause the cluster to be initialized again...

Schnitzel (changes requested):
see comment above

Schnitzel:
interesting idea, though I'm not sure if this really solves our problem:
The current issue is that sometimes an ES node comes back into the cluster and is not fully synchronized yet with the rest of the cluster, because of that the script will run `sgadmin_demo.sh`.
With the new code the same could happen again: The ES node comes back into the cluster, is not fully synced, therefore the user `init` does not exist yet and it will start to initialize again.
Or maybe I'm missing something?

I would suggest to just remove the init code all together, and we document within the Setup Documentation of Lagoon that an admin needs to run `sgadmin_demo.sh` at the very very first time manually.

thom8:
@Schnitzel removed the initialisation in `start.sh` and added docs for the initial setup.

shreddedbacon (commented):
LGTM

Schnitzel (approved):

PR #914, opened 2019-02-25
876 environment limit display

### LINKED ISSUE #876: Display the limit of environments
Current issues:
- Developers might not know about the limit of environments
- Currently there is no way to learn the limit (maybe via API? but still not so obvious)

Proposed solutions:
- Variant A: add the limit of environments to the Lagoon UI
- Variant B: add a (Slack) notification, like  
    > :warning: **[project]** PR [#123 (PR name)](#) opened in Company/project
    > ERROR: environment won't be created because the limit of environments is already reached.
- Variant C: A+B

rtprio:
# Checklist

Previously, there was no way to determine what the development environment limit was. This change will add it as a field to the UI.

# Changelog Entry
 
Improvement - Displays number of used and limit of non-production environments (#876 )

# Closing issues
Closes #876

### DIFF
--- a/services/ui/src/components/Project/index.js
+++ b/services/ui/src/components/Project/index.js
@@ -68,11 +68,11 @@ class Project extends React.Component {
               <div className='field'>{this.props.project.pullrequests}</div>
             </div>
           </div>
-        </div>
-        <div className='field-wrapper envlimit'>
-          <div>
-            <label>Non-Production environments in use</label>
-            <div className='field'>{this.props.project.environments.length} of {this.props.project.developmentEnvironmentsLimit} </div>
+          <div className='field-wrapper envlimit'>
+            <div>
+              <label>Non-Production environments in use</label>
+              <div className='field'>{this.props.project.environments.length} of {this.props.project.developmentEnvironmentsLimit} </div>
+            </div>
           </div>
         </div>
         <div className="environments-wrapper">
--- a/services/ui/src/components/Project/index.js
+++ b/services/ui/src/components/Project/index.js
@@ -74,7 +74,7 @@ class Project extends React.Component {
           </div>
           <div className='field-wrapper envlimit'>
             <div>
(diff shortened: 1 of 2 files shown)


rtprio:
screen shot

services/ui/src/components/Project/index.js
```diff
@@ -68,6 +72,12 @@ class Project extends React.Component {
               <div className='field'>{this.props.project.pullrequests}</div>
             </div>
           </div>
+          <div className='field-wrapper envlimit'>
+            <div>
+              <label>Non-Production environments in use</label>
+              <div className='field'>{this.state.developEnvironmentCount} of {this.props.project.developmentEnvironmentsLimit} </div>
```
rocketeerbkw: Is `developmentEnvironmentsLimit` always a number? It looks like it can be set to null to mean unlimited? Think it would be nice to display "x of unlimited" or "x of ∞" if that's the case.

rtprio: That's a good point.  New projects are created with a default (5), but it is nullable so that's something I should consider.

rtprio: I've updated this to safely tolerate null.

services/ui/src/components/Project/index.js
```diff
@@ -68,6 +72,12 @@ class Project extends React.Component {
               <div className='field'>{this.props.project.pullrequests}</div>
             </div>
           </div>
+          <div className='field-wrapper envlimit'>
+            <div>
+              <label>Non-Production environments in use</label>
```
Schnitzel:
```suggestion
              <label>Development environments in use</label>
```

Let's call them what they are: `Development Environments`, I would like to not introduce a new word "non-production" when they are called "development" in the API.

Schnitzel (changes requested):

PR #917, opened 2019-02-26
#877 - do not announce retry

### LINKED ISSUE #877: Raise timeouts for project removal
**Describe the bug**
When a branch or PR is removed there are a lot of error messages currently. 
This is because our expectation is that we only terminate containers (which is pretty fast) as we also have databases that are provisioned via service catalog the decomission can take longer - even up to a few minutes

**To Reproduce**
1. create branch
2. wait till environment is fully started and database has been provisioned
3. remove branch

**Expected behavior**
Timeouts are set to accomodate slow removals - which will lead to less errors casted towards the ci channels.

**Screenshots**
screenshot 2019-02-01 15 29 43

twardnw:
Remove announcement of retry during environment deletion. The retry can be found in the logs, and announcing the retry in slack/rocketchat causes confusion for customers.

closes #877

### DIFF
--- a/services/openshiftremove/src/index.js
+++ b/services/openshiftremove/src/index.js
@@ -194,4 +194,8 @@ ${lastError}
 
 }
 
-consumeTasks('remove-openshift', messageConsumer, deathHandler)
+const retryHandler = async (msg, error, retryCount, retryExpirationSecs) => {
+  return
+}
+
+consumeTasks('remove-openshift', messageConsumer, retryHandler, deathHandler)
--- a/services/openshiftremove/src/index.js
+++ b/services/openshiftremove/src/index.js
@@ -194,23 +194,4 @@ ${lastError}
 
 }
 
-const retryHandler = async (msg, error, retryCount, retryExpirationSecs) => {
-  const {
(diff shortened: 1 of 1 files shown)


services/openshiftremove/src/index.js
```diff
@@ -194,23 +194,4 @@ ${lastError}
...
-`*[${projectName}]* remove \`${openshiftProject}\` ERROR:
-\`\`\`
-${error}
-\`\`\`
-Retrying in ${retryExpirationSecs} secs`
-  )
-}
-
-consumeTasks('remove-openshift', messageConsumer, retryHandler, deathHandler)
+consumeTasks('remove-openshift', messageConsumer, deathHandler)
```
Schnitzel: that will unfortunately not work, as the `consumeTasks` expects as the third parameter a function that is called during a retry. Now the `deathHandler` will be called on a retry which will send a full error to the notification system on the first retry.

I would suggest to leave the function `retryHandler` existing, just don't do anything in it and just run `return`

Schnitzel (changes requested):

PR #936, opened 2019-03-05
Add minikube setup/run/destroy to makefile

rocketeerbkw:
Copy/paste all the minishift make targets to use minikube.

Todo:
- [ ] Replacing IPs in files now has to know which think is for minishift and which is for minikube
- [ ] What parts of the openshift config have alternatives in k8s? For example, setting up a cluster admin user and roles


smlx:
Can this PR be closed now that we've gone the k3d route?

rocketeerbkw:
Ya, this was attempt #1 and now deprecated

PR #938, opened 2019-03-05
Cleanup Drupal settings files with PHPCBF

dan2k3k4:


Schnitzel:
@dan2k3k4 
I'm a bit confused, so is this a legit PR? :)

dan2k3k4:
@Schnitzel I used `phpcbf --standard=Drupal` on the files.

The changes just make sure we have cleaner looking code, the actual code does not change at all; but since this code is used as a reference for starting to "lagoonize" a website, I felt like standardising the code a little with best practices.

PR #959, opened 2019-03-11
API: Disallow duplicate ssh keys

### LINKED ISSUE #577: Prevent adding same SSH key multiple times
**Describe the bug**
It is possible to add the same SSH key multiple times via the GraphQL API

**To Reproduce**
1. Add any SSH Key via `addSshKey()` mutation
2. Add it a second time, still works

**Expected behavior**
It should throw an error that the key already exists

rocketeerbkw:
# Checklist

When the API has duplicate api keys in the system, it causes problems with some parts of lagoon. This adds a unique key constraint to the api db via ssh key fingerprints, and updates the api to save/get said fingerprints.

# Changelog Entry

Bugfix - API no longer allows adding duplicate ssh keys (#577)

# Closing issues
Closes #577


rocketeerbkw:
I added a [script ](https://github.com/amazeeio/lagoon/blob/577-api-disallow-duplicate-ssh-keys/helpers/577-api-db-generate-ssh-key-fingerprints.sh)to run after deployment to generate fingerprints for all existing ssh keys, but I'm stuck on how to make it work "properly" (like the other scripts in `helpers`.

PR #985, opened 2019-03-22
980 autoidle environments

### LINKED ISSUE #980: Allow per-environment idling (instead of per-project)
Currently Lagoon only allows you to idle environments on a per-project basis - i.e. all on or all off.

It'd be great to be able to set individual environments to non-idle (e.g. for cron testing etc as per #949, or for running larger test plans) - this would also give us more  ability to mitigate the costs involved in projects with multiple environments.

twardnw:
# Checklist

# Changelog Entry
Improvement - Add ability to disable auto-idle per environment #980 
Lagoon administrators can now update an existing environment to disable auto-idling

# Closing issues
Closes #980

### DIFF
--- a/services/auto-idler/idle-services.sh
+++ b/services/auto-idler/idle-services.sh
@@ -53,53 +53,54 @@ echo "$DEVELOPMENT_ENVIRONMENTS" | jq -c '.data.developmentEnvironments[] | sele
           ENVIRONMENT_AUTOIDLE=$(echo "$environment" | jq -r '.autoIdle')
           echo "$OPENSHIFT_URL - $PROJECT_NAME: handling development environment $ENVIRONMENT_NAME"
 
-          # Check if this environment has hits
-          HITS=$(curl -s -u "admin:$LOGSDB_ADMIN_PASSWORD" -XGET "http://logs-db:9200/router-logs-$ENVIRONMENT_OPENSHIFT_PROJECTNAME-*/_search" -H 'Content-Type: application/json' -d'
-          {
-            "size": 0,
-            "query": {
-              "bool": {
-                "filter": {
-                  "range": {
-                    "@timestamp": {
-                      "gte": "now-4h"
+          if [ "$ENVIRONMENT_AUTOIDLE" == "0" ]; then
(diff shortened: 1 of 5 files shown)


services/auto-idler/idle-services.sh
```diff
@@ -68,7 +70,10 @@ echo "$DEVELOPMENT_ENVIRONMENTS" | jq -c '.data.developmentEnvironments[] | sele
             }
           }'| jq ".hits.total")
 
-          if [ ! $? -eq 0 ]; then
+          if [ "$ENVIRONMENT_AUTOIDLE" == "0" ]; then
+            echo "$OPENSHIFT_URL - $PROJECT_NAME: $ENVIRONMENT_NAME idling disabled, skipping"
+            continue
+          elif [ ! $? -eq 0 ]; then
```
Schnitzel: this `elif [ ! $? -eq 0 ]; then` is testing the return code of the previous executed command, which needs to come directly after the `curl` command.

Let's add the check for environment idling before the hits loading :)

Schnitzel (changes requested):

services/auto-idler/idle-services.sh
twardnw: Makes sense to me :)

services/auto-idler/idle-services.sh
```diff
@@ -49,52 +50,57 @@ echo "$DEVELOPMENT_ENVIRONMENTS" | jq -c '.data.developmentEnvironments[] | sele
...
-            "size": 0,
-            "query": {
-              "bool": {
-                "filter": {
-                  "range": {
-                    "@timestamp": {
-                      "gte": "now-4h"
+          if [ "$ENVIRONMENT_AUTOIDLE" == "0" ]; then
+            echo "$OPENSHIFT_URL - $PROJECT_NAME: $ENVIRONMENT_NAME idling disabled, skipping"
+            continue
```
Schnitzel, line 58: let's keep the `if` crazyness here down, it's already wrapped in many ifs in here. 

If we use `continue` we don't need an `else`, as the `continue` anyway will stop executing and go to the next loop.

So we can just run and `fi` instead of an `else`

Schnitzel (changes requested):
just antother small thing.

Schnitzel (approved):

PR #1010, opened 2019-04-01
992 - fixing curator runs by making sure .bashrc is sourced for cronjobs

### LINKED ISSUE #992: curator cronjob does not seem to cleanup elasticsearch
**Describe the bug**
It seems that the curator cronjob does not work correctly at the moment can be checked on ch1-lagoon.

**Expected behavior**
curator runs and cleans up elasticsearch :)

dasrecht:
# Checklist

Curator was not able to connect to elasticsearch due to the missing `LOGSDB_HTTP_AUTH` environment variable. The change makes sure the `.bashrc` file gets sourced which makes in turn sure that the environment variable is present.

# Changelog Entry
Bugfix - The Environment variable was missing so curator could not connect to elasticsearch (#992 )

# Closing issues
closes #992


.lagoon.yml
```diff
@@ -56,5 +56,5 @@ environments:
         service: auto-idler
       - name: curator
         schedule: '01 0 * * * '
-        command: /usr/bin/curator --config /curator/curator.yml /curator/actions.yml
+        command: . /home/.bashrc && /usr/bin/curator --config /curator/curator.yml /curator/actions.yml
```
Schnitzel: let's use `/lagoon/cronjob.sh` as this will not only source bash, but also run all other entrypoints, etc. :)

Schnitzel (changes requested):

PR #1012, opened 2019-04-01
Use a new MariaDB connection object for every api request

### LINKED ISSUE #920: refactor api mariasql usage: open new connection for each graphql request
we're sharing the mariasql connection object within the api container, this is great for memory usage, but unfortunately mariasql itself has a queue. So even if we have multiple graphql requests going on, the requests to the database are done sequently.
see https://github.com/mscdex/node-mariasql/issues/17#issuecomment-16325652
> All queries are placed in a local queue (due to the fact that the MySQL/MariaDB protocol is synchronous).

therefore I think we should initiate a new mariasql object for each graphql request, reuuse during the graphql request and then destroy it after the request has been parsed

rocketeerbkw:
# Checklist

Moved the `sqlClient` constructor inside of the Apollo server and express middlewares so that a new connection will be generated for every request. Putting it in the Apollo Server and passing it to resolvers via context means the connection will be shared when multiple resolvers need to be called to fulfill one request.

90% of the code changes are just adding `sqlClient` to the resolvers and auto-formatting changes. The magic happens in `apolloServer.js` and `authMiddleware.js`

# Changelog Entry

Improvement - Each API request will generate a new MariaDB connection (#920)

# Closing issues
closes #920

### DIFF
--- a/services/api/src/apolloServer.js
+++ b/services/api/src/apolloServer.js
@@ -113,6 +113,17 @@ const apolloServer = new ApolloServer({
       path: error.path,
     };
   },
+  plugins: [
+    {
+      requestDidStart: () => ({
+        willSendResponse: response => {
+          if (response.context.sqlClient) {
+            response.context.sqlClient.end();
+          }
+        },
+      }),
+    },
+  ],
 });
 
 module.exports = apolloServer;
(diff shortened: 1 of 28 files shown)


rocketeerbkw:
Seeing a lot of error messages now about too many connections:

> api_1                            | error:  message=Too many connections, stack=Error: Too many connections, code=1040

It looks like connections must explicitly be closed, combined with a new connection for every request makes sense that it runs out. Just looking at the docs for apollo server I can't see any way to close the connection after a request has been "finished." Will have to dig into code or ask SO or somewhere.

rocketeerbkw:
Found an undocumented plugin system in apollo server that can listen to request pipeline events. I no longer get connection errors, and tests are green.

Issue #1028, opened 2019-04-04
Add support for next-gen image formats (WebP) to php base docker image 


garethhallnz:
This is not really a bug and more of a feature request.

JPEG 2000, JPEG XR, and WebP are image formats that have superior compression and quality characteristics compared to their older JPEG and PNG counterparts. Encoding your images in these formats rather than JPEG or PNG means that they will load faster and consume less cellular data.

WebP is supported in Chrome and Opera and provides better lossy and lossless compression for images on the web.

**To Reproduce**
Given a Drupal site
1.  Install WebP module (https://www.drupal.org/project/webp)
2. Configure an image style using a next-gen format
3. View a page where the image style is used
5. An optimized image will not be generated
4. View watchdog log; You'll see
```Error: Call to undefined function Drupal\webp\imagewebp() in Drupal\webp\Webp->createWebpCopy() ```

**Expected behaviour**
An optimized image should be created

**Comment by Schnitzel** [comment]
Thanks @garethhallnz for the report.

Even though that the error is rather weird (I would have expected for it to not find a php function, rather not finding a function in the Drupal namespace), but this is maybe expected by the module.

In any case I looked at the php-fpm image and found that it does not support WebP:
```
$ docker run --rm -it lagoon/php:7.2-fpm sh
php:/app$ php -r "print_r(gd_info());"
Array
(
    [GD Version] => bundled (2.1.0 compatible)
    [FreeType Support] =>
    [GIF Read Support] => 1
    [GIF Create Support] => 1
    [JPEG Support] => 1
    [PNG Support] => 1
    [WBMP Support] => 1
    [XPM Support] =>
    [XBM Support] => 1
    [WebP Support] => 
    [JIS-mapped Japanese Font Support] =>
)
```

After the above patch applied there is now WebP support:

```
$ docker run --rm -it lagoon/php:7.2-fpm sh
php:/app$ php -r "print_r(gd_info());"
Array
(
    [GD Version] => bundled (2.1.0 compatible)
    [FreeType Support] =>
    [GIF Read Support] => 1
    [GIF Create Support] => 1
    [JPEG Support] => 1
    [PNG Support] => 1
    [WBMP Support] => 1
    [XPM Support] =>
    [XBM Support] => 1
    [WebP Support] => 1
    [BMP Support] => 1
    [JIS-mapped Japanese Font Support] =>
)
```

cdchris12:
Since this was resolved with #1029 , I'm going to close this issue.

[...]

PR #1029, opened 2019-04-04
#1028 support for WebP

Schnitzel:


garethhallnz:
I'd love to review the change and report back to here as to whether it fixed the issue or not, but I simply don't know how. What I mean to say is I am using ```FROM amazeeio/php:7.2-cli-drupal``` in my dockerfile but the change you have made is against the base image higher up the stack (obviously expected).

How would I test it and am I even required too? Happy to help out if I can.

garethhallnz (approved):
I have run this locally and seems to work for me.

PR #1031, opened 2019-04-05
1001 kubernetes cronjobs

### LINKED ISSUE #1001: Reimplement Lagoon Cronjobs with Kubernetes Cronjobs
- [ ] If the defined cronjob in `.lagoon.yml` is every 15mins or often: Keep the current system (injects `CRONJOBS` environment variable into a running pod)
- [ ] If the defined cronjob in `.lagoon.yml` is less often than 15mins: reintroduce the system we had once https://github.com/amazeeio/lagoon/commit/8ac2bd3758f8937afca164c7b70229b39b4167a2

Additional things:
- [x] Update CLI Auto Idler code to check if `CRONJOBS` is set, if not, idle the CLI 

This solves current issues:
1. We can idle more pods (basically all cli pods for production environments) - less load on cluster
2. this will run cronjobs for non production environments

rtprio:
If the cron is less often than 15 minutes, run it in a cronjob native type instead.
If the running pod has a cron [which is more frequent than 15 minutes] exclude it from the auto idler.

# Changelog Entry
Change- Don't idle pods that have cronjobs. If a cron is less often than 15 minutes, run it in a cronjob type rather than in the existing pod. ( #1001 )

# Closing issues
closes #1001


images/oc-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -11,6 +11,24 @@ function outputToYaml() {
   set -x
 }
 
+function checkFrequency() {
+  #takes a unexpanded cron schedule, returns true if it's less often that 15 minutes
+  MINUTE=$(echo $1 | (read -a ARRAY; echo ${ARRAY[0]}) )
+  if [[ $MINUTE =~ ^(M|H|\*)\/([0-5]?[0-9])$ ]]; then
+    STEP=${BASH_REMATCH[2]}
+    if [ $STEP -ge 15 ]; then
+      echo -n true
```
Schnitzel: can we use proper return codes instead of an `echo`? see https://stackoverflow.com/a/19401753/2397973

Schnitzel: also can we rename `checkFrequency` to something that is more clear? As you expect an `true` and `false` as a return, I would call the function also something like `isScheduleMoreOftenThanEvery15Min` (just an example function name)

Schnitzel (changes requested):

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
rtprio: 👍

siva2krish:
mariadb-single: Database - using mariadb direct image for Drupal containers.
Testing is done. However, CronJob version is not yet compatible for deployment with lagoon environment.tested using our helm charts..!

PR #1065, opened 2019-04-30
Allow dashboard log container to be "full height"

### LINKED ISSUE #1064: Allow dashboard log container to be "full height"
When viewing deployment logs in the dashboard, the logs are typically very long, but the log window is limited to 600px heigh.  Given there is nothing else on the deployment page other than the logs, would be good if the height wasn't limited.

There's a few benefits to this:

1. Users don't have to scroll as much because more of the log output fits on their screen at once.
1. Once the container is not-height restricted, the browser's behaviour of showing where search matches occur in the document by flagging them on the its scroll bar kicks in, this is great for jumping between areas areas of the output where a search string exists.

Feng-Shui:
PR for #1064 

See: https://www.govcms.support/support/tickets/2073

# Checklist

# Changelog Entry
Change - Remove the 600px hight limit for the logs container in the deployment section of the dashboard, this allows the container to be full height #1064 

# Closing issues
Closes #1064

### DIFF
--- a/services/ui/src/components/LogViewer/index.js
+++ b/services/ui/src/components/LogViewer/index.js
@@ -16,7 +16,6 @@ const LogViewer = ({ logs }) => (
           font-family: 'Monaco', monospace;
           font-size: 12px;
           font-weight: 400;
-          height: 600px;
           margin: 0;
           overflow-wrap: break-word;
           overflow-x: scroll;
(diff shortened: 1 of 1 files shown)


rocketeerbkw (changes requested):
Can you change this to min-height instead? I think the left sidebar area doesn't handle the main content area being very short well (in the case where logs are not available).

Feng-Shui:
Yep, PR updated to change it to `min-height: 600px`.

rocketeerbkw (approved):
Thanks!

PR #1087, opened 2019-05-20
Improvement: Replace gatter with gather

dan2k3k4:


Feng-Shui:
I'm sad to see this change go in, "gatter" has become part of the general parlance in our Teams chat.

PR #1104, opened 2019-05-28
Add `userIdByEmail` GraphQL query

### LINKED ISSUE #893: Add allUsers query to GQL API
Right now user information can only be accessed whilst they are attached to a project or customer.

If a user is not attached to either and the id of the user in the lagoon db, you cannot extract the information to allocate them to a project or customer. This is largely due to the fact you need to know the Lagoon user ID before you can query or mutate the object.

I believe it sensible to ask for a new query to allow people to identify or find users when these other mechanisms are not available, feel free to make other recommendations. In our case I connected to the API database in order to retrieve this value - but this is obviously not the preferred way.

cdchris12:
When creating a new user, it is possible to lose the id returned by that query. Without knowing that id value, it is then impossible to lookup that user any other way if that user has not yet been added to a project or customer. This new query allows you to query the database for which user is associated to which email address.

# Changelog Entry
Added `userIdByEmail` query to the GraphQL API

# Closing issues
Closes #893


rocketeerbkw (commented):
Is this intended to be used by anyone (instead of [admin only](https://github.com/amazeeio/lagoon/blob/master/services/api/src/resources/user/resolvers.js#L417))? Since we don't have any kind of api limits in place, it would allow anyone with access to do a user enumeration attack.

For the moment, "anyone with access" means existing lagoon users and not the whole internet. But it would allow users to query by email even from other clients/companies using amazee.io.

cdchris12:
@rocketeerbkw The intention for this is indeed to let anyone (read: non-admin users) use it. It would be easy enough to make this endpoint only available to admins, but I think that's a question better posed to @Schnitzel

Schnitzel:
Let's get RBAC 1.0.0 out first and then we can add this

Schnitzel:
@rocketeerbkw please check for RBAC additions, thanks!

rocketeerbkw:
The original paint point of the issue was not being able to add users to projects/customers w/o knowing the ID. Since RBAC no longer requires knowing user IDs (all mutations accept ID or email) and this PR only adds a query that returns a user ID, I think it's not worth updating to work with RBAC.

I could be convinced otherwise though, re-open if you disagree? @cdchris12 @Schnitzel

PR #1108, opened 2019-05-29
Add a .lagoon.env catchall option for variables

### LINKED ISSUE #1107: Provide Lagoon-specific env vars via a file
Currently the only way to provide env vars to Lagoon from code is via the .env file.  However, this is also how we provide local env vars - we have a use case to provide variables to Lagoon that may differ from local ones.

We would use .lagoon.env.$BRANCHNAME - but these would have to be created for every branch - does it make more sense to have a single .lagoon.env file to hold all these?

In our case, we can lock this so that clients can't modify prod variables without going through support, but can test locally themselves via the .env file - and still modify and commit their own .env files as required.

tobybellwood:
Adding a .lagoon.env file to allow overwriting variables for all branches on Lagoon only - not affecting Local dev (customisation possible via .env file)
 
# Checklist

# Changelog Entry
Provides a .lagoon.env file for catch-all variables (#1107 )

# Closing issues
Closes #1107

### DIFF
--- a/images/commons/lagoon/entrypoints/50-dotenv.sh
+++ b/images/commons/lagoon/entrypoints/50-dotenv.sh
@@ -28,7 +28,7 @@ set -a
 [ -f .env.defaults ] && . ./.env.defaults  || true
 [ -f .env ] && . ./.env  || true
 # Provide a file that adds variables to all Lagoon envs, but is overridable in a specific env below
-[ -f .env ] && . ./.lagoon.env  || true
+[ -f .lagoon.env ] && . ./.lagoon.env  || true
 [ -f .lagoon.env.$LAGOON_GIT_BRANCH ] && . ./.lagoon.env.$LAGOON_GIT_BRANCH  || true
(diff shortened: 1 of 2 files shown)


images/commons/lagoon/entrypoints/50-dotenv.sh
```diff
@@ -27,6 +27,8 @@ export -p > $TMPFILE
 set -a
 [ -f .env.defaults ] && . ./.env.defaults  || true
 [ -f .env ] && . ./.env  || true
+# Provide a file that adds variables to all Lagoon envs, but is overridable in a specific env below
+[ -f .env ] && . ./.lagoon.env  || true
```
Schnitzel:
```suggestion
[ -f .lagoon.env ] && . ./.lagoon.env  || true
```

Schnitzel (changes requested):

Schnitzel (approved):

PR #1109, opened 2019-05-29
Add `order` field to `allProjects` and `allEnvironments` GraphQL API endpoints

### LINKED ISSUE #1038: Add the ability to sort environments by last updated
I know sorting/filtering is a GraphQL grey area...

But we have a use case where being able to retrieve a list of all environments, sorted from oldest to newest by `updated` would allow us to ensure that the oldest sites have been re-deployed first (and sites that have been deployed already aren't unncessarily deployed if we set a limit on redeployments)

Is it possible to modify the GraphQL API to give us a last_updated option we can set to asc/desc/null to sort or add a AllEnvironmentsByLastUpdated option?  Alternatively we could implement a date filter for updated and provide a date/time to filter on?

alternatively you could implement sorting/filtering on any field in the API 😀!

cdchris12:
# Description
This PR adds the ability for:
*  The output from the `allProjects` endpoint to be ordered by name or creation date
*  The output from the `allEnvironments` endpoint to be ordered by name or last updated date

# Changelog Entry
Improvement - Add `order` field to `allProjects` and `allEnvironments` GraphQL API endpoints (#1038)

# Closing Notes
Closes #1038

### DIFF
--- a/services/api/src/resources/environment/resolvers.js
+++ b/services/api/src/resources/environment/resolvers.js
@@ -658,8 +658,12 @@ const getAllEnvironments = async (
     'deleted = "0000-00-00 00:00:00"',
   ]);
 
-  const prep = prepare(sqlClient, `SELECT * FROM environment ${where}`);
+  const order = args.order ? ` ORDER BY ${R.toLower(args.order)} ASC` : ''
+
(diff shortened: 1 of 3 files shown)


rocketeerbkw (approved):
👍

PR #1133, opened 2019-07-02
Updating env vars from varchar(300) to text

cdchris12:
This change converts the environment variable storing mechanism from a `varchar(300)` store to a simple `text` store. This will enable much larger environment variable to be defined when using Lagoon.

# Changelog Entry
Improvement - Enable much larger environment variable definitions

### DIFF
--- a/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
+++ b/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
@@ -646,20 +646,21 @@ CREATE OR REPLACE PROCEDURE
 $$
 
 CREATE OR REPLACE PROCEDURE
-  update_env_vars_from_varchar_to_text()
+  convert_env_vars_from_varchar_to_text()
 
   BEGIN
-    IF NOT EXISTS (
-      SELECT NULL
-      FROM INFORMATION_SCHEMA.COLUMNS
-      WHERE
-        table_name = 'env_vars'
-        AND table_schema = 'infrastructure'
-        AND column_name = 'value'
-    ) THEN
+    DECLARE column_type varchar(50);
+
+    SELECT DATA_TYPE INTO column_type
+    FROM INFORMATION_SCHEMA.COLUMNS
+    WHERE
+      table_name = 'env_vars'
+      AND table_schema = 'infrastructure'
+      AND column_name = 'value';
+
+    IF (column_type = 'varchar') THEN
       ALTER TABLE `env_vars`
-      MODIFY `value` TEXT NOT NULL;
-      UPDATE env_vars
+      MODIFY `value` text NOT NULL;
     END IF;
   END;
 $$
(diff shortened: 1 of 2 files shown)


services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -645,6 +645,25 @@ CREATE OR REPLACE PROCEDURE
   END;
 $$
 
+CREATE OR REPLACE PROCEDURE
+  update_env_vars_from_varchar_to_text()
+
+  BEGIN
+    IF NOT EXISTS (
```
Schnitzel: this `if not exist` means that it will check if it exist and if not it will runt he `ALTER TABLE`.
In your case though it will always already exist, therefore the alter table code will never run :)

see https://github.com/amazeeio/lagoon/pull/1133/files#diff-b44d678a275f87787a48186fffc33465R584 for an example on how to change an existing field

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -645,6 +645,25 @@ CREATE OR REPLACE PROCEDURE
   END;
 $$
 
+CREATE OR REPLACE PROCEDURE
+  update_env_vars_from_varchar_to_text()
```
Schnitzel: let's use `convert_` instead of `update_` as other similar procedure are already using `convert`: `convert_task_command_to_text`

Schnitzel (changes requested):

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
cdchris12: Changed.

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
cdchris12: Resolved in https://github.com/amazeeio/lagoon/pull/1133/commits/afcc2c61da5335dde00ff8e9ecf55932dbe03e74 .

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -645,6 +645,26 @@ CREATE OR REPLACE PROCEDURE
   END;
 $$
 
+CREATE OR REPLACE PROCEDURE
+  convert_env_vars_from_varchar_to_text()
+
+  BEGIN
+    DECLARE column_type varchar(50);
```
Schnitzel: just a nit-picking one, can you use
```suggestion
    DECLARE column_type varchar(300);
```
so that is is exactly to what you expect the column_type to be :)

Schnitzel (changes requested):

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
cdchris12: As far as I understand it, this local variable is being used to store the string which describes the type of the `value` column. That type string should fit easily within 40 chars, but there's no harm in giving it a 300 character limit, either. :)

PR #1136, opened 2019-07-03
Updating Makefile and php Dockerfiles to pin correct Apline version for php 7.0

### LINKED ISSUE #1111: remove "curl=7.61.1-r2" restriction now hirak/prestissimo has been updated
How that hirak/prestissimo has been updated to 3.9 (fixing the curl issue reported in #962) - https://github.com/hirak/prestissimo/compare/0.3.8...0.3.9

Can the version limit on curl be lifted?
https://github.com/amazeeio/lagoon/blob/master/images/php/fpm/Dockerfile#L40

Ideally if we could bring it to `7.64.0-r1` (as per Alpine 3.9) - to address high/medium outstanding CVEs with 7.61.1-r2.

I've tested locally by forcing a curl update prior to composer install using the 0.23.1 php-cli image (now with prestissimo 3.9) and have seen no issues with prestissimo or curl and drupal/scaffold

cdchris12:
Updating makefile and php Dockerfiles to pin correct apline version for php 7.0. Upstream naming has changed from `php:7.0-fpm-alpine` to `php:7-fpm-alpine`, so our current version of alpine in the PHP containers is about 6 months behind where it should be.

# Changelog Entry
Change - Updated makefile and php Dockerfiles to pin correct apline version for php 7.0

# Closing issues
Closes #1111


cdchris12:
@Schnitzel The tests pass, and this one should be good to merge now. :)

images/php/fpm/Dockerfile
```diff
@@ -36,9 +37,18 @@ COPY ssmtp.conf /etc/ssmtp/ssmtp.conf
...
 
-RUN apk del --no-cache curl libcurl \
-    && apk add --no-cache "curl=7.61.1-r2" "libcurl=7.61.1-r2" --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main/ \
-    && apk add --no-cache fcgi \
+RUN apk del --no-cache curl libcurl
+
+# PHP 5.6 and 7.0 are officially EOL as of January, 2019.
+# This command below pins the last known good curl packages for each version
+RUN if [ ${PHP_VERSION%.*} == "7.3" ] ; then \
+        apk add --no-cache "curl=7.65.1-r0" "libcurl=7.65.1-r0" --repository http://dl-cdn.alpinelinux.org/alpine/v3.10/main/ ; \
```
Schnitzel: do we need to pin a specific version here? couldn't we just install the newest version? I would like to pin as few packages as possible.

images/php/fpm/Dockerfile
```diff
@@ -36,9 +37,18 @@ COPY ssmtp.conf /etc/ssmtp/ssmtp.conf
...
+    elif [ ${PHP_VERSION%.*.*} == "7" ] ; then \
+        apk add --no-cache "curl=7.65.1-r0" "libcurl=7.65.1-r0" --repository http://dl-cdn.alpinelinux.org/alpine/v3.10/main/ ; \
```
Schnitzel: same question regarding pinning as in php 7.3

images/php/fpm/Dockerfile
```diff
@@ -36,9 +37,18 @@ COPY ssmtp.conf /etc/ssmtp/ssmtp.conf
 
 ENV NEWRELIC_VERSION=8.6.0.238
 
-RUN apk del --no-cache curl libcurl \
-    && apk add --no-cache "curl=7.61.1-r2" "libcurl=7.61.1-r2" --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main/ \
-    && apk add --no-cache fcgi \
+RUN apk del --no-cache curl libcurl
```
Schnitzel: maybe we run the `apk del` only for php 5 and leave it installed for the other ones?

Schnitzel (changes requested):

images/php/fpm/Dockerfile
cdchris12: Per our conversation in Zoom, I'm going to remove our curl pinning entirely.

PR #1147, opened 2019-07-10
Adding the DeleteSshKeyById API endpoint.

### LINKED ISSUE #1145: Add `deleteSSHKeyById` API Route
Our current `deleteSSHKey` relies on deleting a specific SSH key through referencing it by name, which is a non-unique field according to the docs (https://lagoon.readthedocs.io/en/latest/administering_lagoon/graphql_api/)

```
mutation {
  addSshKey(
    input: {
      # TODO: Fill in the name field
      # This is a non-unique identifier for the SSH key
      name: ""
      # TODO: Fill in the keyValue field
      # This is the actual SSH public key (without the type at the beginning and without the comment at the end, ex. `AAAAB3NzaC1yc2EAAAADAQ...3QjzIOtdQERGZuMsi0p`)
      keyValue: ""
      # TODO: Fill in the keyType field
      # Valid values are either SSH_RSA or SSH_ED25519.
      keyType: SSH_RSA
      # TODO: Fill in the userId field
      # This is the user ID that we noted from the addUser query
      userId: 0
    }
  ) {
    id
  }
}
```

We should create a new endpoint (or possibly update the current `deleteSSHKey` endpoint) such that we could use a unique identifier (the ssh key's `id` integer) to specify which SSH key we are deleting.

cdchris12:
This PR adds the `DeleteSshKeyById` GrpahQL API endpoint, which should help us ensure we are deleting the correct SSH key, as multiple SSH keys can possibly share the same key name.

# Changelog Entry
Improvement - Adds the `DeleteSshKeyById` GraphQL API route.

# Closing issues
Closes #1145


Schnitzel:
this will need some additional work for RBAC, pushing to 1.1.0

Schnitzel:
@rocketeerbkw please check for RBAC additions.

PR #1151, opened 2019-07-12
Added lagoonenvclean.sh script

### LINKED ISSUE #1150: Lagoon environment cleanup based on Github PR
This script will help to keep Lagoon environments clean, by removing old environments related to Pull Requests in `closed` state.

The script accept 2 parameters: name of the customer and number of months to chec "last update date" of an environment (and PR).
The idea is to develop the script and make it as much "project-wide" as possible

vincenzodnp:
# Checklist

This script helps to keep as clean as possible customer environments, by deleting Lagoon environments related to Github (for the moment only supported) Pull Requests in `closed` state.
It means script will check, for each environment named as `pr-`, if the related PR is closed and when it was LastUpdated. If it is closed and environment `lastupdate` is older than number of months passed as second script parameter, script deletes the environment.

# Changelog Entry

Improvement - Script to keep clean customer environment (#1150)

# Closing issues
closes #1150


helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
...
+# CLEANDATE: environments older than that date will be remove if related to a closed PR
+# ENDPOINT: Lagoon API endpoint to query
+# GHENDPOINT: GitHub API endpoint
+# AMZBEARERTKN: Lagoon API token
+# GHUTKN: GitHub API Authentication token
+
+CLEANDATE=$(date +%Y%m%d -d "-$MONTHS months")
+ENDPOINT="https://api-lagoon-master.lagoon.ch.amazee.io/graphql"
+GHENDPOINT="https://api.github.com"
+AMZTKN=""
```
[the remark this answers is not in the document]
Schnitzel: also do we expect that users of this script go into here and add their token? Can we make the functionality in a way that users can define the token via an actual environment variable?

like:
```
LAGOON_API_TOKEN=xxxxx
./lagoonenvclean customer number_of_months yes|no 
```

helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
+#!/bin/bash
+#
+# Author: Vincenzo De Naro Papa
+# Email: vincenzo.denaropapa@amazee.io
+#
+# Description: Run a Lagoon old environments cleanup based on closed GitHub's PRs
+# Usage: ./lagoonenvclean customer number_of_months yes|no
```
Schnitzel: the dryrun is `false` or `true`

helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
...
+lagoonenv_clean() {
+
+	# Substitue and escape characters in GraphQL query
+	ENVDELETE=$(echo $QLENVDELETE | sed "s/ENV/\"$1\"/g" | sed "s/PROJECT/\"$2\"/g" | sed "s/BOOLEAN/\"$3\"/g" | sed 's/"/\\"/g' | sed 's/\\n/\\\\n/g' | awk -F'\n' '{if(NR == 1) {printf $0} else {printf "\\n"$0}}')
+	if [ "$3" = "false" ]; then
+		echo "Dry-run env delete"
+		echo $ENVDELETE
+	else
+		echo "Deleting the $1 env of $2 project"
+		echo $ENVDELETE
```
Schnitzel: I would expect that this actually deletes the environment? instead it just shows the command like in the dry run?

helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
...
+
+QLENVSQUERY="query envbyproject { projectByName (name: \"PROJECT\") { gitUrl, id, environments { name, openshiftProjectName } }}"
+
+QLENVQUERY="query envbyname { environmentByName (name: NAME, project: ID) { updated }}"
+
+QLENVDELETE="mutation delenv { deleteEnvironment(input: {name: ENV, project: PROJECT, execute: BOOLEAN})}"
+
+# Function to retrieve customer's projects
+allproject_query() {
+	PROJECTS=`curl -s -k -X POST -H 'Content-Type: application/json' -H "$AMZBEARERTKN" $ENDPOINT -d "{\"query\": \"$PROJECTQUERY\"}" | jq -r '.[].customerByName.projects[].name'`
```
Schnitzel: no backticks subshells please, they are not safe

helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
...
+	echo "Lagoon $1 project"
+
+	# Substitute variables and escape characters in the query
+	ENVQUERY=$(echo $QLENVSQUERY | sed "s/PROJECT/$1/g" | sed 's/"/\\"/g' | sed 's/\\n/\\\\n/g' | awk -F'\n' '{if(NR == 1) {printf $0} else {printf "\\n"$0}}')
+
+	# Run the query
+	RESULT=`curl -s -k -X POST -H 'Content-Type: application/json' -H "$AMZBEARERTKN" $ENDPOINT -d "{\"query\": \"$ENVQUERY\"}"`
+
+	# From the results, take only environemnts called with a "pr-" to identify Pull Requests
+	ENVS=`echo $RESULT | jq -r '.[].projectByName.environments[].name'|grep -E -i "^(pr-)"`
```
Schnitzel: there is an actual field `deployType` that tells you if the environment is a PR or a Branch deployment, let's use that one instead of relying on the name (which could be different in the future)

helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
...
+		echo "Dry-run env delete"
+		echo $ENVDELETE
+	else
+		echo "Deleting the $1 env of $2 project"
+		echo $ENVDELETE
+	fi
+}
+
+# Function to query GitHub PR's status
+ghpr_query() {
```
Schnitzel: this function does not only query the github API, but it actually runs `lagoonenv_clean`, therefore I would suggest to call the function something that actually suggest that the clean will happen in here

helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
...
+# AMZBEARERTKN: Lagoon API token
+# GHUTKN: GitHub API Authentication token
+
+CLEANDATE=$(date +%Y%m%d -d "-$MONTHS months")
+ENDPOINT="https://api-lagoon-master.lagoon.ch.amazee.io/graphql"
+GHENDPOINT="https://api.github.com"
+AMZTKN=""
+AMZBEARERTKN="Authorization: bearer $AMZTKN"
+GHTKN=""
+GHBEARERTKN="Authorization: bearer $GHTKN"
```
Schnitzel: I would expect some kind of explanation what type of Github Token is actually needed (which permissions does the token need)

helpers/lagoonenvclean.sh
```diff
@@ -0,0 +1,182 @@
...
+		exit 1
+	fi
+	allproject_query $1
+	for i in $PROJECTS
+	do
+		allenvironment_query $i
+		for env in ${LAGOONENVS[$i,envs]}
+		do
+			id=${LAGOONENVS[$i,id]}
+			environment_query "$env" $id
```
Schnitzel: the result of `environment_query` does not seem to be used anywhere?

Schnitzel (changes requested):

helpers/lagoonenvclean.sh
vincenzodnp: I added in the body of the script what kind of permissions are needed.

helpers/lagoonenvclean.sh
vincenzodnp: Perfect thanks! I changed the logic and I'm using that

helpers/lagoonenvclean.sh
vincenzodnp: I added an option to run the script in dry-run mode. By default it runs no in dry-run and it removes envs

helpers/lagoonenvclean.sh
vincenzodnp: Removed

shreddedbacon:
@vincenzodnp this is not RBAC compliant, can you re-work it taking into consideration that customer object no longer exists

seanhamlin:
I like the idea, I wonder how much of an issue this would be in reality? When you close a PR, the webhook should inform Lagoon to delete the environment. I guess there could be a case when this does not happen for what ever reason, and you want to manually tidy this up.

Some suggestions I would have

* Does this script work for branches as well as PR environments? If not, can it be made to support both? This would make it more useful. Either way, the docs at the top should make this clear.
* Test it on Mac OSX, I know there will be issues with `date`, you need to use `gdate` on Mac OSX for the full range of GNU functions.
* Rename the script to `lagoon-delete-from-removed-github-branches.sh` or similar to ensure that it is known immediately that only Github is supported (and not say Gitlab, or Bitcucket).
* Wrap comments to be 80 characters, at the moment you need to horizontally scroll.

vincenzodnp:
> I like the idea, I wonder how much of an issue this would be in reality? When you close a PR, the webhook should inform Lagoon to delete the environment. I guess there could be a case when this does not happen for what ever reason, and you want to manually tidy this up.

This is one case. For example if you didn't set the webhook to act when a PR is created/closed.

> Some suggestions I would have
> 
> * Does this script work for branches as well as PR environments? If not, can it be made to support both? This would make it more useful. Either way, the docs at the top should make this clear.

The script checks if a PR is closed or if it is older than a date you set. Let it working for branch might be a bit tricky because, AFAIK, you cannot "close" a branch.

> * Test it on Mac OSX, I know there will be issues with `date`, you need to use `gdate` on Mac OSX for the full range of GNU functions.

Added a workaround to use `gdate` if exist instead of `date`

> * Rename the script to `lagoon-delete-from-removed-github-branches.sh` or similar to ensure that it is known immediately that only Github is supported (and not say Gitlab, or Bitcucket).

Well the goal is to support other systems, so I will keep it as much general as possible.

> * Wrap comments to be 80 characters, at the moment you need to horizontally scroll.

I will do all I can...but not promise :smile:

vincenzodnp:
@Schnitzel @seanhamlin can you review it?
Thanks

vincenzodnp:
@tobybellwood @Schnitzel any news for that PR?

tobybellwood:
This should be resubmitted against https://github.com/amazeeio/helper-scripts

PR #1158, opened 2019-07-24
Update version of oc used in oc image

### LINKED ISSUE #1083: Un-idle targets in service auto idler is not working correctly
The `idle-services.sh` script currently sets all the un-idle targets for services so that services un-idle at the same time, but it does not appear to work correctly.

https://github.com/amazeeio/lagoon/blob/master/services/auto-idler/idle-services.sh#L95-L102

When an endpoint is requested, only that service is un-idled initially before the other endpoints are requested.

shreddedbacon:
# Checklist

The version of oc (3.7.2) used in the oc image has an issue when updating annotations that contain a `/` in them, version `3.9.0` contains a fix for this which will resolve issue #1083 when trying to set `idling.alpha.openshift.io/unidle-targets=..`

# Changelog Entry
Change - Update oc version in oc image from `3.7.2` to `3.9.0`

# Closing issues
closes #1083


Schnitzel:
@shreddedbacon 
so this fails here: https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/scripts/exec-openshift-create-route.sh#L5
Logs:
```
[35;1mopenshiftbuilddeploymonitor_1    |[0m +++ oc --insecure-skip-tls-verify -n ci-node-node10 patch route customdomain-will-be-main-domain.com -p '{"metadata":{"annotations":{"kubernetes.io/tls-acme":"true","haproxy.router.openshift.io/hsts_header":"null"}},"spec":{"to":{"name":"node"},"tls":{"insecureEdgeTerminationPolicy":"Redirect"}}}'
[35;1mopenshiftbuilddeploymonitor_1    |[0m route "customdomain-will-be-main-domain.com" not patched
```
the problem is that `oc patch` in version 3.7 does return a shell returncode of `0` if a patch doesn't change anything, now in 3.9 `oc patch` returns with returncode of non-0 if there is nothing to patch.
So we should find a way that `oc patch` does not fail with non-0 if there was nothing to patch, or find a way that the bash script can catch this.

shreddedbacon:
Ahh, bugger. I'll see what I can figure out

tobybellwood:
It's fixed upstream in openshift 3.11 https://github.com/openshift/origin/pull/20665 - can we use that in testing - or do we need to use 3.9?

shreddedbacon:
Maybe we can use 3.11, will try it out

PR #1162, opened 2019-07-31
Disable PHP notices on production envs by default

cdchris12:
This commit disables PHP notices by default on all production environments so as to prevent those notices from being logged to the watchdog process.

This behavior can be overridden by setting the environment variable `production_notices` to `true`.

# Changelog Entry
Change - Disable PHP notices on production environments by default


images/php/fpm/entrypoints/51-production-detection.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [[ ${LAGOON_ENVIRONMENT_TYPE} == "production" && ( -z "${production_notices}" || $production_notices == "true" ) ]]; then
```
Schnitzel: what is `production_notices` and how is it defined?

Schnitzel: also can we change the script so that if somebody wants to overwrite `PHP_ERROR_REPORTING` from the outside via an environment variable that this is possible? Currently it's not possible to set it, as it will be set by the entry point script all the time

Schnitzel (changes requested):

images/php/fpm/entrypoints/51-production-detection.sh
cdchris12: `production_notices` is just an env var that can be defined to enable notices on a production environment. 

I can add a customer supplied `PHP_ERROR_REPORTING` variable as well, sure.

cdchris12: Added logic to accept a customer supplied value for `PHP_ERROR_REPORTING`. If that value is not set, then we set the value automatically based on production/development status.

docs/using_lagoon/environment_variables.md
```diff
@@ -78,3 +78,13 @@ If you want to define environment variables different per environment you can cr
 ### `.env` and `.env.defaults`
 `.env` and `.env.defaults` will act as the default values for environment variables if none other is defined. For example
 as default environment variables for Pull-Request environments (see [Worfklows](./workflows.md#pull-requests)).
+
+## Special Environment Variables
+
+### `PHP_ERROR_REPORTING`
```
Schnitzel, line 84: @cdchris12 
can you please add this documentation here: https://lagoon.readthedocs.io/en/latest/using_lagoon/docker_images/php-fpm/

images/php/fpm/entrypoints/51-production-detection.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+if [[ -z "${PHP_ERROR_REPORTING}" ]]; then
+    if [[ ${LAGOON_ENVIRONMENT_TYPE} == "production" ]]; then
+        export PHP_ERROR_REPORTING="E_ALL & ~E_DEPRECATED & ~E_STRICT & ~E_NOTICE"
+    else
```
Schnitzel: I wonder if we need this else, as we already have defaults in `00-lagoon-php.ini.tpl` that are the same as in the else I think this is not really necessary?

Schnitzel (changes requested):

Schnitzel:
@cdchris12 can you have a look at my requests?

cdchris12:
Sorry about that; somehow missed the notification of your previous comments. I'll get them resolved on Monday. :)

images/php/fpm/entrypoints/51-production-detection.sh
cdchris12: Agreed. Resolved in commit 9ea8185e92817094bef86cd67eb2ac2664d7b407

docs/using_lagoon/environment_variables.md
cdchris12: Added in d9c6f3616230e85408c79cfc8fc1692c44cf9ede . Would you like to remove this block from `docs/using_lagoon/environment_variables.md` as well? @Schnitzel

PR #1178, opened 2019-08-13
Implement Native Cronjob Removal

cdchris12:
# Checklist

Currently, native cronjobs are never removed from the system. This PR implements the logic necessary to detect and remove cronjobs which are no longer defined in `.lagoon.yml` as well as cronjobs which have been modified so as to no longer be ran natively.

# Changelog Entry

Improvement - Implement Native Cronjob Removal


tests/files/features/.lagoon.yml.test
```diff
@@ -0,0 +1,53 @@
...
+  git_sha: 'true'
+
+environments:
+  branch/routes:
+    routes:
+    - node:
+      - customdomain-will-be-main-domain.com
+      - customdomain-will-be-not-be-main-domain.com
+  branch/cronjob:
+    cronjobs:
```
Schnitzel: can we remove all cronjobs here? with that we also make sure that mulitple cronjobs can also be removed

tests/files/features/.lagoon.yml
```diff
@@ -35,6 +35,10 @@ environments:
       schedule: "H * * * *"
       command: printf '%s_%s' "CRONJOB" "H" >> /files/cron.txt
       service: node
+    - name: test removal
+      schedule: "* * * * *"
```
Schnitzel: this will not create a native cronjob object, but instead an inside-pod-cronjob which we know they work?
see https://github.com/amazeeio/lagoon/pull/1178/files#diff-f5443eafb5de7b143fbba445a79a5ee8R31 how to force a native cronjob

tests/files/features/.lagoon.yml.test
```diff
@@ -0,0 +1,53 @@
+docker-compose-yaml: docker-compose.yml
```
Schnitzel: can we call this file something else than `.lagoon.yml.test`? give it a name like `.lagoon-no-cronjobs.yml`

tests/files/features/.lagoon.yml.test
```diff
@@ -0,0 +1,53 @@
...
+      service: node
+    - name: test H
+      schedule: "H * * * *"
+      command: printf '%s_%s' "CRONJOB" "H" >> /files/cron.txt
+      service: node
+  branch/env-specific-template:
+    templates:
+      node: .lagoon.node-env-specific.yml
+
+ci-multiproject1:
```
Schnitzel: this also can be removed, as we don't use this at all for a deployment

tests/tasks/webhook-gitlab/git-update-lagoon-commit-push.yaml
```diff
@@ -0,0 +1,38 @@
+- name: "{{ testname }} - make sure we are on the branch {{ branch }}"
+  shell: git checkout -b {{ branch }} || git checkout {{ branch }}
+  args:
+    chdir: /{{git_repo_name}}
+
+#This is where we need to move the .lagoon.yml.test over the .lagoon.yml file
```
Schnitzel: let's ass the move functionaly to `git-add-commit-push.yaml` like we already do for the `docker_files` variable, we can have a functionality which will take a filename (like `.lagoon-no-cronjobs.yml`) as an argument, and if the variable is filled ansible will automatically rename it for you.

tests/tests/features/cronjobs.yaml
```diff
@@ -62,6 +62,24 @@
...
 
+# This is where we need to modify the lagoon versions and do another deploy
+- name: "{{ testname }} - change .lagoon.yml to remove cron, git push"
+  hosts: localhost
+  serial: 1
+  vars:
+    git_files: "features/"
+  tasks:
+  - include: ../../tasks/git-update-lagoon-commit-push.yaml
+
```
Schnitzel, line 92: after changing the code, you need to trigger a deployment first via `../../tasks/api/deploy-no-sha.yaml`, see https://github.com/amazeeio/lagoon/pull/1178/files#diff-22bec30a3d720f348368edc7531a31b3R11-R18

tests/files/features/.lagoon.yml
Schnitzel: actually let's test both: removal for native and inside-pod cronjobs

Schnitzel (changes requested):

tests/files/features/.lagoon.yml.test
cdchris12: It won't be as easy as simply removing the cronjobs, due to the way the node test works. It just dumps the contents of the `files/cron.txt` into its HTTP response. Let me sleep on it and I'm sure I can think of something to test both of these cases.

tests/files/features/.lagoon.yml
cdchris12: Whoops; I got the native vs in pod time limits backwards. I'll fix that on the next update.

tests/tasks/webhook-gitlab/git-update-lagoon-commit-push.yaml
cdchris12: Done in 7e4bba47

tests/files/features/.lagoon.yml.test
cdchris12: Done in 7e4bba47

tests/files/features/.lagoon.yml.test
cdchris12: Done in 7e4bba47

tests/files/features/.lagoon.yml.test
cdchris12: Renamed as suggested in 7e4bba47

tests/tasks/git-add-commit-push.yaml
```diff
@@ -14,6 +14,12 @@
     dest: /{{git_repo_name}}/
   when: docker_files is defined
 
+- name: "{{ testname }} - removing cronjobs from .lagoon.yml"
+  copy:
+    src: /{{git_repo_name}}/{{lagoon_no_crons}}
+    dest: /{{git_repo_name}}/.lagoon.yml
+  when: lagoon_no_crons is defined
```
Schnitzel: let's call this variable `overwrite_lagoon_yml` as then it will be much more general and can be used by all kind of tests to overwrite the `.lagoon.yml` and not only for crons

Schnitzel (changes requested):

tests/tasks/git-add-commit-push.yaml
cdchris12: Resolved in c40d7205f6d88c081ac99c50eb965ae789810322

PR #1190, opened 2019-08-16
Update Dockerfile naming convention to <purpose>.dockerfile

dan2k3k4:
To improve syntax highlighting we can use the following naming scheme for `Dockerfile`s: `<purpose>.dockerfile` such as:
- db.dockerfile
- php.dockerfile
- nginx.dockerfile

The `docker-compose.yml` files have been updated, along with the sample Dockerfiles and documentation.


dan2k3k4:
Switched PR from using `dan2k3k4:master` branch to `dan2k3k4:update-dockerfile-naming` branch as it's bad-practice to create PRs from default branches.

AlexSkrypnyk:
This change is going to break upgradability of so many websites. IDE should support configuration of file types by mask, e.g. `Dockerfile.*`

simesy:
@alexdesignworks I can't think of what will break in an existing project.
References from your docker-compose.yml point to your own Dockerfiles, not Lagoon Dockerfiles. Otherwise they point directly to images, and not be effected by this change.

simesy:
Do you just mean if you are merging changes from lagoon Dockerfiles upstream into your project?

dan2k3k4:
@alexdesignworks the changes only concern the `/docs` folder and the `/tests` folder - both with updated `docker-composer.yml` files to point to the updated `Dockerfile` files.

---

Here's a little more background.

In multi-stage builds, we tend to require multiple `Dockerfile` files. Some organisations opt for separating these into their own folders such as:
```
- /root
-- /docker
--- /cli
---- Dockerfile
--- /db
---- Dockerfile
--- /nginx
---- Dockerfile
```

This is OK, but can be difficult to find "the right file" you want to edit in an IDE as you have to rely on being able to see the path name.

Another solution is to add the context to the Dockerfile name, such as for the extension:
```
- /root
-- /docker
--- Dockerfile.cli
--- Dockerfile.db
--- Dockerfile.nginx
```

But this creates another issue: we **no longer have syntax highlighting** out-of-the-box.

What if we standardise and use `<purpose>.dockerfile` as the preferred naming convention? Such as:
```
- /root
-- /docker
--- cli.dockerfile
--- db.dockerfile
--- nginx.dockerfile
```

The `.dockerfile` extension already works for syntax highlighting in `VS Code` using this [glob pattern](https://github.com/microsoft/vscode-docker/blob/b39024077508500cdcfbba469dfb7c9d413d51e6/src/constants.ts#L33)

Also, using a lowercase `.dockerfile` would match with general naming convention for other filetype extensions. I think we should avoid capitalising `.Dockerfile` but keeping `Dockerfile` for extensionless naming (e.g. in the case of only ever requiring the use of one `Dockerfile` for your project).

PR #1202, opened 2019-08-28
Adding ServerAliveInterval and ServerAliveCountMax directives to the …

cdchris12:
…PHP container's SSH config file

This change makes the PHP containers use SSH keepalive messages instead of TCP keepalive messages.

# Changelog Entry
Change: Make the PHP containers use SSH keepalive messages instead of TCP keepalive messages.


Schnitzel (changes requested):

Schnitzel (approved):

images/php/cli/ssh_config
```diff
@@ -2,4 +2,6 @@ Host *
   StrictHostKeyChecking no
   UserKnownHostsFile /dev/null
   IdentityFile /home/.ssh/lagoon_cli.key
-  IdentityFile /home/.ssh/key
\ No newline at end of file
+  IdentityFile /home/.ssh/key
+  ServerAliveInterval 60
+  ServerAliveCountMax 1440 # max keepalive of 24h
```
Schnitzel:
```suggestion
  ServerAliveCountMax 1440
```
ssh_config does not support comments at the end of a config line, and fails with
```
/etc/ssh/ssh_config line 7: garbage at end of line; "#"
```

Schnitzel (changes requested):

Issue #1157, opened 2019-07-22
Yarn builds fail on Node version 12


cdchris12:
**Describe the bug**
When I attempted to run `yarn upgrade <package>`, all the resulting yarn image builds failed using Node version 12. In order to make those tests complete, I had to install Node 10.

**To Reproduce**
1. Install Node Version 12 and Yarn
2. Try to remove a package with `yarn remove git-url-parse`
3. See (many) error(s)

**Expected behavior**
We should either rework these builds to pass with Node versions > 10, or should force yarn to error out on any node version other than 10

dasrecht:
Hi @cdchris12 - During the work on #1215 i saw that we're not having node12 built yet - was that something you tested locally?

dasrecht:
https://github.com/amazeeio/lagoon/compare/node_12?expand=1

tobybellwood:
This is being fixed in main - the mariasql library update was first, then the node update to 14/16

PR #1216, opened 2019-09-02
Node 12

cdchris12:
This commit enables the use of Node version 12 containers.

# Changelog Entry
Feature - Added containers for Node version 12

Possibly related to #1157


Schnitzel:
can we also get some tests for node12? looks like currently we only test node6 and node8, so we should probably add test for 10 and 12 :)

Schnitzel:
also let's switch to

PR #1226, opened 2019-09-05
Enable RabbitMQ mirrored queues

### LINKED ISSUE #1098: Set RabbitMQ queues in HA mode (Mirrored queues)
**Describe the bug**
"By default, contents of a queue within a RabbitMQ cluster are located on a single node (the node on which the queue was declared)".  
If for some reason, the pod where the queue is declared, dies, the RabbitMQ service become unavailable.

https://www.rabbitmq.com/ha.html
https://www.rabbitmq.com/clustering.html

**To Reproduce**
At the moment it is the default behavior

**Expected behavior**
Queues can be served by every pod of the `broker` Stateful Set.

vincenzodnp:
# Checklist

Currently the RabbitMQ cluster is not set to use native HA mechanism called **"Mirrored Queues"**. 
 
It allows to create queues mirrored on 2 or more nodes, making the cluster fully High Availability.   
In fact, at the moment, if the node (pod) where the queue is declared dies, queue's consumer are not able to use the it.   

With mirrored queues, not only queue is still available for consumers but the mirror node is automatically elected new queue master (writable).   

When the mirror node(s) is(are) online again, the queue is automatically synced all over the nodes defined in the policy.

# Changelog Entry

Improvement - Set RabbitMQ policy to enable Mirrored Queue by default for all "lagoon-" queues  (#1098 )

# Closing issues
closes #1098


images/rabbitmq/Dockerfile
```diff
@@ -3,9 +3,20 @@ FROM rabbitmq:3-management
...
 RUN rabbitmq-plugins enable --offline rabbitmq_delayed_message_exchange;
 
-ENV RABBITMQ_DEFAULT_USER=guest \
-    RABBITMQ_DEFAULT_PASS="guest"
+# Copy startup schema with vhost, users, permissions and policies
+COPY definitions.json /etc/rabbitmq/definitions.json
+RUN chgrp 0 /etc/rabbitmq/definitions.json; chmod g+rw /etc/rabbitmq/definitions.json
+
+# Copy a custom entrypoint
+ADD cluster-rabbit.sh /
```
Schnitzel: let's use `COPY` here as we use everywhere else.

images/rabbitmq/Dockerfile
```diff
@@ -3,9 +3,20 @@ FROM rabbitmq:3-management
 ARG LAGOON_VERSION
 ENV LAGOON_VERSION=$LAGOON_VERSION
 
-COPY rabbitmq_delayed_message_exchange-3.7.0.ez /plugins
+ENV RABBITMQ_DEFAULT_USER=guest \
+    RABBITMQ_DEFAULT_PASS="guest"\
+    RABBITMQ_DEFAULT_HA_PATTERN='^lagoon-'
```
Schnitzel: let's use as default `^$`, aka a pattern that does not match.

services/broker/.lagoon.app.yml
```diff
@@ -120,6 +120,8 @@ objects:
                   key: RABBITMQ_PASSWORD
             - name: RABBITMQ_USE_LONGNAME
               value: "true"
+            - name: RABBITMQ_DEFAULT_HA_PATTERN
+              value: "^lagoon-"
```
Schnitzel: let's add this to the `services/broker/Dockerfile` instead of into the broker statefulset, as then it is part of the Docker Image already.

images/rabbitmq-cluster/cluster-rabbit.sh
```diff
@@ -13,5 +13,12 @@ echo NODENAME=rabbit@${HOSTNAME}.${SERVICE_NAME}-headless.${POD_NAMESPACE}.svc.c
 echo cluster_formation.k8s.hostname_suffix=.${SERVICE_NAME}-headless.${POD_NAMESPACE}.svc.cluster.local >> /etc/rabbitmq/rabbitmq.conf
 echo cluster_formation.k8s.service_name=${SERVICE_NAME}-headless >> /etc/rabbitmq/rabbitmq.conf
 
+# Replace placeholders with ENV values and copy file to rabbitmq conf directory
+if [ -e /etc/rabbitmq/definitions.json ]; then
+  sed -i -e "s/REPLACE_USERNAME/${RABBITMQ_DEFAULT_USER}/g" \
```
Schnitzel: we already use envplate: https://github.com/amazeeio/lagoon/blob/484d8c9ecf6b227587a015bb96d25398ca4a9500/images/commons/Dockerfile#L16 example: https://github.com/amazeeio/lagoon/blob/master/images/php/fpm/entrypoints/70-php-config.sh#L4 for needs of replacing env variables inside files. Let's use the same system also here, so we are consistent across Lagoon.

images/rabbitmq/definitions.json
```diff
@@ -0,0 +1,14 @@
+{
+  "vhosts":[
+        {"name": "/"}
```
Schnitzel: let's make this also into an Env Variable.

Schnitzel (changes requested):

images/rabbitmq/Dockerfile
vincenzodnp: Done.

images/rabbitmq/Dockerfile
vincenzodnp: I need to test the behavior of such pattern, because it could match ALL queues, either temporary queues.

services/broker/.lagoon.app.yml
vincenzodnp: So you prefer users let override the variable during the build of the image instead during the creation of StatefulSet. is it correct?

images/rabbitmq/definitions.json
vincenzodnp: I agree.

vincenzodnp: Shall we create other Env Variables like the name of the policy or perhaps the ha-mode (es: 'all')?

Schnitzel: no, if somebody want's to change that deep I would suggest they update the whole `definitions.json` file themselves :)

images/rabbitmq/Dockerfile
vincenzodnp: Tested and the pattern is not matched by default by any queue

images/rabbitmq-cluster/cluster-rabbit.sh
vincenzodnp: Added envplate binary and set variables in definitions.json

services/broker/.lagoon.app.yml
Schnitzel: how users overwrite it themselves I don't care, it's the users decision. But I would like that Lagoon when it uses the base images sets the ENV variables needed for itself to correctly work inside the Service Dockerfiles and not the YAML files.

vincenzodnp: Added ENV variables to define default VHOST and default PATTERN in broker service

vincenzodnp (commented):
Pushed last commit to fix it.
Please review.
Thanks

images/rabbitmq/Dockerfile
```diff
@@ -3,9 +3,26 @@ FROM rabbitmq:3-management
 ARG LAGOON_VERSION
 ENV LAGOON_VERSION=$LAGOON_VERSION
 
-COPY rabbitmq_delayed_message_exchange-3.7.0.ez /plugins
+ENV RABBITMQ_DEFAULT_USER=guest \
```
dasrecht: can you check that the quotes are consistent. some of the env vars don't have some quotes, some singleticks and some double ticks. It would be nice when they are consistent

images/rabbitmq/Dockerfile
```diff
@@ -3,9 +3,26 @@ FROM rabbitmq:3-management
...
+    RABBITMQ_DEFAULT_PASS="guest"\
+    RABBITMQ_DEFAULT_HA_PATTERN='^$'\
+    RABBITMQ_DEFAULT_VHOST='/'
 
+RUN apt-get -y update \
+    && apt-get -y install curl \
+    && curl -sLo /bin/ep https://github.com/kreuzwerker/envplate/releases/download/1.0.0-RC1/ep-linux \
```
dasrecht: as we already have ep in our commons image we could easily take it from there that would make the code a bit cleaner

see - https://github.com/amazeeio/lagoon/blob/master/images/docker-host/Dockerfile#L13

services/broker/Dockerfile
```diff
@@ -1,2 +1,4 @@
 ARG IMAGE_REPO
 FROM ${IMAGE_REPO:-lagoon}/rabbitmq-cluster
+ENV RABBITMQ_DEFAULT_HA_PATTERN='^lagoon-'\
+    RABBITMQ_DEFAULT_VHOST='/'
```
dasrecht: The RABBITMQ_DEFAULT_VHOST is already set here - do we need to re-set it https://github.com/amazeeio/lagoon/pull/1226/files#diff-38a0607e913442c3f99084261e69cffdR9 ?

dasrecht (changes requested):

images/rabbitmq/Dockerfile
vincenzodnp: I'm choosing single ticks for all variables because better protection for special characters like `/` or `^`

images/rabbitmq/Dockerfile
vincenzodnp: Good to know! Changed to COPY from commons

services/broker/Dockerfile
vincenzodnp: Going to remove from there because, if you define a VHOST here different from `/`, you need also to change configuration of consumers.

images/rabbitmq/Dockerfile
```diff
@@ -1,11 +1,25 @@
...
+COPY --from=commons /bin/ep /bin/
 
+COPY rabbitmq_delayed_message_exchange-3.7.0.ez /plugins
 RUN rabbitmq-plugins enable --offline rabbitmq_delayed_message_exchange;
 
-ENV RABBITMQ_DEFAULT_USER=guest \
-    RABBITMQ_DEFAULT_PASS="guest"
+# Copy startup schema with vhost, users, permissions and policies
+COPY definitions.json /etc/rabbitmq/definitions.json
+RUN chgrp 0 /etc/rabbitmq/definitions.json; chmod g+rw /etc/rabbitmq/definitions.json
```
Schnitzel: this can be done with `fix-permissions`

Schnitzel (changes requested):

images/rabbitmq/Dockerfile
vincenzodnp: Changed

vincenzodnp (commented):
Ready for review

images/rabbitmq/definitions.json
```diff
@@ -0,0 +1,14 @@
...
+        {"name": "${RABBITMQ_DEFAULT_VHOST}"}
+  ],
+  "users": [
+        { "name": "${RABBITMQ_DEFAULT_USER}", "password": "${RABBITMQ_DEFAULT_PASS}", "tags": "administrator" }
+  ],
+  "permissions":[
+        { "user": "${RABBITMQ_DEFAULT_USER}", "vhost": "${RABBITMQ_DEFAULT_VHOST}", "configure": ".*", "write": ".*", "read": ".*" }
+  ],
+  "policies":[
+        {"vhost":"${RABBITMQ_DEFAULT_VHOST}","name":"lagoon-ha","pattern":"${RABBITMQ_DEFAULT_HA_PATTERN}", "definition":{"ha-mode":"exactly","ha-params":2,"ha-sync-mode":"automatic","ha-sync-batch-size":5}}
```
Schnitzel, line 12: reading https://www.rabbitmq.com/ha.html it seems that we should set `ha-sync-batch-size` to `1` as we want every single message by synced and not packs of 5.

images/rabbitmq/Dockerfile
```diff
@@ -1,11 +1,25 @@
...
 
-ENV RABBITMQ_DEFAULT_USER=guest \
-    RABBITMQ_DEFAULT_PASS="guest"
+# Copy startup schema with vhost, users, permissions and policies
+COPY definitions.json /etc/rabbitmq/definitions.json
+RUN fix-permissions /etc/rabbitmq/definitions.json
+
+# Copy a custom entrypoint
+COPY cluster-rabbit.sh /
+RUN chgrp 0 /cluster-rabbit.sh; chmod +x /cluster-rabbit.sh
```
Schnitzel: let's also use `fix-permissions` for the `chgrp 0` stuff.

Schnitzel (changes requested):

images/rabbitmq/definitions.json
vincenzodnp: In the link you sent I see: 

> Earlier versions will synchronise 1 message at a time by default. By synchronising messages in batches, the synchronisation process can be sped up considerably.

It seems that a batch-size greater than 1, increase the speed of synchronization process.

images/rabbitmq/Dockerfile
```diff
@@ -1,11 +1,25 @@
...
 
-ENV RABBITMQ_DEFAULT_USER=guest \
-    RABBITMQ_DEFAULT_PASS="guest"
+# Copy startup schema with vhost, users, permissions and policies
+COPY definitions.json /etc/rabbitmq/definitions.json
+RUN fix-permissions /etc/rabbitmq/definitions.json
+
+# Copy a custom entrypoint
+COPY cluster-rabbit.sh /
+RUN fix-permissions /cluster-rabbit.sh; chmod +x /cluster-rabbit.sh
```
Schnitzel:
```suggestion
RUN fix-permissions /cluster-rabbit.sh && chmod +x /cluster-rabbit.sh
```
let's always use `&&` instead of `;` for running multiple commands in a `RUN`, as the `;` will swallow any errors.

Schnitzel (changes requested):

vincenzodnp:
Added documentation about image

docs/using_lagoon/docker_images/rabbitmq.md
```diff
@@ -0,0 +1,34 @@
+# rabbitmq Image
+amazee.io RabbitMQ Dockerfile with management plugin installed, based on the official rabbitmq:3-management image at [docker-hub](https://hub.docker.com/_/rabbitmq).
```
Schnitzel: We're not using "amazee.io" everything is called "Lagoon", please rename

docs/using_lagoon/docker_images/rabbitmq.md
```diff
@@ -0,0 +1,34 @@
...
+Into the `definitions.json` schema file, are definied the minimal entities to let the container running: virtualhost (vhost), username and password to access management UI, permissions and policies.  
+
+By default a policy called `lagoon-ha` is created at startup but it is not active because it doesn't match any queue's name pattern (see default [Environment Variables](#environment-variables)).  
+```
+"policies":[
+        {"vhost":"${RABBITMQ_DEFAULT_VHOST}","name":"lagoon-ha","pattern":"${RABBITMQ_DEFAULT_HA_PATTERN}", "definition":{"ha-mode":"exactly","ha-params":2,"ha-sync-mode":"automatic","ha-sync-batch-size":5}}
+  ]
+```
+By default, the `ha-mode` is set to `exactly` which controls the exact number of mirroring nodes for a queue (mirrors). The number of nodes, is controller by `ha-params`.   
+For furhter and custom configuration, please refer to [official RabbitMQ documentation](https://www.rabbitmq.com/ha.html).
```
Schnitzel: typo `furhter` -> `further`

Schnitzel (changes requested):

docs/using_lagoon/docker_images/rabbitmq.md
vincenzodnp: Ok. Should I change also nxing and php readme.md pages?

Schnitzel: yes if you want, but in a separate PR

docs/using_lagoon/docker_images/rabbitmq.md
vincenzodnp: fixed

docs/using_lagoon/docker_images/rabbitmq.md
vincenzodnp: Ok

Schnitzel:
@vincenzodnp 
the testing currently fails:
```
Step 1/13 : FROM ${IMAGE_REPO:-lagoon}/commons as commons
pull access denied for lagoon/commons, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

Makefile:149: recipe for target 'build/rabbitmq' failed

make: *** [build/rabbitmq] Error 1
```
this is because it tries to load `lagoon/commons`, but the env variable `IMAGE_REPO` would overwrite this to something like `jenkinslagoonpr122616`.
We missed the `ARG` on top of the `rabbitmq` Dockerfile, see like https://github.com/amazeeio/lagoon/blob/master/images/nginx/Dockerfile#L1

Can you add that? thanks

vincenzodnp:
@Schnitzel added the `ARG IMAGE_REPO`.
Can you review?
Thanks

PR #1260, opened 2019-09-19
Allow tasks to access production via temporary maintainer tokens

### LINKED ISSUE #1229: Inject Token with Maintainer Access into Task Pods
We're running into an issue with the Task Pods not having enough permissions, the situation is this:

1. User with Developer Role creates a Task to sync production to development environment
2. Lagoon creates a Task Pod within development environment which wants to ssh into production environment
Problem is that the Task Pod is running with the `guest` role which cannot access the production environment and therefore fails.

We could solve this with giving the guest role ssh access to the production environment, but this is not the idea of RBAC. 
We could also create a new role for the default-user and give that role ssh production access, the problem is that this role will also be used by the CLI pod and there is the risk that somebody with just development access has access to the CLI pod and "steals" the credentials of the CLI pod and can access the production environment. 

Therefore our current idea to solve this would be:
When Lagoon creates the Task Pod it also gets a JWT Token from the API which has maintainer role access and is valid only for a couple of minutes. This JWT is then used by the Task Pod to authenticate itself against the SSH Service (instead of the SSH Key) and gets access to the production environment.

I think these are the needed tasks:
- [ ] Update the SSH Service so that a client has the capability to pass a specific JWT token into the Remote Shell connection which is then used to check if the user has access to an environment or not (https://github.com/amazeeio/lagoon/blob/master/services/ssh/home/rsh.sh#L35)
- [ ] Update the Task Pod Creator of Lagoon to get a JWT Token from the API with Maintainer permission and inject it into the Task Pod (The pod will then use it to get access to the production environment)

rocketeerbkw:
# Checklist

With the RBAC changes, access to the API via cli containers was given via guest permissions. This prevented the tasks from doing syncs from production environments. This PR adds a default maintainer user to every project, with no password and no ssh key (so logins with this user aren't allowed). Additionally, the api client requests a new access token for the maintainer and passes that to the task container. The ssh server was modified to allow passing a `LAGOON_TOKEN` environment variable that it will uses to check permissions.

# Changelog Entry

Improvement - Allow tasks to access production via temporary maintainer tokens (#1229)
Bugfix - Prevent lagoon images from interfering with pygmy

# Closing issues
closes #1229


Schnitzel:
@rocketeerbkw can you fix the conflict? thank you

tobybellwood:
default-uisers were given maintainer access a while back, closing!

PR #1293, opened 2019-10-14
Allow use of private docker registries

shreddedbacon:
# Checklist

This change allows for private docker registries to be used to pull pre-built images into lagoon.

# Changelog Entry

Feature - Allow the use of private docker registries to pull own images
Documentation - Configuration of private docker registries


docs/using_lagoon/lagoon_yml.md
```diff
@@ -260,3 +260,37 @@ Each definition is keyed by a unique name (`secrets` and `logs-db-secrets` in th
...
+
+To consume an image from a private registry, it needs to be used as the `FROM` in a `Dockerfile.<service>`
+
+```
+#Dockerfile.mariadb
+FROM dockerhubuser/my-private-database:tag
+```
+
+And the service updated to use build instead of image
+
```
Schnitzel, line 287: let's document here that we're talking about the `docker-compose.yml` file here

docs/using_lagoon/lagoon_yml.md
```diff
@@ -260,3 +260,37 @@ Each definition is keyed by a unique name (`secrets` and `logs-db-secrets` in th
 * `path` - the path to the yaml file
 * `command` - can either be `create` or `apply`, depending on if you like to run `kubectl create -f [yamlfile]` or `kubectl apply -f [yamlfile]`
 * `ignore_error` - either `true` or `false` (default), this allows you to instruct the lagoon build script to ignore any errors that might are returned during running the command. (This can be useful to handle the case where you want to run `create` during every build, so that eventual configurations are created, but don't want to fail if they already exist).
+
+#### `private-registries`
+
+The `private-registries` block allows you to define your own private docker registries to pull private images. If you don't specify a `url` in your yaml, it will default to using docker hub.
+
+The password can be defined as an environment variable name that is set in the Lagoon api [Environment Variables (Lagoon API) » Docker Registry Environment Variables (Lagoon API)](./environment_variables.md). If one is not set, it will default to the value defined in the `.lagoon.yml` file.
```
Schnitzel: can you try explain that a bit better? honestly I'm confused where I should define the password and where.

Schnitzel (changes requested):
just documentation review for now, didn't had time to check the code too much yet

docs/using_lagoon/lagoon_yml.md
shreddedbacon: Now that I read it back, I can see how its confusing. I'll word it better.

docs/using_lagoon/environment_variables.md
```diff
@@ -67,6 +67,27 @@ mutation addBuildtimeEnv {
...
 }
 ```
+### Docker Registry Environment Variables (Lagoon API)
+
+Docker Registry Environment Variables are only available during a Build and are used when attempting to log in to a private registry. It is used to store the password for the user defined in [Specials » `private-registries`](./lagoon_yml.md). They can be applied at the project or environment level.
+
+This defines a project wide docker registry variable (available in all environments) for the project with id `463`:
+```
+mutation addDockerRegistryEnv {
+  addEnvVariable(input:{type:PROJECT, typeId:463, scope:DOCKER, name:"MY_OWN_REGISTRY_PASSWORD", value:"MySecretPassword"}) {
```
Schnitzel: I'm not sure about the name `DOCKER` here, I would have called it `REGISTRY`? also I think we should slowly get rid of Docker all together, as it might not exist for much longer, so maybe even `CONTAINER_REGISTRY`?

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -184,6 +184,54 @@ do
...
+    if [ $PRIVATE_DOCKER_REGISTRY_URL != "false" ]; then
+      echo "Attempting to log in to $PRIVATE_DOCKER_REGISTRY_URL with user $PRIVATE_DOCKER_REGISTRY_USERNAME - $PRIVATE_DOCKER_REGISTRY_PASSWORD"
+      docker login --username $PRIVATE_DOCKER_REGISTRY_USERNAME --password $PRIVATE_REGISTRY_CREDENTIAL $PRIVATE_DOCKER_REGISTRY_URL
+    else
+      echo "Attempting to log in to docker hub with user $PRIVATE_DOCKER_REGISTRY_USERNAME - $PRIVATE_DOCKER_REGISTRY_PASSWORD"
+      docker login --username $PRIVATE_DOCKER_REGISTRY_USERNAME --password $PRIVATE_REGISTRY_CREDENTIAL
+    fi
+  fi
+done
+set +x
```
Schnitzel: this should probably be an `set -x`

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -541,7 +541,7 @@ CREATE OR REPLACE PROCEDURE
         AND column_name = 'scope'
     ) THEN
       ALTER TABLE `env_vars`
-      ADD `scope` ENUM('global', 'build', 'runtime') NOT NULL DEFAULT 'global';
+      ADD `scope` ENUM('global', 'build', 'runtime', 'docker') NOT NULL DEFAULT 'global';
```
Schnitzel: this unfortunately is not gonna work. As the `if` in this procedure just checks if the `scope` column exist or not, and currently all lagoons will have that column already existing it will never run the `ADD`.

Instead we should create a new procedure which specifically checks if `docker` (or `registry` - see above) exists in the list of enums of column `scope` and if not add it.

Schnitzel (changes requested):
nice work! some small things

docs/using_lagoon/environment_variables.md
shreddedbacon: Yep, good idea. I think go with `CONTAINER_REGISTRY`. I'll make changes

shreddedbacon:
I've renamed references to docker registries to container registries.

I wasn't sure on the procedure for modifying the enum so put an `alter table` in.

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
Schnitzel:
```suggestion
      ADD `scope` ENUM('global', 'build', 'runtime', 'container_registry') NOT NULL DEFAULT 'global';
```

docs/using_lagoon/lagoon_yml.md
```diff
@@ -260,3 +260,42 @@ Each definition is keyed by a unique name (`secrets` and `logs-db-secrets` in th
 * `path` - the path to the yaml file
 * `command` - can either be `create` or `apply`, depending on if you like to run `kubectl create -f [yamlfile]` or `kubectl apply -f [yamlfile]`
 * `ignore_error` - either `true` or `false` (default), this allows you to instruct the lagoon build script to ignore any errors that might are returned during running the command. (This can be useful to handle the case where you want to run `create` during every build, so that eventual configurations are created, but don't want to fail if they already exist).
+
+#### `private-registries`
```
Schnitzel: can we also call this `container-registries`? then it will be clean through the whole system, also this technically can be used for non-dockerhub but public registries, so using `container-registries` would fit even more :)

Schnitzel (changes requested):

shreddedbacon:
Have update the PR with the latest suggestions

PR #1315, opened 2019-10-17
Configurable elastic endpoints

### LINKED ISSUE #1313: Configurable elasticsearch endpoints
**Describe the bug**
Some services are hardcoded to use `logs-db:9200`. It would be good to be able to specify a custom elasticsearch endpoint.

shreddedbacon:
# Checklist

This change is to allow `logs-db:9200` to be configurable, this means that either another elasticsearch can be used. Or a service with a `ClusterIP` can be used instead to reduce issues that can come with a headless service and DNS.

# Changelog Entry

Improvement - Allow the `api` to use a configurable elasticsearch endpoint
Improvement - Allow the `auto-idler` to use a configurable elasticsearch endpoint

# Closing issues
closes #1313


tobybellwood:
would this allow to specify an Elasticsearch located outside Lagoon?

shreddedbacon:
Potentially

Schnitzel:
@shreddedbacon 
that's a great change, though I just switched to use `logs-db-service` by default (which exists since Lagoon 1.1.0

can you update this in this PR? thanks!

Schnitzel (changes requested):
use `logs-db-service` by default, also check that this might cause issues for local development, where we probably need to come up with a clever way to handle this.

Schnitzel:
nice idea! there are more places though where we should use this, see https://github.com/amazeeio/lagoon/commit/47a53ea462dd526203006322c08f549ce2ee9389:
- logs-forwarder
- logs2logs-db 
- logs-db-curator

shreddedbacon:
Yup, will work more on this shortly

Schnitzel:
@shreddedbacon 
please check https://github.com/amazeeio/lagoon/pull/1350/files which solves the problem of local `logs-db-service` in a more elegant version. 
I still like the fact that we can define an external elasticsearch all together, but I think we don't need to have the logic switching between `logs-db` and `logs-db-service` for lagoon anymore

shreddedbacon:
Agree, that is definitely more elegant. I'll re-work this to just allow for configurable ES endpoints only.

shreddedbacon:
I've updated this branch to just being able to configure elastic endpoints using envvars, this goes along with needing #1350 to be merged too.

PR #1332, opened 2019-10-21
Add Github usernames to closed Pull Requests

### LINKED ISSUE #1327: Include the GitHub username that created the PR in completed build notifications
It would be very useful to include the GitHub username that created the PR in completed build notifications, so that devs can be notified directly.

One possible solution would be (from @dan2k3k4):

- Add an option on the Lagoon UI per project to change the Slack Message from standard to advanced, with advanced allowing customisation of what end-result looks like:
`[<project>] <pr-num> Build <build-num> completed by <PR Base Username>`
- Then within Slack, the user would just need to add their Github Name to the Get Notified by these words section inside Slack.

Or more simply, the GitHub username that created the PR could be included globally on all completed build notifications.

cc @Schnitzel

dan2k3k4:
Closes #1327

### DIFF
--- a/services/webhooks2tasks/src/handlers/githubPullRequestClosed.js
+++ b/services/webhooks2tasks/src/handlers/githubPullRequestClosed.js
@@ -14,6 +14,8 @@ async function githubPullRequestClosed(webhook: WebhookRequestData, project: Pro
       giturl,
       uuid,
       body,
+      user,
+      sender,
     } = webhook;
 
     const meta = {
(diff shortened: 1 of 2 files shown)


Schnitzel:
@dan2k3k4 
merged! thank you! but this will only work for closed PRs, I think in #1327 we're talking about completed builds for PRs? this is not really a close?

dan2k3k4:
@Schnitzel I believe the original question on slack was for closed PRs, and from what I can tell - we don't know who triggered the build only the PR? So I don't really see how we would solve "build completed by `<PR sender user>`" unless we track this at build notification too?

PR #1334, opened 2019-10-22
Improvement: Extract favicon nginx conf.

### LINKED ISSUE #1333: Override default @empty behaviour.
**Describe the bug**
We have a clear default favicon that we want to use if none is specified. The `= /favicon.ico` location doesn't allow us to define a supplementary configuration to handle overriding the favicon.

I understand that this is suitable for 99% of users, if we declared the favicon handler in a separate file we could easily override for our use case and keep the configuration simple for the base.

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a server include to declare a location for `/favicon.ico` with a try_files directive to the local 
2. Navigate to `/favicon.ico`

**Expected behavior**
1. Ability to change behaviour of `@empty` or `location = /favicon.ico`

steveworley:
Allows favicon.conf to be overriden by images that inherit the base and allows them to continue using the base Drupal configuration.

# Changelog Entry

Improvement - Extracts favicon nginx configuration to an include file so that downstream images can override without inheriting the drupal.conf file (#1333 )

# Closing issues

Closes #1333


images/nginx-drupal/drupal/favicon.conf
```diff
@@ -0,0 +1,12 @@
+## Support for favicon. Return an 1x1 transparent GIF if it doesn't
+## exist.
+location = /favicon.ico {
+  expires 30d;
+  try_files /favicon.ico @empty;
+}
+
+## Return an in memory 1x1 transparent GIF.
```
Schnitzel: let's leave @empty in the default config file, I don't think we need that to be easy overwrite-able and instead let's just include the `location = /favicon.ico` inside `drupal/favicon.conf`

Schnitzel (changes requested):

images/nginx-drupal/drupal/favicon.conf
steveworley: Good point, we might want `@empty` to exist for other places as well - defining that centrally is a better solution.

images/nginx-drupal/drupal.conf
```diff
@@ -134,18 +134,12 @@ server {
...
-  location @empty {
-    expires 30d;
-    empty_gif;
+  ## Return an in memory 1x1 transparent GIF.	
+  location @empty {	
+    expires 30d;	
+    empty_gif;	
   }
 
+  include /etc/nginx/conf.d/drupal/favicon.conf
```
tobybellwood:
```suggestion
  include /etc/nginx/conf.d/drupal/favicon.conf;
```
`nginx: [emerg] invalid number of arguments in "include" directive in /etc/nginx/conf.d/app.conf:144`

PR #1339, opened 2019-10-24
Improvement: wget to curl so proxies can be used for docker image building  (#1338)

### LINKED ISSUE #1338: Improvement: convert wget to curl to allow use with proxies
wget in alpine does not play nicely with proxies with https. curl does.

man page from curl
```

-o, --output <file>
      Write output to <file> instead of stdout. If you are using {} or [] to fetch multiple documents, you can use '#' followed by a number in the <file> speci‐
      fier. That variable will be replaced with the current string for the URL being fetched. Like in:

       curl http://{one,two}.example.com -o "file_#1.txt"

      or use several variables like:

       curl http://{site,host}.host[1-5].com -o "#1_#2"

      You may use this option as many times as the number of URLs you have. For example, if you specify two URLs on the same command line, you can use  it  like
      this:

        curl -o aa example.com -o bb example.net

      and  the order of the -o options and the URLs doesn't matter, just that the first -o is for the first URL and so on, so the above command line can also be
      written as

        curl example.com example.net -o aa -o bb

      See also the --create-dirs option to create the local directories dynamically. Specifying the output as '-' (a single dash) will force the  output  to  be
      done to stdout.

      See also -O, --remote-name and --remote-name-all and -J, --remote-header-name.


-O, --remote-name
      Write output to a local file named like the remote file we get. (Only the file part of the remote file is used, the path is cut off.)

      The  file  will  be  saved in the current working directory. If you want the file saved in a different directory, make sure you change the current working
      directory before invoking curl with this option.

      The remote file name to use for saving is extracted from the given URL, nothing else, and if it already exists it will be overwritten.  If  you  want  the
      server  to  be  able  to choose the file name refer to -J, --remote-header-name which can be used in addition to this option. If the server chooses a file
      name and that name already exists it will not be overwritten.

      There is no URL decoding done on the file name. If it has %20 or other URL encoded parts of the name, they will end up as-is as file name.

      You may use this option as many times as the number of URLs you have.

-S, --show-error
      When used with -s, --silent, it makes curl show an error message if it fails.

-s, --silent
      Silent  or  quiet  mode. Don't show progress meter or error messages.  Makes Curl mute. It will still output the data you ask for, potentially even to the
      terminal/stdout unless you redirect it.

      Use -S, --show-error in addition to this option to disable progress meter but still show error messages.

      See also -v, --verbose and --stderr.


-L, --location
      (HTTP)  If  the server reports that the requested page has moved to a different location (indicated with a Location: header and a 3XX response code), this
      option will make curl redo the request on the new place. If used together with -i, --include or -I, --head, headers  from  all  requested  pages  will  be
      shown. When authentication is used, curl only sends its credentials to the initial host. If a redirect takes curl to a different host, it won't be able to
      intercept the user+password. See also --location-trusted on how to change this. You can limit the amount of redirects to follow by using the  --max-redirs
      option.

      When  curl  follows  a  redirect  and  the  request  is not a plain GET (for example POST or PUT), it will do the following request with a GET if the HTTP
      response was 301, 302, or 303. If the response code was any other 3xx code, curl will re-send the following request using the same unmodified method.

      You can tell curl to not change the non-GET request method to GET after a 30x response by using the dedicated options for that: --post301,  --post302  and
      --post303.
```

duttonw:
# Changelog Entry #1338 
Convert wget to curl calls so proxied systems can still build images

# Closing issues
closes #1338

### DIFF
--- a/local-dev/api-data-watcher-pusher/api-watch-push.sh
+++ b/local-dev/api-data-watcher-pusher/api-watch-push.sh
@@ -16,7 +16,7 @@ send_graphql_query() {
     # Create a correct json string
     json="{\"query\": \"$data\"}"
 
-    wget --header "Content-Type: application/json" --header "$bearer" api:3000/graphql --post-data "$json" --content-on-error -O -
+    curl -s -L api:3000/graphql --header "Content-Type: application/json" --header "$bearer" --data "$json" --show-error
 }
 
 watch_apidatafolder() {
--- a/services/ssh/authorize.sh
+++ b/services/ssh/authorize.sh
@@ -21,7 +21,7 @@ ssh_username=$1
 ssh_fingerprint=$2
 
 data="{\"fingerprint\": \"$ssh_fingerprint\"}"
-keys=$(wget --header "Content-Type: application/json" --header "$bearer" $api/keys --post-data "$data" -q --content-on-error -O -)
+keys=$(curl curl -s -L $api/keys --header "Content-Type: application/json" --header "$bearer" --data "$data" --show-error)
 
 options="no-port-forwarding,no-X11-forwarding,no-agent-forwarding"
 
--- a/services/ssh/home/grant.sh
+++ b/services/ssh/home/grant.sh
@@ -31,4 +31,4 @@ data="{\"userId\": \"$USER_ID\", \"grant\": true}"
 
 # Submit the token request as a POST request with the JSON data
 # containing the key.
-echo $(wget "$server/generate" --header "$header" --post-data "$data" -q -O -)
+echo $(curl -s -L "$server/generate" --header "$header" --data "$data")
--- a/services/ssh/home/token-debug.sh
+++ b/services/ssh/home/token-debug.sh
@@ -34,9 +34,9 @@ header="Content-Type: application/json"
 # Prepare the post (containing the user id) as a JSON object.
 data="{\"userId\": \"$USER_ID\"}"
 
-token=$(wget "$server/generate" --header "$header" --post-data "$data" -d -v -O - 2>&1)
+token=$(curl -s -L "$server/generate" --header "$header" --data "$data" -d -v 2>&1)
 echo "token='${token}'" >> /proc/1/fd/1
 
 # Submit the token request as a POST request with the JSON data
 # containing the key.
-echo $(wget "$server/generate" --header "$header" --post-data "$data" -q -O -)
+echo $(curl -s -L "$server/generate" --header "$header" --data "$data")
--- a/services/ssh/home/token.sh
+++ b/services/ssh/home/token.sh
@@ -31,4 +31,4 @@ data="{\"userId\": \"$USER_ID\"}"
 
 # Submit the token request as a POST request with the JSON data
 # containing the key.
-echo $(wget "$server/generate" --header "$header" --post-data "$data" -q -O -)
+echo $(curl -s -L "$server/generate" --header "$header" --data "$data")
--- a/images/mariadb-galera/Dockerfile
+++ b/images/mariadb-galera/Dockerfile
@@ -51,6 +51,7 @@ RUN apk add --no-cache --virtual .common-run-deps \
         rsync \
         sqlite-libs \
         tzdata \
+        curl \
         wget && \
     apk add --no-cache --virtual .galera-build-deps \
         asio-dev && \
--- a/images/mariadb/Dockerfile
+++ b/images/mariadb/Dockerfile
@@ -41,6 +41,7 @@ RUN \
     net-tools \
     pwgen \
     tzdata \
+    curl \
     wget; \
     rm -rf /tmp/* /var/tmp/* /var/cache/apk/* /var/cache/distfiles/*; \
     rm -rf /var/lib/mysql/* /etc/mysql/; \
--- a/images/node/builder/Dockerfile
+++ b/images/node/builder/Dockerfile
@@ -25,9 +25,10 @@ RUN apk update \
         bash \
         ca-certificates \
         wget \
+        curl \
         libpng-dev \
-    && wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub \
-    && wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.28-r0/glibc-2.28-r0.apk \
+    && curl -s -L -q -o /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub \
+    && curl -s -L -q -O https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.28-r0/glibc-2.28-r0.apk \
     && apk add glibc-2.28-r0.apk \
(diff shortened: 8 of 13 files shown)


local-dev/api-data-watcher-pusher/api-watch-push.sh
```diff
@@ -16,7 +16,7 @@ send_graphql_query() {
     # Create a correct json string
     json="{\"query\": \"$data\"}"
 
-    wget --header "Content-Type: application/json" --header "$bearer" api:3000/graphql --post-data "$json" --content-on-error -O -
+    curl -s -L api:3000/graphql --header "Content-Type: application/json" --header "$bearer" --data "$json" --show-error
```
Schnitzel, line 19: this will never use external proxies, please revert

services/ssh/authorize.sh
```diff
@@ -21,7 +21,7 @@ ssh_username=$1
 ssh_fingerprint=$2
 
 data="{\"fingerprint\": \"$ssh_fingerprint\"}"
-keys=$(wget --header "Content-Type: application/json" --header "$bearer" $api/keys --post-data "$data" -q --content-on-error -O -)
+keys=$(curl curl -s -L $api/keys --header "Content-Type: application/json" --header "$bearer" --data "$data" --show-error)
```
Schnitzel, line 24: this will never use external proxies, please revert

services/ssh/home/grant.sh
```diff
@@ -31,4 +31,4 @@ data="{\"userId\": \"$USER_ID\", \"grant\": true}"
 
 # Submit the token request as a POST request with the JSON data
 # containing the key.
-echo $(wget "$server/generate" --header "$header" --post-data "$data" -q -O -)
+echo $(curl -s -L "$server/generate" --header "$header" --data "$data")
```
Schnitzel, line 34: this will never use external proxies, please revert

services/ssh/home/token-debug.sh
```diff
@@ -34,9 +34,9 @@ header="Content-Type: application/json"
 # Prepare the post (containing the user id) as a JSON object.
 data="{\"userId\": \"$USER_ID\"}"
 
-token=$(wget "$server/generate" --header "$header" --post-data "$data" -d -v -O - 2>&1)
+token=$(curl -s -L "$server/generate" --header "$header" --data "$data" -d -v 2>&1)
```
Schnitzel, line 37: this will never use external proxies, please revert

services/ssh/home/token-debug.sh
```diff
@@ -34,9 +34,9 @@ header="Content-Type: application/json"
...
 echo "token='${token}'" >> /proc/1/fd/1
 
 # Submit the token request as a POST request with the JSON data
 # containing the key.
-echo $(wget "$server/generate" --header "$header" --post-data "$data" -q -O -)
+echo $(curl -s -L "$server/generate" --header "$header" --data "$data")
```
Schnitzel, line 42: this will never use external proxies, please revert

services/ssh/home/token.sh
```diff
@@ -31,4 +31,4 @@ data="{\"userId\": \"$USER_ID\"}"
 
 # Submit the token request as a POST request with the JSON data
 # containing the key.
-echo $(wget "$server/generate" --header "$header" --post-data "$data" -q -O -)
+echo $(curl -s -L "$server/generate" --header "$header" --data "$data")
```
Schnitzel, line 34: this will never use external proxies, please revert

Schnitzel (changes requested):
@duttonw

Thanks for this, I think that's a good idea, just can we revert the changes where this is only used for internal requests (see my change requests). 
Thanks!

duttonw:
> @duttonw
> 
> Thanks for this, I think that's a good idea, just can we revert the changes where this is only used for internal requests (see my change requests).
> Thanks!

Hi @Schnitzel ,

It should never go to the external proxy if the no_proxy settings is set correctly.  These could be reverted but I feel it is nicer to keep it all consistent with curl. Has the CICD pipeline been correct to get a passing build yet?

If you do want to keep it as wget for those internal internal requests i'm welcome to you pushing those changes onto this branch before merging.

smlx:
This PR replaces busybox wget calls in e.g. php images, and also GNU wget calls in e.g. node images. It seems unlikely that both these wget implementations have the same issue with proxies, so I wonder if the problem is a configuration issue instead?

Could you elaborate on the specific issue this PR solves?

duttonw:
Here's the issues that others have been having with wget alpine/busybox. If wget is replaced with gnu wget, then it would work better, but then thats a little bit bigger size then the busybox version. 

https://github.com/wurstmeister/kafka-docker/issues/487
https://gitlab.alpinelinux.org/alpine/aports/issues/10446
https://gitlab.com/gitlab-org/gitlab-ci-yml/merge_requests/197
https://unix.stackexchange.com/a/525474

smlx:
So the specific issue that this PR addresses is that wget is unable to access HTTPS through a HTTP proxy?

This bug was [fixed](http://lists.busybox.net/pipermail/busybox/2018-March/086282.html) in [busybox 1.29.0](https://www.busybox.net/). The earliest release of alpine with this version was [3.9](https://pkgs.alpinelinux.org/packages?name=busybox&branch=v3.9).

According to https://github.com/amazeeio/lagoon/issues/1324 only mariadb and varnish images use an old enough alpine to be affected by that bug. Is that where you were seeing the issue?

duttonw:
I'm still on the learning curve of containers and docker needing to work in a restricted network where the only way to communicate to the internet is via a proxy.

We have this problem with image amazeeio/python:2.7-ckan-v0.23.1 

which salsadigital built for us in ckan plugin https://github.com/qld-gov-au/ckanext-data-qld-theme/blob/develop/.docker/Dockerfile.ckan

from the looks of https://hub.docker.com/r/amazeeio/python/tags?page=1&name=ckan
would this issue go away if we change the version to tag 2.7-ckan-v1.1.4 ?

smlx:
That image does seem to be new enough:
```
$ docker run -it --rm  amazeeio/python:2.7-ckan-v0.23.1 -- cat /etc/alpine-release
3.9.4
```

smlx:
after some testing, busybox wget _does_ seem to work if you set `http_proxy`, but it has some issues with redirects via the proxy.. so curl is definitely a more robust solution :-)
sorry for the noise..

duttonw:
set http_proxy and https_proxy to http://externalip:3128 which is a squid3 proxy. wget to a https address in busybox image, the squid logs show it does not use connect and fails, which curl does correctly.

smlx:
@duttonw ping to see if you are able to address @Schnitzel's comments on this PR?

duttonw:
Hi @smlx , per https://github.com/amazeeio/lagoon/pull/1339#issuecomment-555308203 he has not responded if he still wants me to revert or keep them consistent as per my reply.

smlx:
ok, thanks. I missed this bit:
> If you do want to keep it as wget for those internal internal requests i'm welcome to you pushing those changes onto this branch before merging.

I might push some changes as per your comment :slightly_smiling_face:

duttonw:
just re-based due to conflicts. your welcome to pull in house and merge when happy :)

PR #1165, opened 2019-08-03
UI Deployments Improvements

rocketeerbkw:
This PR addresses a number of issues related to deployments shown in the UI. See changelog for more details.

1. New button style and help text added to the `deploy` button to reduce confusion. The help text is deploy type aware:

image
image
image

2. New API mutation and button in UI to cancel running/stuck deployments
image
```graphql
mutation cancelDeployment {
  cancelDeployment(input: {
    deployment:{
      name: "lagoon-11"
      environment: {
        name:"drupal8-composer-73-mariadb"
        project:{
          name: "ci-drupal"
        }
      }
    }
  })
}
```

The mutation makes some changes to the RBAC permissions:

- Deployment resource has a new scope `cancel`
- `cancelDeployment` requires users to be at least maintainers for the project

 

# Checklist
- [X] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated.
- [X] Changelog entry has been written

Explain the **details** for making this change. What existing problem does the pull request solve?

# Changelog Entry

Improvement - Environment deploy button help text and site-wide button styles (#1013)
Improvement - Add cancelDeployment API mutation
Improvement - Add button in UI to cancel running/stuck deployments (#1155)
Bugfix - Environments of type "promote" were not marked as such on the environments page

# Closing issues
Closes #1013, Closes #1155

PR #1346, opened 2019-10-26
Add Task for running Drupal core-cron.

### LINKED ISSUE #1340: Add drush cron "task" to admin UI
The ability to trigger a drush cron run via the tasks system would be of great benefit to a number of GovCMS agencies (given their restricted access to CLI).

Being able to trigger a cron run via a task/GQL, integrated with RBAC would allow for regulated access to be able to perform ad-hoc cron runs.

The task should be adequately protected from multi-activation/spamming - adding in an inbuilt check for last-run/elapsed time (e.g. min 5m between runs) would reduce load on the sites.

fubarhouse:
# Checklist

This change introduces a task in the API and UI to trigger Drush to run `core-cron` or `core:cron`. The change has previously been discussed directly and an issue for this exists where any discussion is yet to occur. (#1340)

I'd appreciate some guidance/feedback before marking this as ready, if it's possible 🙂

# Changelog Entry
* Feature: Add cron task to UI to trigger Drush cron. (#1340)

# Closing issues
* Closes #1340


services/api/src/resources/task/resolvers.js
```diff
@@ -347,6 +347,40 @@ const taskDrushCacheClear = async (
...
+  const command =
+    'drupal_version=$(drush status drupal-version --format=list) && \
+  if [ ${drupal_version%.*.*} == "8" ]; then \
+    drush core:cron; \
+  elif [ ${drupal_version%.*} == "7" ]; then \
+    drush cc core-cron; \
+  else \
+    echo "could not execute cron for found Drupal Version ${drupal_version}"; \
+    exit 1; \
+  fi';
```
seanhamlin: Curious as to why you are running 2 different commands based on Drupal version. I thought `drush cron` is fairly ubiquitous.

services/ui/src/components/AddTask/components/DrushCron.js
```diff
@@ -0,0 +1,96 @@
...
+            />
+          </div>
+          <button
+            onClick={() =>
+              taskDrushCron({
+                variables: {
+                  environment: pageEnvironment.id
+                }
+              })
+            }
```
seanhamlin, line 66: Would be good to have protection in the UI against hitting the button repeatedly

seanhamlin (changes requested):
Nice work! Had a few comments.

services/api/src/resources/task/resolvers.js
fubarhouse: `dush cron` is an alias for `core-cron` (D7) and `core:cron` (D8).
If you aren't bothered I will change them to use the alias `cron`.

fubarhouse: Done.

services/ui/src/components/AddTask/components/DrushCron.js
fubarhouse: Got a change ready for this one, I just need to test it first.

tobybellwood: What's a sensible protection here.  I was originally thinking placing an elapsed time check in the task itself, but realise that setting it to more than a minute may negate the viability of the task.

fubarhouse: It should be noted that the other tasks don't have a mechanism to delay based on usage.

I think it's a great idea but I believe that should not be controlled by the individual task, but where the command is actually evaluated. The vector already exists - so shouldn't the solution for that problem be made generic enough and implementable for all tasks...?

(i can open a separate issue for this?)

seanhamlin: Sounds good, a new issue sounds like a great. The main idea from my perspective is to prevent accidental double clicks.

services/api/src/resources/task/resolvers.js
seanhamlin: Yeah,  the command `cron` is what everyone understands. Also it makes this code a little more future proof when Drupal 9 comes out.

services/ui/src/components/AddTask/components/DrushCron.js
fubarhouse: I'll get this change tested tonight, and submit a new issue for adding task constraints.
Opened #1348 in response to https://github.com/amazeeio/lagoon/pull/1346#discussion_r339320421

fubarhouse: **Step 1**
image

**Step 2**
image

**Step 3**
image

fubarhouse: @seanhamlin,
Would you consider this review as resolved, or is there more work to do?

services/keycloak/start.sh
```diff
@@ -827,6 +827,18 @@ EOF
...
+    /opt/jboss/keycloak/bin/kcadm.sh create clients/$CLIENT_ID/authz/resource-server/permission/scope --config $CONFIG_PATH -r lagoon -f - <<EOF
+{
+  "name": "Run Drush cron",
+  "type": "scope",
+  "logic": "POSITIVE",
+  "decisionStrategy": "UNANIMOUS",
+  "resources": ["task"],
+  "scopes": ["drushCron:production","drushCron:development"],
+  "policies": ["User has access to project","Users role for project is Guest"]
+}
```
Schnitzel: we need to move this outside of the `configure_api_client()` function, as `configure_api_client` is only called once (at the very beginning when keycloak starts the very first time)`.

see https://github.com/amazeeio/lagoon/blob/d9875b2a5b1d53c197d18c29b84b6a2b40e3bf81/services/keycloak/start.sh#L1324 as an example how we add additional things to the keycoak api (like upgrade paths)

Schnitzel (changes requested):

services/keycloak/start.sh
```diff
@@ -153,7 +153,7 @@ function configure_api_client {
...
+    resource_scope_names=(add add:development add:production addGroup addNoExec addNotification addOrUpdate:development addOrUpdate:production addUser delete delete:development delete:production deleteAll deleteNoExec deploy:development deploy:production drushArchiveDump:development drushArchiveDump:production drushCacheClear:development drushCacheClear:production drushCron:developement drushCron:production drushRsync:destination:development drushRsync:destination:production drushRsync:source:development drushRsync:source:production drushSqlDump:development drushSqlDump:production drushSqlSync:destination:development drushSqlSync:destination:production drushSqlSync:source:development drushSqlSync:source:production environment:add:development environment:add:production environment:view:development environment:view:production getBySshKey project:add project:view removeAll removeGroup removeNotification removeUser ssh:development ssh:production storage update update:development update:production view view:token view:user viewAll viewPrivateKey)
```
rocketeerbkw: There's a small typo here `drushCron:developEment` should be `drushCron:development`. Also, see @Schnitzel comment on lines +830 to +841, all these new additions should happen in a new function call, not part of `configure_api_client`. This applies to line 172 also.

rocketeerbkw (changes requested):

rocketeerbkw:
This will conflict with #1165. Easiest is probably to merge this first, then fix conflicts over there.

services/keycloak/start.sh
fubarhouse: I have corrected the typo and adjusted the keycloak configuration function to support upgrade paths.

services/keycloak/start.sh
fubarhouse: I have adjusted the logic to support an upgrade path for existing installations.

(see https://github.com/amazeeio/lagoon/pull/1346/commits/df4d39a472d70cfec2612d9a0c51b15285b622ba)

docs/administering_lagoon/rbac.md
```diff
@@ -136,6 +136,8 @@ Here is a table that lists the roles and the access they have:
 | taskDrushSqlDump                          | task         | drushSqlDump:production              | projectID  | Yes                 | Yes                 | Yes   | Yes        | Yes       | No       | No    |      |
 | taskDrushCacheClear                       | task         | drushCacheClear:development          | projectID  | Yes                 | Yes                 | Yes   | Yes        | Yes       | Yes      | Yes   |      |
 | taskDrushCacheClear                       | task         | drushCacheClear:production           | projectID  | Yes                 | Yes                 | Yes   | Yes        | Yes       | Yes      | Yes   |      |
+| taskDrushCron                             | task         | taskDrushCron:development            | projectID  | Yes                 | Yes                 | Yes   | Yes        | Yes       | Yes      | Yes   |      |
```
rocketeerbkw: I just caught that the scope names here don't match in code, shouldn't have the `task` prefix.

services/keycloak/start.sh
```diff
@@ -1293,6 +1293,31 @@ function add_group_viewall {
...
+  echo Configuring group:drushCron
+
+  /opt/jboss/keycloak/bin/kcadm.sh create clients/$CLIENT_ID/authz/resource-server/permission/scope --config $CONFIG_PATH -r lagoon -f - <<EOF
+{
+  "name": "Run Drush cron",
+  "type": "scope",
+  "logic": "POSITIVE",
+  "decisionStrategy": "UNANIMOUS",
+  "resources": ["task"],
+  "scopes": ["drushCron:production","drushCron:development"],
```
rocketeerbkw, line 1325: Thanks for moving this. Unfortunately, this will still fail in an upgrade path since the two new scopes will only be added during a fresh install. And by "fail" I mean the install script will complete w/o errors, but evaluating these permissions will always error (and return permission denied).

In this function you'll need to add the two new scopes (can copy the `# Resource Scopes` section line ~159), and then add those scopes to the `task` resource. For an example of that you can look at my PR for UI improvements. Unfortunately you have to send a PATCH request will all scopes, not just the ones that need adding: https://github.com/amazeeio/lagoon/pull/1165/files#diff-8c9af22e910ccf7cd25a29a92c5faadaR1307-R1308

When those are moved, it'd be best to remove them from `configure_api_client` also that way there won't be any errors on a fresh install.

rocketeerbkw (changes requested):

services/keycloak/start.sh
fubarhouse: @rocketeerbkw,

I believe I've tended to what you've described.

Thanks for your patience.

rocketeerbkw: Sorry this update mechanism isn't very intuitive. I've fix it in https://github.com/amazeeio/lagoon/commit/a39d6d2c36278da1a7457498b2865d39bbd756d1, can you merge that in? Should be good to go after that

fubarhouse: I have cherry-picked a39d6d2c36278da1a7457498b2865d39bbd756d1

rocketeerbkw (approved):

PR #1349, opened 2019-10-28
Expose max_allowed_packet to env vars.

steveworley:
Allows max_allowed_packets to be overridden with an env var.

 
 

# Checklist

Exposes an environment variable that can control the max_allowed_packet.

# Changelog Entry

Improvement - Allow max_allowed_packet size to be controlled via env var.


seanhamlin:
Do we need docs on this? The docs could explain why this would be useful as well.

Schnitzel:
yes, the documentation PR has just been merged, so let's add that env variable to here: https://github.com/amazeeio/lagoon/blob/master/docs/using_lagoon/docker_images/mariadb.md

PR #1352, opened 2019-10-30
Keycloak start.sh script needs python installed

rocketeerbkw:
# Checklist

The recent switch to ubi8-minimal for the keycloak image means python is no longer installed by default. We use python for the install/upgrade scripts in `start.sh`.

# Changelog Entry

Bugfix - Keycloak container switched to different base image that doesn't include python

# Closing issues
n/a

### DIFF
--- a/services/keycloak/Dockerfile
+++ b/services/keycloak/Dockerfile
@@ -4,7 +4,7 @@ FROM jboss/keycloak
 
 USER root
 
-RUN microdnf install -y nc \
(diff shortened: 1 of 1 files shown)


fubarhouse:
Tested the same change the other day 🙂

tobybellwood:
@rocketeerbkw - if #1358 is the complete solution (it passed tests) - can we close this?

PR #1355, opened 2019-10-30
Use KVM as the minishift vm-driver on Linux

smlx:
Performance is far better than virtualbox, and it's the native/default hypervisor.
 
# Checklist

KVM is the [upstream recommended](https://docs.okd.io/latest/minishift/getting-started/setting-up-virtualization-environment.html#for-linux) driver.

# Changelog Entry
Improvement - Use the KVM driver for minishift on linux

# Closing issues
n/a


twardnw (approved):

dasrecht (approved):

dasrecht:
Looks good to me - testing this currently on testbed1 - the only issue i had that there were still some files in ~./minishift that lead to minishift still trying to interface with virtualbox. 
After removing those everything started to work

dasrecht:
The only thing missing would be adapting the documentation to send people towards - https://docs.okd.io/latest/minishift/getting-started/setting-up-virtualization-environment.html

For the testbed i created a ansible playbook to install the things needed - we could also opensource that part

Schnitzel:
this fails the automated testing, as it is running on Linux and the CI server does not have KVM installed.
Can we get KVM installed there before we merge this, or all the CI will fail ;)

dasrecht:
yes sure - will work on that - actually got the whole thing scripted so it shoudln't be a big thing to get that started

smlx:
For reference, the required packages to install and group to add the test user to are here: https://wiki.debian.org/KVM#Installation

dasrecht:
I'll run https://github.com/amazeeio/lagoon-testbed/blob/master/playbooks/lagoon-requisites.yml (non public yet) against our build system tomorrow so we can get the kvm builds also going

Schnitzel:
@dasrecht status update on this? We would like to add this to 1.2.0

dasrecht:
I worked a bit on it and got it running on the testbed with 18.04 - sadly our CI host is on Ubuntu 16.04 which somehow has some broken KVM things which i first need to look into

vincenzodnp (commented):
As `Makefile` is currently set up, if you're on OSX you will run `minishift` by using `Virtualbox` driver while if on GNU/Linux with KVM. So, if you're on GNU/Linux and would like to run `minishift` with `Virtualbox`, you have to edit the Makefile before run `make minishift`.

Then, why not create a `MINISHIFT_VM_DRIVER` variable, set with default values depending on OS (so KVM for GNU/Linux and VirtualBox for OSX) but that could be easily overridden with `MINISHIFT_VM_DRIVER=kvm make minishift`?

smlx:
@vincenzodnp maybe that can be a followup PR?

dasrecht:
Yes let's do this as followup PR - I'll work on that lateron with @Schnitzel as we're adding a few new testing nodes that are on ubuntu 18.04 i'm fairly positive we won't need to fix this on 16.04 anymore.

Will check back at Standup and then update the ticket here.

PR #1358, opened 2019-10-30
Patch keycloak Dockerfile and entrypoint for new base image

smlx:
# Checklist

The base image our keycloak is `jboss/keycloak`. This had a new release tagged, and as we don't currently pin base images against a tag we get `latest` when building from scratch. This new version has an incompatibility with our entrypoint scripts.

Therefore this PR implements the following:

* install python2, as it is no longer part of the base image
* enable script upload, as this is disabled by default in 7.0.1
* pin base image version to 7.0.1 (current latest version)

# Changelog Entry
Bugfix - Fix incompatibility and pin new upstream keycloak version

# Closing issues
Supersedes #1352

### DIFF
--- a/services/keycloak/Dockerfile
+++ b/services/keycloak/Dockerfile
@@ -1,10 +1,10 @@
 ARG IMAGE_REPO
 FROM ${IMAGE_REPO:-lagoon}/commons as commons
-FROM jboss/keycloak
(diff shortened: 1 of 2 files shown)


Schnitzel:
thanks @smlx !

PR #1361, opened 2019-10-30
Use keycloak service name in docker-compose.yaml

smlx:
# Checklist

The localhost DNS alias doesn't exist on Linux, so instead use the docker network internal service name and port.

# Changelog Entry
Improvement - Use docker keycloak service name and port in `docker-compose.yaml` for cross-platform compatibility

# Closing issues
n/a


smlx:
Ah, found this breaks redirects. hmm.

PR #1372, opened 2019-11-04
Ensure that cron jobs that are every 15 minutes do not create a kuber…

### LINKED ISSUE #1371: Use CLI pods (and not k8s native cronjobs) for crons that happen every 15 minutes
**Describe the bug**

Currently, the cronjobs get created as native cronjobs if the run time is 15 minutes or less, but the logic checks it as `less than 15`, so `*/15 * * * *` will actually run in a native job, but in heavy load clusters this seems to be not very helpful as every project has a cronjob that runs every 15 minutes.

https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L14-L33

I believe this issue will still be consistent with the original issue #1001.

**To Reproduce**
1. Create new project with default cron `*/15 * * * *`
2. Notice upon deploy you have a k8s native cronjob

**Expected behavior**
Only create k8s native cronjobs for workloads with more fine grained requirements, e.g. 'run at 10am on a Thursday'

seanhamlin:
…netes native cronjob, and instead just fire in the CLI pod.

# Checklist

# Changelog Entry
Bugfix - Cron will now use the CLI pod instead of kubernetes native cronjobs when run at least every 15 minutes. This will help clusters with high amounts of 15 minute crons to not create a lot of cronjobs. (#1371)

# Closing issues
Closes #1371

### DIFF
--- a/images/oc-build-deploy-dind/build-deploy-docker-compose.sh
+++ b/images/oc-build-deploy-dind/build-deploy-docker-compose.sh
@@ -12,22 +12,26 @@ function outputToYaml() {
 }
 
 function cronScheduleMoreOftenThan15Minutes() {
-  #takes a unexpanded cron schedule, returns 0 if it's more often that 15 minutes
+  # Work out if a particular cron schedule runs at least every 15 minutes.
+  # returns 0 if cron runs at least every 15 minutes. This includes a 15 minute
+  # cron as well. Running a cron every 10 minutes will also return 0.
+  # returns 1 if cron runs less often, say every 30 minutes.
   MINUTE=$(echo $1 | (read -a ARRAY; echo ${ARRAY[0]}) )
   if [[ $MINUTE =~ ^(M|H|\*)\/([0-5]?[0-9])$ ]]; then
-    # Match found for M/xx, H/xx or */xx
-    # Check if xx is smaller than 15, which means this cronjob runs more often than every 15 minutes.
+    # Match found for `M/xx`, `H/xx` or `*/xx`.
+    # Check if xx is smaller or equal to 15, which means this cronjob runs at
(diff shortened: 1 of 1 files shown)


images/oc-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -12,22 +12,26 @@ function outputToYaml() {
...
     else
       return 1
     fi
   elif [[ $MINUTE =~ ^\*$ ]]; then
-    # We are running every minute
+    # We are running every minute.
     return 0
   else
-    # all other cases are more often than 15 minutes
+    # All other cases are less often than 15 minutes.
```
Schnitzel: I don't think that "less" is correct now, this would mean that "other cases" are 15minute and higher (like 1 hour, 16min, etc.). But we return 1 when we have 14, 13, 12, etc.

So I would leave it at `All other cases are more often than 15 minutes.`, as before it actually would have needed to be called `All other cases are more often than or exactly 15 minutes.`

Schnitzel (changes requested):

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
seanhamlin: Yeah, the wording in general was doing my head in by the end of this PR. Will change.

Schnitzel:
can we combine this with https://github.com/amazeeio/lagoon/pull/1412 ? 
I would like to use `-le` instead of `-lt` but we also discussed to use the configurable environment variable.

tobybellwood:
implemented in #1412

PR #1375, opened 2019-11-05
Add mariadb-client MARIADB_MAX_ALLOWED_PACKET variable.

### LINKED ISSUE #1374: MySQL Client max_allowed_packet mismatch
**Describe the bug**
MySQL client in the CLI image doesn't have the same configuration for max_allowed_packet, this causes issues with running particular MySQL operations, eg. mysqldump. As the PHP Image is bound entirely by the MariaDB server configuration it can insert rows that exceed the ability for the CLI container to fetch.

**To Reproduce**
1. Insert a row with > 16M data
2. Select the row with a mysql query

Example:
```
mysql -h $MARIADB_HOST -u$MARIADB_USERNAME -p$MARIADB_PASSWORD $MARIADB_DATABASE -e "select value from key_value limit 5908, 1;"
ERROR 2020 (HY000) at line 1: Got packet bigger than 'max_allowed_packet'
```
```
drush sql-dump > backup.sql
ERROR 2020 (HY000) at line 1: Got packet bigger than 'max_allowed_packet' at 5908
```

```
mysql --help | grep max-allowed-packet
max-allowed-packet                16777216
```

**Expected behavior**
The client can access rows that the application can insert.

steveworley:
The MySQL client also has a configuration value for max_allowed_packet and this can be different from the server (which in our case is 64M). This is currently using the default value of 16M. This causes issues when performing MySQL operations directly against the DB (PHP relies on the server configuration and is not subject to this problem), specifically mysqldump (drush sql-dump). If a row is inserted that is higher than the 16M default, the mysql client cannot retrieve the value.

# Changelog Entry

Improvement - Allow client max_allowed_packet to be configured by an environment variable. (#1374 )

# Closing issues
Closes #1374


Schnitzel:
that's a great addition, thanks @steveworley 

I don't think though this will work right now, as the file `mariadb-client.cnf` will end up with having `${MARIADB_MAX_ALLOWED_PACKET:-64M}` inside the file. In order to do this correctly we need to envplate this file, here an example:
https://github.com/amazeeio/lagoon/blob/master/images/php/fpm/entrypoints/70-php-config.sh 

also we would need a documentation in here for this: https://lagoon.readthedocs.io/en/latest/using_lagoon/docker_images/php-cli/

tobybellwood:
If there's no issue with the entrypoint ordering (I don't _think_ so) - can we renumber this entrypoint to keep the amazeeio-provided entrypoints to <100 - just in case anyone else is adding entrypoints to our images downstream?


```
cli-drupal:/lagoon/entrypoints$ ls -a /lagoon/entrypoints | sort
.
..
00-umask.sh
05-ssh-key.sh
10-passwd.sh
10-ssh-agent.sh
50-dotenv.sh
50-ssmtp.sh
55-cli-helpers.sh
55-generate-env.sh
60-php-xdebug.sh
61-php-xdebug-cli-env.sh
70-php-config.sh
71-php-newrelic.sh
80-shell-timeout.sh
90-composer-path.sh
90-cronjobs.sh
```

steveworley:
Sure - I didn't know if there was a naming convention there and I looked through the mariadb image and it was using 200 for the envplate so I just assumed that would be okay.

https://github.com/amazeeio/lagoon/tree/master/images/mariadb/entrypoints

images/php/cli/Dockerfile
```diff
@@ -56,6 +56,10 @@ RUN echo "source /lagoon/entrypoints/80-shell-timeout.sh" >> /home/.bashrc
 COPY 61-php-xdebug-cli-env.sh /lagoon/entrypoints/
 RUN echo "source /lagoon/entrypoints/61-php-xdebug-cli-env.sh" >> /home/.bashrc
 
+# Copy mariadb-client configuration.
+COPY 200-mariadb-envplate.sh /lagoon/entrypoints/
```
Schnitzel:
```suggestion
COPY 90-mariadb-envplate.sh /lagoon/entrypoints/
```

Schnitzel (changes requested):

images/php/cli/Dockerfile
Schnitzel: @steveworley just a small fix :)

Schnitzel:
so this fails because the CLI pod cannot be started anymore:
```
jenkins@ax51-2:/home/michael.schmid$ oc logs -f pod/cli-1-swhkf
Agent pid 20
Identity added: /home/.ssh/key ((unnamed))
2019/12/06 17:18:50 [ ERROR ] Error while parsing '/etc/my.cnf.d/mariadb-server.cnf': open /etc/my.cnf.d/mariadb-server.cnf: permission denied
```

we need to run a fix-permissions after copying files into Dockerfiles that we want to envplate, an example:
https://github.com/amazeeio/lagoon/blob/master/images/mariadb/Dockerfile#L58-L60

@steveworley 
can you add an 
```
RUN fix-permissions /etc/my.cnf.d/
```
after you copied stuff into `/etc/my.cnf.d/` inside the Dockerfile?

PR #1381, opened 2019-11-05
lte & gte w/ month and next month

### LINKED ISSUE #1363: HitsMonth calculation fails with parse_exception
**Describe the bug**
When running the following query, the result returns with errors
```
query {
  allProjects { 
      _:environments(includeDeleted: true, type: PRODUCTION) { 
      project { 
        name 
      } 
      openshiftProjectName
      name
      __:hitsMonth(month: "2019-10") { 
        total 
      } 
    }
  } 
}
```

The errors are
```
[parse_exception] failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]: [failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]] (and) [parse_exception] failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]: [failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]] (and) [parse_exception] failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]: [failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]] (and) [parse_exception] failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]: [failed to parse date field [2019-10] with format [strict_date_optional_time||epoch_millis]]
```

The call to elasticsearch is failing https://github.com/amazeeio/lagoon/blob/master/services/api/src/resources/environment/resolvers.js#L356-L385

A successful request needs to contain a full timestamp `YYYY-MM-DDTHH:MM:SS` rather than just the `YYYY-MM`

The following query succeeds.
```
GET router-logs-project-*/_count
{
  "query": {
    "bool": {
      "must": [
        {
          "range": {
            "@timestamp": {
              "gte": "2019-10-01T00:00:00",
              "lte": "2019-11-01T00:00:00"
            }
          }
        }
      ],
      "must_not": [
        {
          "match_phrase": {
            "request_header_useragent": {
              "query": "StatusCake"
            }
          }
        }
      ]
    }
  }
}
```

wintercreative:
Fix for #1363 that builds off https://github.com/amazeeio/lagoon/pull/1380

Bugfix - Define the Date Format for loading of Hits

Closing issues
closes #1363

### DIFF
--- a/services/api/src/resources/environment/resolvers.js
+++ b/services/api/src/resources/environment/resolvers.js
@@ -349,11 +349,27 @@ const getEnvironmentHitsMonthByEnvironmentId = async (
 ) => {
   await hasPermission('environment', 'storage');
 
-  const interested_date = args.month ? new Date(args.month) : new Date();
-  const year = interested_date.getFullYear();
-  const month = interested_date.getMonth() + 1;
+  const splits = args.month.split('-');
+  const yearSplit = splits[0];
+  const monthSplit = splits[1];
+
+  const interestedDate = new Date(`${yearSplit}/${monthSplit}/1`);
+  const year = interestedDate.getFullYear();
+  const month = interestedDate.getMonth() + 1;
+
+  const nextMonthDate = new Date(interestedDate.getTime());
+  nextMonthDate.setMonth(interestedDate.getMonth() + 1);
+  const nextYear = nextMonthDate.getFullYear();
+  const nextMonth = nextMonthDate.getMonth() + 1;
+
   // This generates YYYY-MM
-  const interested_year_month = `${year}-${month < 10 ? `0${month}` : month}`;
+  const interested_year_month_lte = `${nextYear}-${
+    nextMonth < 10 ? `0${nextMonth}` : nextMonth
+  }`;
+  const interested_year_month_gte = `${year}-${
+    month < 10 ? `0${month}` : month
+  }`;
+
   try {
     const result = await esClient.count({
       index: `router-logs-${openshiftProjectName}-*`,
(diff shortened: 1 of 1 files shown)


services/api/src/resources/environment/resolvers.js
```diff
@@ -364,8 +380,9 @@ const getEnvironmentHitsMonthByEnvironmentId = async (
               {
                 range: {
                   '@timestamp': {
-                    gte: `${interested_year_month}||/M`,
-                    lte: `${interested_year_month}||/M`,
+                    gte: `${interested_year_month_gte}||/M`,
+                    lte: `${interested_year_month_lte}||/M`,
```
Schnitzel: so according to https://codesandbox.io/s/condescending-boyd-vwpqn?fontsize=14

this will now be 
```
> 2019-12 
< 2020-01 
```
but `||/M` will then round down for `gte` and round up for `lte` (see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#range-query-date-math-rounding)

So I think in the end that query would actually cover two months, like `2019-12-01`-`2020-01-31`?

I think we would need to remove the `||/M` as we don't depend on ES doing the rounding anymore.

wintercreative: Interesting. That may have accounted for the discrepancy in the hits I was seeing. My concern with dropping the `||/M` is a note from the original issue

> A successful request needs to contain a full timestamp YYYY-MM-DDTHH:MM:SS rather than just the YYYY-MM

Instead of dropping the rounding, I think we can just add `-1` to the gte and lte dates we send. 

Example below and updated PR.

https://codesandbox.io/s/condescending-boyd-vwpqn?fontsize=14

justinlevi: I think I misunderstood how the rounding was working. I've updated this one more time. `|| /M removed`

justinlevi:
This PR is no longer needed

PR #1387, opened 2019-11-08
Corrects PHP-FPM to use host.docker.internal for xdebug

### LINKED ISSUE #1088: Xdebug host is not discovered correctly
**Describe the bug**
When `XDEBUG_ENABLE: "true"` is specified, the `xdebug. remote_host` is always set to `172.22.0.1`. 

Running the command from https://github.com/amazeeio/lagoon/blob/master/images/php/fpm/entrypoints/60-php-xdebug.sh#L6 _after_ container started returns correct results.

It looks like the timeout is too low to get this command running correctly during container starting.

**To Reproduce**
Steps to reproduce the behavior on MacOS:
1. Set `XDEBUG_ENABLE: "true"` in `docker-compose.yml`.
2. `docker-compose up -d --build`
3. Create a page with `phpinfo()` and run it.
4. Check the value of `xdebug.remote_host`. It will be `172.22.0.1`. 

**Expected behavior**
The value of `xdebug.remote_host` should be `docker.for.mac.localhost`.

It is possible to override this by setting `DOCKERHOST: "docker.for.mac.localhost"` in `docker-compose.yml`, but this defeats the purpose of the auto-discovery mechanism in https://github.com/amazeeio/lagoon/blob/master/images/php/fpm/entrypoints/60-php-xdebug.sh.

fubarhouse:
# Checklist

# Changelog Entry
Bugfix - Improves discoverability of Docker host for xdebug

# Closing issues
Closes #1088


images/php/fpm/entrypoints/60-php-xdebug.sh
```diff
@@ -3,20 +3,8 @@
...
-  # https://docs.docker.com/docker-for-windows/release-notes/#docker-community-edition-17060-win13-release-notes-2017-06-01-17060-rc1-ce-win13-edge
-  if busybox timeout -t 1 ping -c1 docker.for.win.localhost &> /dev/null; then
-    echo "docker.for.win.localhost"
-    return
-  fi
-
-  # https://github.com/amazeeio/pygmy/blob/267ba143158548628f190f05ecb5cb2c19212038/lib/pygmy/resolv_osx.rb#L26
-  if busybox timeout -t 1 ping -c1 172.16.172.16 &> /dev/null; then
-    echo "172.16.172.16"
+  if busybox nslookup host.docker.internal &> /dev/null; then
```
Schnitzel: as this is run inside an entrypoint and a DNS issue would block the whole container from coming up: can we add a `timeout` command in front of it with 1sec timeout? Just to be sure that we don't have any weird DNS issues affecting our containers :)

Schnitzel (changes requested):

images/php/fpm/entrypoints/60-php-xdebug.sh
fubarhouse: Switched to `busybox timeout 1 nslookup host.docker.internal`.

images/php/fpm/entrypoints/60-php-xdebug.sh
```diff
@@ -3,20 +3,8 @@
...
-  # https://docs.docker.com/docker-for-windows/release-notes/#docker-community-edition-17060-win13-release-notes-2017-06-01-17060-rc1-ce-win13-edge
-  if busybox timeout -t 1 ping -c1 docker.for.win.localhost &> /dev/null; then
-    echo "docker.for.win.localhost"
-    return
-  fi
-
-  # https://github.com/amazeeio/pygmy/blob/267ba143158548628f190f05ecb5cb2c19212038/lib/pygmy/resolv_osx.rb#L26
-  if busybox timeout -t 1 ping -c1 172.16.172.16 &> /dev/null; then
-    echo "172.16.172.16"
+  if busybox timeout 1 nslookup host.docker.internal &> /dev/null; then
```
Schnitzel:
```suggestion
  if busybox timeout 1 busybox nslookup host.docker.internal &> /dev/null; then
```
is making sure that we always run with busybox, as there could be other implementations of nslookup exist

Schnitzel (changes requested):

PR #1419, opened 2019-11-19
fixing postgres not starting issue

### LINKED ISSUE #1418: Postgres failing master build
master is currently failing with a postgres build, this prevents all PR builds to fail as well. fix asap

Schnitzel:
the newest postgres alpine version has a check if it is sourced or not: https://github.com/docker-library/postgres/blob/6dfdc0eacba0ae39b837df5eef63f89f13556e50/11/alpine/docker-entrypoint.sh#L274-L276
which is the case with our lagoon entrypoints system.
Instead of running it through the lagoon entrypoint system, we now execute the original entrypoint directly

closes #1418

### DIFF
--- a/images/postgres/Dockerfile
+++ b/images/postgres/Dockerfile
@@ -24,7 +24,6 @@ RUN chmod g+w /etc/passwd \
 
 ENV LAGOON=postgres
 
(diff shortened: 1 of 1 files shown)


tobybellwood:
w00t!  Tests passed 🎉

tobybellwood (approved):
Tests passed ok

PR #1423, opened 2019-11-20
Notifications updates

### LINKED ISSUE #1272: actions triggered by API do not have notifications
When triggering deployments/promotions/deletions via the graphql API, no notification is published.
`api:deployEnvironmentBranch`
`api:deployEnvironmentLatest`
`api:deleteEnvironment`
are known, check resolvers.js for more

### LINKED ISSUE #1274: `deleteNotificationRocketChat` searches for type `slack`
at

https://github.com/amazeeio/lagoon/blob/5f04751db22c9400273378e93596f1093a07c5eb/services/api/src/resources/notification/resolvers.js#L100

the type should be `rocketchat`

### LINKED ISSUE #1319: suppress pull request related notifications
If a project has `pullRequests: false` set in the API, an opening/closing/merging of PRs should not trigger messages into the notifications of that project

twardnw:
Cleanup of several notifications-related tickets

closes #1272
closes #1274 
closes #1319


node-packages/commons/src/tasks.js
```diff
@@ -394,7 +408,100 @@ async function createRemoveTask(removeData: Object) {
 
   switch (project.activeSystemsRemove) {
     case 'lagoon_openshiftRemove':
-      return sendToLagoonTasks('remove-openshift', removeData);
+    case 'lagoon_openshiftBuildDeploy':
```
Schnitzel: I think this `case 'lagoon_openshiftBuildDeploy':` is too much here, it will never end up here, as `project.activeSystemsRemove` is never defined as this.

node-packages/commons/src/tasks.js
```diff
@@ -394,7 +408,100 @@ async function createRemoveTask(removeData: Object) {
...
+              } did not match branchname, not deploying`,
+            );
+            throw new NoNeedToDeployBranch(
+              `configured regex '${
+                project.branches
+              }' does not match branchname '${branchName}'`,
+            );
+          }
+        }
+      } else if (type === 'pullrequest') {
```
Schnitzel, line 456: this is missing a `type === 'promote'` case, which means that environments of type `promote` are never deleted and therefore the CI fails. As `promote` environments are always created manually you can just assume that we can delete them like:

```
} else if (type === 'promote') {
  return sendToLagoonTasks('remove-openshift', removeData);
}
```

Schnitzel (changes requested):

PR #1456, opened 2019-12-04
Add storybook to Lagoon UI

### LINKED ISSUE #1279: Use Storybook for UI development/refactoring
[Storybook](https://storybook.js.org) is a very useful tool for UI development. It displays designed elements in a style guide that makes it easier to:

- build independent components
- document/find reusable elements
- refactor without visual regressions

JohnAlbin:
# Checklist

From the new `services/ui/README.md`:

Storybook has been added to `services/ui`. Storybook is a set of web pages that can be browsed to see an example of all the components in the Lagoon UI.

To build a static set of webpages in `services/ui/storybook-static`, run:
```
cd services/ui
yarn install
yarn build-storybook
```
The static HTML files can be served through any method or with the provided `yarn serve-storybook`.

To run the Storybook development server (with hot-loading on file change), run:
```
cd services/ui
yarn install
yarn storybook
```

# Known issues

Currently, there are 5 components that are misbehaving inside Storybook. They will be fixed in follow-ups.

1. **Pages / Deployments**: GraphQL subscriptions don't currently work in our mocked Storybook environment. The browser freezes when `subscribeToMore()` is run, so the components can't be shown until that is fixed.
2. **Pages / Tasks**: same as above.
3. **Pages / Environment / Default**: When navigating to this page, it shows a "Internal Server Error". However, when the page is reloaded, the component is displayed without error. I've added a note in Storybook to explain this under **Pages / Environment / `@TODO`**.
4. **Components / DeployLatest**: A few of the stories under this heading are marked `@TODO` because their state is dependent on a GraphQL Mutation that I don't know yet how to mock properly in Storybook.
5. **Components / ProjectDetailsSidebar**: The component depends on CSS in the `Project` component to display properly. Both of these components need some refactoring to fix this.

Additionally, several components use React's `UseState` to track internal state. If you interact with those components in Storybook, you can get them to display those states, but you can't have Storybook show you those states on initial page load since Storybook can't mock the internal state.

# Changelog Entry
Improvement - UI is now documented in Storybook (#1279)

# Closing issues
closes #1279


JohnAlbin:
The most interesting file from this PR is going to be [services/api/src/mocks.js](https://github.com/amazeeio/lagoon/blob/d6a23678eb0cc68d4b09e8132a6dc484987dfbba/services/api/src/mocks.js), which makes mocked Apollo resolvers for all `type`s in `typedefs.js`.

simesy:
Cool :)

services/ui/.storybook/decorators/ApiConnection.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import { AuthContext } from 'lib/Authenticator';
+import { ApolloClient } from 'apollo-client';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { SchemaLink } from 'apollo-link-schema';
+import { ApolloProvider } from 'react-apollo';
+import { makeExecutableSchema, addMockFunctionsToSchema } from 'graphql-tools';
+import typeDefs from 'api/src/typeDefs';
```
rocketeerbkw, line 9: How does this translate to the correct `/services/api` path on disk? In all the other services the way to reference code outside the service folder was to import a module, eg `@lagoon/commons`.

A related question is... since only the `ui` folder is copied into the container, storybook would not be able to run inside docker I presume? Is it only useful to run locally or would we want to consider building storybook inside the containers also?

JohnAlbin: > How does this translate to the correct /services/api path on disk? In all the other services the way to reference code outside the service folder was to import a module, eg @lagoon/commons.

Honestly, I was scratching my head on this one too. PHPStorm rewrote the import path when I was moving files around. And, yet, it still worked fine from my local command line. There's no … config that I saw that referenced the services/api directory, so I have no idea why it actually works.

What's the proper way to import a file from one `/services` directory to another?

> A related question is... since only the ui folder is copied into the container, storybook would not be able to run inside docker I presume?

If we can do a proper import of typedefs and mocks, then we can.

rocketeerbkw: What's the _proper_ way? Excellent question. The _current_ way is to add code that needs to be in multiple services to the `@lagoon/commons` package in `node-packages/commons` and that can be imported anywhere it's needed.

Technically the API doesn't use the mocks, so it could go in the ui only. But that seems weird. Open to suggestions. This can also be a follow up issue with the other known issues.

rocketeerbkw (approved):
Overall LATM (looks awesome to me), with known issues that can have follow-ups

PR #1461, opened 2019-12-05
Stub file timeout

### LINKED ISSUE #1453: aliases.drushrc.php should fail more gracefully if api is not available
**Describe the bug**
@shreddedbacon  did already some work on an updated version of the drush aliases file that fails more gracefully in case of the lagoon api not being reachable

**Expected behavior**
- Unavailability of API does not lead to cronjobs failing due to a long aliases timeout

shreddedbacon:
# Checklist

Sometimes if the API is unavailable, drush will take a long time trying to get a token and eventually fail. This can cause issues with cronjobs and also some users that may be unable to generate a token.

This change defaults to a 10 second timeout, but it is adjustable client side with specifying an envvar `LAGOON_OVERRIDE_SSH_TIMEOUT=10` or setting in `ssh_port_timeout: 10` in `.lagoon.yml`

# Changelog Entry

Improvement - The drush aliases stub file will timed out if api/ssh is unavailable

# Closing issues
closes #1453


shreddedbacon:
This only solves part of the problem at the moment, changing to draft

Schnitzel:
@shreddedbacon 
is this a draft? or do you think we can merge that?

shreddedbacon:
I'll review this today

shreddedbacon:
The other part of this has been added to the example drush aliases.
Basically checks that the aliases stub file is even accessible before attempting to use it, this will fail sooner if the stub file is completely unavailable like in the event that drush-aliases service is down or there is no active connection on the internet

PR #1489, opened 2019-12-12
Fix regex to handle all possible IP addresses

smlx:
# Checklist

If you run `make minishift` more than once then the regex may not match if the last octet of the minishift IP doesn't have 3 digits. This PR updates the regex to handle all possible IPs.

# Changelog Entry
Improvement - increase robustness of build scripts for lagoon development environment

# Closing issues
n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -793,11 +793,11 @@ endif
 ifeq ($(ARCH), darwin)
 	@OPENSHIFT_MACHINE_IP=$$(./local-dev/minishift/minishift --profile $(CI_BUILD_TAG) ip); \
 	echo "replacing IP in local-dev/api-data/01-populate-api-data.gql and docker-compose.yaml with the IP '$$OPENSHIFT_MACHINE_IP'"; \
-	sed -i '' -e "s/192.168\.[0-9]\{1,3\}\.[0-9]\{3\}/$${OPENSHIFT_MACHINE_IP}/g" local-dev/api-data/01-populate-api-data.gql docker-compose.yaml;
+	sed -i '' -e "s/192.168\.[0-9]\{1,3\}\.\([2-9]\|[0-9]\{2,3\}\)/$${OPENSHIFT_MACHINE_IP}/g" local-dev/api-data/01-populate-api-data.gql docker-compose.yaml;
 else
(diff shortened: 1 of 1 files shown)


Schnitzel:
@smlx I don't think that is correct, the IP that is in all files that should be replaced is `192.168.42.100`, for example here: https://github.com/amazeeio/lagoon/blob/master/docker-compose.yaml#L235

But your regex now also matches `192.168.42.1` which should not be replaced, example: https://github.com/amazeeio/lagoon/blob/master/docker-compose.yaml#L287 and so the tests fail.

can you update the regex so it matches, `192.168\.[0-9]{1,3}\.[0-9]{1,3}` but NOT `192.168.42.1` ?

smlx:
ah, yeah that's a problem. I'll fix it up.

smlx:
ok, should be good to go

shreddedbacon:
@smlx this is broken with darwin :(
The version of sed installed doesn't support the regex pattern properly.

```
$ sed --version
sed: illegal option -- -
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

The `-E` flag seems to support regex rather than `-e`
```
-E      Interpret regular expressions as extended (modern) regular expressions rather than basic regular expressions (BRE's).  The re_format(7) manual page fully describes both formats.
-e command
             Append the editing commands specified by the command argument to the list of commands.
```


So this works when run outside of make
```
sed -i '' -E "s/192.168\.[0-9]{1,3}\.([2-9]|[0-9]{2,3})/192.168.42.100/g" local-dev/api-data/01-populate-api-data.gql docker-compose.yaml
```
Changing the make file to use `-E` fails though, possibly the escaping breaks it?
```
sed -i '' -E "s/192.168\.[0-9]\{1,3\}\.\([2-9]\|[0-9]\{2,3\}\)/$${OPENSHIFT_MACHINE_IP}/g" local-dev/api-data/01-populate-api-data.gql docker-compose.yaml;
```
As it works when not escaped like this in the make file
```
sed -i '' -E "s/192.168\.[0-9]{1,3}\.([2-9]|[0-9]{2,3})/$${OPENSHIFT_MACHINE_IP}/g" local-dev/api-data/01-populate-api-data.gql docker-compose.yaml;
```

smlx:
@shreddedbacon sorry about that D: could you fix it up in a new PR?

PR #1491, opened 2019-12-12
Automated billing helpers fix

wintercreative:
# Checklist

The helper scripts for returning things like all projects not in billing groups were failing. This fixes that. 

# Changelog Entry
Automated billing helper script update


rocketeerbkw (changes requested):
It looks like there is a similar issue in https://github.com/amazeeio/lagoon/blob/master/services/api/src/models/project.ts#L33, which is only called from `billingGroupsCost` https://github.com/amazeeio/lagoon/blob/master/services/api/src/models/group.ts#L628. Do you wanna fix/test that in this PR also?

wintercreative:
Updated.

PR #1497, opened 2019-12-17
Fix reference to old testing subnet

smlx:
# Checklist

Fix an instance of the old local testing IP address that was missed in the recent change.

# Changelog Entry
This is basically a typo, so don't think it requires a changelog entry.

# Closing issues
n/a


tobybellwood (approved):
straightforward change

PR #1503, opened 2019-12-18
Sync output of parallel make tasks in CI

smlx:
# Checklist

The parallel tasks in CI (`-jN`) all write to STDOUT at once and the output is garbled. This change syncs the output to avoid the interspersed messages from parallel make tasks.

# Changelog Entry
Improvement - Sync output of parallel CI tasks

# Closing issues
n/a

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,6 +9,15 @@ node {
       try {
         env.CI_BUILD_TAG = env.BUILD_TAG.replaceAll('%2f','').replaceAll("[^A-Za-z0-9]+", "").toLowerCase()
         env.SAFEBRANCH_NAME = env.BRANCH_NAME.replaceAll('%2f','-').replaceAll("[^A-Za-z0-9]+", "-").toLowerCase()
+        env.SYNC_MAKE_OUTPUT = 'target'
+        // make/tests will synchronise (buffer) output by default to avoid interspersed
+        // lines from multiple jobs run in parallel. However this means that output for
+        // each make target is not written until the command completes.
+        //
+        // See `man -P 'less +/-O' make` for more information about this option.
+        //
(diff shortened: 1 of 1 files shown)


tobybellwood (approved):
much easier - nice!

Schnitzel:
This is definitely much easier to read, but unfortunately this biffers the logs while they are happening and only paste them after they are done. 
And many times to debug I need realtime logs to see check which test is maybe failing to login into the worker node and check stuff on the CLI.
If make is buffering the logs this won't be possible anymore :(

Is there maybe a possibility that we run multiple `sh` parallel steps of Jenkins which each runs just a single make test? I guess the question is then how do we explain Jenkins it should only run two tests parallel and not all of them? But I'm sure @smlx you will find a clever way :)

Schnitzel (changes requested):
see comment above

smlx:
hmm. okay so I've done a bit of research. `make`'s `sync-output` functionality uses `tmpfile()`, which uses `open(..., O_TMPFILE)`. This means that the output buffer never actually gets a pathname so we can't tail it.

I think that re-implementing `make`'s parallel functionality in the `Jenkinsfile` will get messy..

maybe the best way to do this is to have a variable in the `Makefile` that can be set to control `sync-output` functionality. By default output can be synced, and you can flip the variable to disable syncing when debugging. What do you think of that solution @Schnitzel?

Schnitzel:
ugh, yea looks like Jenkins doesn't really have a way to do this, see also https://issues.jenkins-ci.org/browse/JENKINS-44085
I guess your suggestion @smlx with a variable would be the easiest, can you add this to this PR?

smlx:
ready for review again @Schnitzel

Schnitzel (approved):

PR #1396, opened 2019-11-13
API Automated Billing Calculations

justinlevi:
# Checklist
- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated.
- [ ] Changelog entry has been written

Implements API BillingCost Calculations for lagoon projects.

To run the billing calculations unit tests, on your host machine:
```
$ cd services/api
$ yarn
$ yarn test billingCalculations
```

You can run individual tests by filtering via the tags at the end of each test description:
```
$ yarn test billingCalculations -t "#storage #USD #FC"
```

To test the billing calculation resolvers:
```
$ cd services/api && yarn test resolvers -t "#billing"
```

# Changelog Entry
Adding in lagoon billing calculations to the API

PR #1505, opened 2019-12-19
Billing Cost Modifiers

### LINKED ISSUE #1504: Billing costs need a modifier in order to handle non-standard billing edge cases
Billing cannot be fully automated until all edge cases are handled. A Billing Modifier persistent storage is needed to make this happen.

justinlevi:
# Checklist

Many billing situations arise that need to have temporary and long term modifiers applied. This is the approach to handle this.

# Changelog Entry
Improvement - Add billing modifier CRUD graphql mutations/queries. ( #1396 )

# Closing issues
closes #1504

### DIFF
--- a/services/api-db/docker-entrypoint-initdb.d/00-tables.sql
+++ b/services/api-db/docker-entrypoint-initdb.d/00-tables.sql
@@ -90,6 +90,20 @@ CREATE TABLE IF NOT EXISTS project (
   private_key                      varchar(5000)
 );
 
+CREATE TABLE IF NOT EXISTS billing_modifier (
+  id                              int NOT NULL auto_increment PRIMARY KEY,
+  group_id                             varchar(36),
+  weight                          int NOT NULL DEFAULT 0,
+  start_date                      datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  end_date                        datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  discount_fixed                  DECIMAL NULL DEFAULT 0.0,
+  discount_percentage             FLOAT NULL DEFAULT 0.0,
+  extra_fixed                     DECIMAL NULL DEFAULT 0.0,
+  extra_percentage                FLOAT NULL DEFAULT 0.0,
+  customer_comments               text,
+  admin_comments                  text
+);
+
 CREATE TABLE IF NOT EXISTS environment (
   id                     int NOT NULL auto_increment PRIMARY KEY,
   name                   varchar(100),
@@ -169,16 +183,16 @@ CREATE TABLE IF NOT EXISTS environment_service (
 );
 
 CREATE TABLE IF NOT EXISTS task (
-       id           int NOT NULL auto_increment PRIMARY KEY,
-       name         varchar(100) NOT NULL,
-       environment  int NOT NULL REFERENCES environment (id),
-       service      varchar(100) NOT NULL,
-       command      varchar(300) NOT NULL,
-       status       ENUM('active', 'succeeded', 'failed') NOT NULL,
-       created      datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
-       started      datetime NULL,
-       completed    datetime NULL,
-       remote_id    varchar(50) NULL
+  id           int NOT NULL auto_increment PRIMARY KEY,
+  name         varchar(100) NOT NULL,
+  environment  int NOT NULL REFERENCES environment (id),
+  service      varchar(100) NOT NULL,
+  command      varchar(300) NOT NULL,
+  status       ENUM('active', 'succeeded', 'failed') NOT NULL,
+  created      datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  started      datetime NULL,
+  completed    datetime NULL,
+  remote_id    varchar(50) NULL
 );
 
 CREATE TABLE IF NOT EXISTS s3_file (
--- a/services/api/src/helpers/billingGroups.ts
+++ b/services/api/src/helpers/billingGroups.ts
@@ -13,22 +13,9 @@
 
 import * as R from 'ramda';
 import projectHelpers from '../resources/project/helpers';
-//import { logger } from '@lagoon/commons/src/local-logging';
 import { getSqlClient, USE_SINGLETON } from '../clients/sqlClient';
 import { getKeycloakAdminClient } from '../clients/keycloak-admin';
 import { Group, BillingGroup } from '../models/group';
-// import { keycloakAdminClient } from '../clients/keycloakClient';
-
-const keycloakAuth = {
-  username: 'admin',
-  password: R.pathOr(
-    '<password not set>',
-    ['env', 'KEYCLOAK_ADMIN_PASSWORD'],
-    process,
-  ) as string,
-  grantType: 'password',
-  clientId: 'admin-cli',
-};
 
 interface IGroup {
   name: string;
--- a/services/api/src/helpers/billingGroups.ts
+++ b/services/api/src/helpers/billingGroups.ts
@@ -13,7 +13,7 @@
 
 import * as R from 'ramda';
 import projectHelpers from '../resources/project/helpers';
-import { getSqlClient, USE_SINGLETON } from '../clients/sqlClient';
+import { getSqlClient } from '../clients/sqlClient';
 import { getKeycloakAdminClient } from '../clients/keycloak-admin';
 import { Group, BillingGroup } from '../models/group';
 
@@ -29,7 +29,7 @@ interface IGroup {
 
 export const getAllProjectsNotInBillingGroup = async () => {
   const keycloakAdminClient = await getKeycloakAdminClient();
-  const sqlClient = getSqlClient(USE_SINGLETON);
+  const sqlClient = getSqlClient();
   const GroupModel = Group({keycloakAdminClient });
 
   // GET ALL GROUPS
--- a/services/api/src/models/billing.ts
+++ b/services/api/src/models/billing.ts
@@ -1,11 +1,8 @@
 import moment from 'moment';
 import R from 'ramda';
-import { Group, BillingGroup } from './group';
-import { getKeycloakAdminClient } from '../clients/keycloak-admin';
-import { getSqlClient, USE_SINGLETON } from '../clients/sqlClient';
 import { query } from '../util/db';
 import Sql from '../resources/billing/sql';
-import { GroupInput } from "./group";
+import { GroupInput, Group, BillingGroup } from "./group";
 
 export interface BillingModifierBase {
   id?: number;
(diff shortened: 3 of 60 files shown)


services/api/src/resources/billing/helpers.ts
```diff
@@ -56,11 +64,90 @@ const availabilityFilterFn = filterKey => ({ availability }) =>
...
+  await query(sqlClient, sql);
+  return 'success';
+}
+
+export default {
+  extractMonthYear,
+  projectWithBillingDataFn,
+  getProjectsData,
+  availabiltyProjectsCosts,
+  handleAddBillingModifier,
```
rocketeerbkw: the helpers.js files for other resources are all what I consider poor models. Since all the handle* functions here are only used in the billingModifier model, lets just inline these and get rid of helper files eventually.

services/api/src/resources/billing/helpers.ts
```diff
@@ -56,11 +64,90 @@ const availabilityFilterFn = filterKey => ({ availability }) =>
...
+  const sql = Sql.deleteAllBillingModifiersByBillingGroup(gid);
+  await query(sqlClient, sql);
+  return 'success';
+}
+
+export default {
+  extractMonthYear,
+  projectWithBillingDataFn,
+  getProjectsData,
+  availabiltyProjectsCosts,
```
rocketeerbkw: Misspelled availability, but looks like this function isn't used anywhere else? or maybe it's not part of this diff...

services/api/src/models/billing.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+export interface AddBillingModifierInput {
+  input: BillingModifierInput;
+}
+
+export interface UpdateBillingModifierInput {
+  input: {
+    id: number;
+    patch: BillingModifier;
```
rocketeerbkw: This, and the other interfaces, are tightly coupling the model to the graphql API. If we ever wanted to use the model outside of a graphql resolver (like in a script, or maybe we need to sync billing groups to/from another system in the background), we'd need to create an object that resembles a graphql request first. And if the API ever changes, that might require changes to non-api sections.

I would prefer to keep the typings separate between the API and the model. See the `updateGroup` groups model and and resolvers for example. It's more verbose, but the resolver function I think should be the place where we translate API objects into the model objects.

services/api/src/models/billing.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * Create/Add Billing Modifier
+ *
+ * @param {AddBillingModifierInput} input The modifier values
+ *
+ * @return {BillingModifier} The created modifier
+ */
+export const addBillingModifier = async (input: BillingModifierInput) => {
+  const { group, billingModifier } = await prepareAddBillingModifierFromInput(
```
rocketeerbkw: Related to previous comment: all these prepare* functions feel like they should be in the resolver, since they are mapping the API object into the expected model object.

services/api/src/resources/billing/resolvers.ts
```diff
@@ -0,0 +1,136 @@
...
+ *     { sqlClient, hasPermissions, keycloakGrant, requestCache }
+ *
+ * @return {String} Success
+ */
+export const updateBillingModifier = async (_, args: UpdateBillingModifierInput, context: {models: any, hasPermission: any}) => {
+  const { hasPermission } = context;
+  const { input : { id, patch } } = args;
+
+  // Permissions
+  await hasPermission('group', 'update');
```
rocketeerbkw: For most permissions checks, keycloak needs to know more context (attributes). In this case, it needs to know what group is being checked, eg https://github.com/amazeeio/lagoon/blob/master/services/api/src/resources/group/resolvers.ts#L159

services/api/src/resources/billing/resolvers.ts
```diff
@@ -0,0 +1,136 @@
...
+  _,
+  args: DeleteBillingModifierInput,
+  context: {models: any,
+  hasPermission: any}
+) => {
+  const { hasPermission } = context;
+  const { input: { id } } = args;
+
+  // Permissions
+  await hasPermission('group', 'delete');
```
rocketeerbkw: needs group id attribute

services/api/src/resources/billing/resolvers.ts
```diff
@@ -0,0 +1,136 @@
...
+  _,
+  args: DeleteAllBillingGroupModifiersInput,
+  context: {models: any,
+  hasPermission: any}
+) => {
+  const { hasPermission } = context;
+  const { input } = args;
+
+  // Permissions
+  await hasPermission('group', 'delete');
```
rocketeerbkw: There is a specific "deleteAll" scope that is currently granted only to Admins. The main reason we have deleteAll* api mutations is to support the `local-api-data-watcher-pusher` service. These generally aren't ever used on a regular basis.

So... do you think it's useful to let "normal" users delete all billing modifiers for a group (in which case, it needs the group id attribute)? The alternative is to change this mutation to delete all modifiers (for all groups) and switch to `deleteAll` scope.

services/api/src/resources/billing/resolvers.ts
```diff
@@ -0,0 +1,136 @@
...
+ *     { sqlClient, hasPermissions, keycloakGrant, requestCache }
+ *
+ * @return {[BillingModifier]} All modifiers associated to the billing Group
+ */
+export const getBillingModifiers = async (_, args: BillingModifiersInput, context: {models: any, hasPermission: any}) => {
+  const { hasPermission } = context;
+  const { input: groupInput, month } = args;
+
+  // Permissions
+  await hasPermission('group', 'view');
```
rocketeerbkw: There is no `view` scope for groups. All users who are a member of a group are assumed to be able to view it. So here, you'd want to check if the user can view groups they're not a member of (viewAll scope) and if not, check if they're a member of the group given in input. See [`getGroupsByProjectId`](https://github.com/amazeeio/lagoon/blob/master/services/api/src/resources/group/resolvers.ts#L63) for example

services/api/src/resources/billing/resolvers.ts
```diff
@@ -0,0 +1,136 @@
+import * as R from 'ramda';
+import billingModel, { AddBillingModifierInput, BillingModifier, BillingModifiersInput, UpdateBillingModifierInput, DeleteBillingModifierInput, DeleteAllBillingGroupModifiersInput } from '../../models/billing';
```
rocketeerbkw: line is too long, I only bring it up to ask if you have formatting setup in vscode

services/api/src/typeDefs.js
```diff
@@ -562,6 +577,10 @@ const typeDefs = gql`
     Returns the costs for all billing groups
     """
     allBillingGroupsCost(month: String!): JSON
+    """
+    Returns the Billing Group Modifiers for a given Billing Group (all if no month filter defined)
+    """
+    billingModifiers(input: GroupInput!, month: String): [BillingModifier]
```
rocketeerbkw: I think the naming convention here would be `allBillingModifiers`

services/api/src/typeDefs.js
```diff
@@ -562,6 +577,10 @@ const typeDefs = gql`
     Returns the costs for all billing groups
     """
     allBillingGroupsCost(month: String!): JSON
+    """
+    Returns the Billing Group Modifiers for a given Billing Group (all if no month filter defined)
```
rocketeerbkw: `all if no month filter defined` < should that be "all if no GroupInput filter is defined"?

services/api/src/typeDefs.js
```diff
@@ -1006,6 +1025,43 @@ const typeDefs = gql`
     parentGroup: GroupInput
   }
 
+
+
+  input BillingModifierInput {
```
rocketeerbkw: Naming convention here would be AddBillingModifierInput

rocketeerbkw (changes requested):

services/api-db/docker-entrypoint-initdb.d/00-tables.sql
```diff
@@ -90,6 +90,20 @@ CREATE TABLE IF NOT EXISTS project (
   private_key                      varchar(5000)
 );
 
+CREATE TABLE IF NOT EXISTS billing_modifier (
+  id                              int NOT NULL auto_increment PRIMARY KEY,
+  gid                             varchar(36),
```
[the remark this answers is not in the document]
justinlevi: ahh, good catch. I was totally falling back to how I've done it in the past. I'm all for sticking with conventions, and I like `group_id` a bit more than gid, so let's go with that. 

Addressed here: 
https://github.com/justinlevi/lagoon/commit/3461eb8fd7f95ef26e3fa595dd5425c142bfdff1

services/api/src/resources/billing/helpers.ts
justinlevi: I hear what you're saying, definitely makes sense. In going back and looking at the "handler" functions now, that all feels a bit odd and overly abstracted. 

When I see "model" files, I think more MVC though and we don't really have any consistent controllers. 

I'm fine with inlining these, but let's talk more about putting an aspirational design pattern in practice, and maybe adding some notes in the developer docs. 

addressed in: 7cf799df07ea21fd4833aef2a67a6ef830302a6b

services/api/src/resources/billing/helpers.ts
justinlevi: ugh, good catch. thanks. 
346322264e6fa90aa9afda5899b90784a671f386

services/api/src/models/billing.ts
justinlevi: yeah, this was my attempt at strongly typing everything I was working on. I do think we should avoid "any" typing if possible. The code duplication between the API SDL and Typescript could be addressed with an ORM. 

I'll need to think about this one a bit more. The current convention and ultimate direction we're headed with Typescript isn't super clear to me yet.

services/api/src/resources/billing/resolvers.ts
justinlevi: This was definitely added to help me with tests with a very slight potential that this could be useful... Not sure it's needed though

services/api/src/resources/billing/resolvers.ts
justinlevi: prettier was giving me a nightmare formatting things wrong so I turned it off. I'll try and get that setup again. 

Addressed in: 5d93a8c42e6c30f0b7b438dc3e4047400746ca9e

services/api/src/resources/billing/helpers.ts
rocketeerbkw: They're called models because that's what the apollo docs called them, not related to MVC. Of course, now the docs are updated and they don't use that name anymore 🤷

Let's have a chat about what problem I'm trying to solve with the model concept and that probably clears up any questions

services/api/src/typeDefs.js
justinlevi: Addressed in 82c1c2134ef0c3f6097293faabbf95ce44e64346

services/api/src/typeDefs.js
justinlevi: Hmm, my thought was that a user would want all billing modifiers returned for a Billing Group if they omit a month input variable. I can't think of a use case, outside of internal audits, where we'd want to get/display all billing modifiers for all groups. 

Now that I'm thinking about it, that's why I originally named this as `billingModifiers` and not `allBillingModifiers` because we were returning the billingModifiers for a specific Billing Group and not all Modifiers for all groups. 

It might make sense to provide queries for both, I'll need to think on that more.

I can definitely make the comment more clear though...

services/api/src/models/billing.ts
justinlevi: Addressed in 4a6a75dfaeb3cbcb6da384581a440223fd5d83cb

services/api/src/typeDefs.js
justinlevi: addressed in: 18a1d2012c92ca915b841bc0ef6df11874840484

services/api/src/resources/billing/resolvers.ts
justinlevi: Addressed in bcb60674748dc28ef2fa5f75027f9c618e78950c

services/api/src/typeDefs.js
justinlevi: Addressed

justinlevi:
@rocketeerbkw In general, I'd like to talk about the permissions for this PR as they don't really seem to fit into any of the other existing permissions situations as I see it. As far as I understand it, the modifiers should really only be `readonly` for anyone but the lagoon admin users. Nobody should have the ability to add/edit/delete except for whoever is in charge of the billing tools. I'm not sure we have the current scopes/roles setup for that

services/api/src/resources/billing/billingCalculations.ts
```diff
@@ -75,22 +85,35 @@ export const calculateProjectEnvironmentsTotalsToBill = environments => {
...
 
   const environmentCost = { prod, dev };
-  const total = hitCost + storage + prod + dev;
-  return {
+
+  // Apply Modifiers
+  let total = hitCost + storage + prod + dev;
+  modifiers
+    .sort((a:BillingModifier, b:BillingModifier) => a.weight < b.weight? -1 : 1)
+    .forEach(modifier => {
```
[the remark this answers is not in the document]
justinlevi: Addressed in 664416d8f872ff4a6ce99cec2b68e729e95369bf

services/api/src/resources/billing/resolvers.ts
justinlevi: Addressed in: bcb60674748dc28ef2fa5f75027f9c618e78950c although I'd like to talk through all of the permissions related to the Billing Modifiers a bit more

services/api/src/resources/billing/resolvers.ts
justinlevi: Addressed in bcb60674748dc28ef2fa5f75027f9c618e78950c but let's talk more about permissions.

services/api/src/resources/billing/resolvers.ts
justinlevi: Addressed in bcb60674748dc28ef2fa5f75027f9c618e78950c but let's talk more about permissions.

services/api/src/models/billing.ts
justinlevi: @rocketeerbkw I think this should be addressed in a6acb58b511b54795590fb88783d36066efc7caf

services/api/src/models/billing.ts
```diff
@@ -0,0 +1,149 @@
...
+  weight?: number;
+}
+export interface BillingModifier extends BillingModifierBase{
+  group?: Group;
+}
+
+/**
+ * Create/Add Billing Modifier
+ *
+ * @param {AddBillingModifierInput} input The modifier values
```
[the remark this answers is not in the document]
rocketeerbkw: there are more in the rest of the file too

services/api/src/resources/billing/resolvers.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (R.isEmpty(R.omit(['group', 'startDate', 'endDate'], input))) {
+    throw new Error('You must provide a discount value or extra cost.');
+  }
+
+  if (new Date(input.startDate) >= new Date(input.endDate)) {
+    throw new Error('You must provide a start date before the end date.');
+  }
+
+  // Permissions
+  await hasPermission('group', 'add');
```
rocketeerbkw: We talked about how we probably need a whole new set of resources/scopes for billing modifiers. Just noting here for paper trail.

rocketeerbkw (changes requested):
Last issue is the permission stuff, like we already talked about.

services/api/src/models/billing.ts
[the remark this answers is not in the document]
justinlevi: Ahh, good catch. Would be nice if VSCode threw a notice if JSDoc comments didn't match up. Might be an extension for that. I'll do a bit of searching

services/api/src/models/billing.ts
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * @return {BillingModifier} The created modifier
+ */
+export const addBillingModifier = async (modifier: BillingModifier) => {
```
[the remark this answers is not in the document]
justinlevi: Yeah, TS knows the return type because we cast `R.prop(0, rows)` : any to `BillingModifer`. As long as you type a return value TS is able to infer the type of that function whenever you call it elsewhere.

justinlevi:
> Last issue is the permission stuff, like we already talked about.

I think this is all set. @rocketeerbkw can you double check the hasPermissions in the resolvers is correctly implemented with the new `billing_modifier` scope/resource?

rocketeerbkw (approved):

justinlevi:
@rocketeerbkw Can you do another look over this? I cleaned up my commit history and added in the Admin UI work.

services/api/src/helpers/billingGroups.ts
```diff
@@ -42,7 +29,7 @@ interface IGroup {
 
 export const getAllProjectsNotInBillingGroup = async () => {
   const keycloakAdminClient = await getKeycloakAdminClient();
-  const sqlClient = getSqlClient(USE_SINGLETON);
+  const sqlClient = getSqlClient();
```
rocketeerbkw, line 32: w/o the singleton mechanic, any use of `getSqlClient()` outside the apollo context must explicitly close its own connection. I'm not sure if that still applies with this one-off, non-long-running script though. I would assume that when the node process ends, everything gets cleaned up, but I've been wrong about that before, and we clean up the connections in the other cli scripts (eg https://github.com/amazeeio/lagoon/blob/master/services/api/src/helpers/sync-groups-opendistro-security.ts#L41)

So to be explicit here, you'd add `sqlClient.destroy()` somewhere before this function ends. Fair warning with this change, for some reason the act of closing the connection keeps the node process running for at least a minute after all the work has been completed.

services/api/src/models/group.ts
```diff
@@ -53,38 +62,6 @@ interface AttributeFilterFn {
   (attribute: { name: string; value: string[] }, group: Group): boolean;
 }
 
-interface GroupModel {
```
rocketeerbkw, line 56: Was this redundant? I used it when creating the model to figure out what the API should look like and get autocomplete working

rocketeerbkw (changes requested):

services/api/src/helpers/billingGroups.ts
justinlevi: cool. will add this in. Not sure how often this helper will ever get run, but good to put this in.

services/api/src/models/group.ts
justinlevi: Yeah, if you ever need to get the type of the GroupModel, you can do: 

```
ReturnType<typeof GroupModel>
```

This will return whatever is defined in the function

justinlevi: I do that here in fact
https://github.com/justinlevi/lagoon/blob/dfade1133717f6bd1125dfa39416bf637c2c9a70/services/api/src/resources/billing/helpers.ts#L30

services/api/src/helpers/billingGroups.ts
justinlevi: fixed in upcoming PR

PR #1521, opened 2020-01-03
Removing Node versions 6, 8, and 9 and PHP versions 5.6, 7.0, and 7.1

cdchris12:
This PR removes support for old PHP and Node versions
 
# Checklist

# Changelog Entry
Change - Remove support for Node 6, 8, and 9 and for PHP 5.6, 7.0, and 7.1


seanhamlin:
Having this merged would have saved me a few hours as I went down a rabbit hole in #1554 to fix an issue with the PHP 7.0 images not building.

tobybellwood:
This looks really good - before approving, can we look at the following?

* should we not be also removing node 9 - as per https://nodejs.org/en/about/releases/ (potentially keeping node8 as it's closest to EOL)?

* In addition, there is some logic that documents deprecations in https://github.com/amazeeio/lagoon/blob/master/images/node/Dockerfile#L40-L44 that should also be updated?

cdchris12:
I've removed support for Node 9 and the nag notifications for depreciated versions in the Node Dockerfile.

smlx:
According to [this link](https://lucene.apache.org/solr/downloads.html#about-versions-and-support) all `solr` versions < 7.7 are also EOL - that means _none_ of the versions we currently define are in support :grimacing: . I wonder if bumping solr versions should go in this PR or a separate one?

tobybellwood:
@smlx - solr is a whole other story/PR - the alpine-based images we use are also EOL...

tobybellwood (approved):
This looks good to me now, and will unblock a lot of the downstream image issues we've seen.  We should flag the deprecation of Node 6/8/9 - we've already notified users about PHP5.6/7.0/7.1 in the last Lagoon 1.2.0 release

Schnitzel:
the `node6_subfolder` actually tested the capability to have a Dockerfile in a Subfolder. with removing the `node6` test we're removing also this test. I moved this test into the `features-subfolder` test now.

Schnitzel:
also I don't think we need to install xdebug in an older version anymore, as the newest version of xdebug supports 7.1 and newer: https://pecl.php.net/package/xdebug/2.9.1

PR #1525, opened 2020-01-06
Run the securityadmin demo script as part of make up

smlx:
# Checklist

When running `make up`, the `logs-db` container will not be available until the `securityadmin_demo.sh` script has been executed as per https://lagoon.readthedocs.io/en/latest/administering_lagoon/install/#opendistrosecurity

This change automates that step for local development.

It also removes the initial `docker-compose up` of `api-db` which doesn't seem to be necessary.

# Changelog Entry
Improvement - Automate Elastic OpenDistro Security setup for local lagoon development

# Closing issues
n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -749,9 +749,9 @@ logs:
 
 # Start all Lagoon Services
 up:
-	IMAGE_REPO=$(CI_BUILD_TAG) docker-compose -p $(CI_BUILD_TAG) up -d api-db
-	sleep 20
 	IMAGE_REPO=$(CI_BUILD_TAG) docker-compose -p $(CI_BUILD_TAG) up -d
+	sleep 20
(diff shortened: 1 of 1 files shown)


shreddedbacon:
Might also be worth when starting minishift to also configure the external service with the following to leverage logging locally in the `lagoon` namespace

```
apiVersion: v1
kind: Service
metadata:
  name: application-logs
spec:
  externalName: 192.168.42.1
  sessionAffinity: None
  type: ExternalName
```

smlx:
@shreddedbacon as discussed lets put that in a separate PR

Schnitzel:
Actually I really would like this to run automatically also when you deploy Lagoon, as we had issues with Lagoon develop before, so I created #1548

PR #1526, opened 2020-01-06
Add workaround for missing localhost dns on linux

smlx:
# Checklist

The `docker.for.mac.localhost` hostname obviously is not available on linux. This PR implements the following workarounds:

* use bare keycloak service name in docker-compose.yaml where possible
* make keycloak service name configurable where necessary
* configure the keycloak service URL in the `make up` target on linux

# Changelog Entry
Improvement - implement configurable keycloak endpoint for linux lagoon development

# Closing issues
n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -835,7 +835,15 @@ logs:
 
 # Start all Lagoon Services
 up:
+ifeq ($(ARCH), darwin)
 	IMAGE_REPO=$(CI_BUILD_TAG) docker-compose -p $(CI_BUILD_TAG) --compatibility up -d
+else
+	# once this docker issue is fixed we may be able to do away with this
+	# linux-specific workaround: https://github.com/docker/cli/issues/2290
+	KEYCLOAK_URL=$$(docker network inspect -f '{{(index .IPAM.Config 0).Gateway}}' bridge):8088 \
+		IMAGE_REPO=$(CI_BUILD_TAG) \
+		docker-compose -p $(CI_BUILD_TAG) --compatibility up -d
+endif
 	grep -m 1 ".opendistro_security index does not exist yet" <(docker-compose -p $(CI_BUILD_TAG) logs -f logs-db 2>&1)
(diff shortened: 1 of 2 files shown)


tobybellwood:
Can someone else test this on a mac setup to ensure that the KEYCLOAK_URL isn't being set?

Schnitzel:
https://github.com/moby/moby/pull/40007 has been merged, but looks like `20.03.0` is not out yet.
@smlx can you check if we would still need this when docker `20.03.0` is out?

smlx:
There are several issues here.

* https://github.com/moby/moby/pull/40007 is the PR that has been merged, and it makes it possible to run `docker run --add-host host.docker.internal:host-gateway` to get the desired DNS entry. However it is _not_ enabled by default on linux. You need that argument.
* There's an issue here about enabling it by default: https://github.com/docker/cli/issues/2290
* There's another issue tracking this feature over here: https://github.com/docker/for-linux/issues/264
* Code freeze for docker 20.x is apparently happening on May 1 (https://github.com/moby/moby/issues/40838), which will be the first release with this feature.

So I guess once a 20.x docker is released we _should_ be able to avoid this workaround :tada:

smlx:
@Schnitzel I've been dealing with this issue in local development and it's a pain to cherry-pick this patch around between branches. I think maybe it should be merged because:

* We do a similar hack for darwin anyway e.g. https://github.com/amazeeio/lagoon/blob/master/Makefile#L1013
* without this patch, `logs-db-ui` currently spams the CI logs with `Keycloak still not running, waiting 5 seconds` for the duration of every test run.

If/when a better alternative becomes available in upstream docker we can use it. What do you think?

Schnitzel:
> I've been dealing with this issue in local development and it's a pain to cherry-pick this patch around between branches. I think maybe it should be merged

sounds good, let's to this. Can you fix the conflict? will merge after that.

PR #1531, opened 2020-01-06
Updating Python images to use latest version of Alpine.

cdchris12:
This PR changes our Python Dockerfile such that we can continue using the latest alpine image (3.11 currently)
 
# Checklist

# Changelog Entry
Change - Update Python images to use latest version of alpine

### DIFF
--- a/images/python/Dockerfile
+++ b/images/python/Dockerfile
@@ -1,7 +1,8 @@
+  
 ARG PYTHON_VERSION
 ARG IMAGE_REPO
 FROM ${IMAGE_REPO:-lagoon}/commons as commons
-FROM python:${PYTHON_VERSION}-alpine3.10
+FROM python:${PYTHON_VERSION}-alpine
 
 LABEL maintainer="amazee.io"
 ENV LAGOON=python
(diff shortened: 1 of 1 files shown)


smlx:
I'm not sure if we want to use bare `-alpine` because of issues like this (from the Alpine 3.11 [release notes](https://alpinelinux.org/posts/Alpine-3.11.0-released.html):
> Python 2 is deprecated. Majority of Python 2 packages was removed and will be completely removed in next release.

Would it be better to pin to a specific version (currently `-alpine3.11`) so that we can control the lagoon-wide base image upgrade to newer alpine version?

tobybellwood:
yup - the alpine version for python images is now pinned in the makefile (https://github.com/amazeeio/lagoon/blob/master/Makefile#L236) - i've personally no issue with pinning deprecated versions of python back to older Alpine releases.  Let's look at this as a broader rehash of python support (should we use pip or apk's etc).

Schnitzel:
@tobybellwood 
any decision what we should do here?

tobybellwood:
This PR is outdated (make now has py2/3 logic) - the core question is whether to use pip/apk, which is probably a seperate ticket.

PR #1542, opened 2020-01-13
Switch local-dev/git image to alpine

smlx:
# Checklist

Switch from a third-party image for git/ssh server used in testing, just use Alpine base images.

# Changelog Entry
Improvement - Reduce third-party dependencies on test images.

# Closing issues
n/a

### DIFF
--- a/local-dev/git/Dockerfile
+++ b/local-dev/git/Dockerfile
@@ -1,12 +1,23 @@
-FROM unixtastic/git-ssh-server
+FROM alpine:3.11
 
-RUN apt-get update && apt-get install -y nginx fcgiwrap
+RUN apk add --no-cache \
+      bash \
+      fcgiwrap \
+      git-daemon \
+      nginx \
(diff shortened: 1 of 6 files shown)


local-dev/git/Dockerfile
```diff
@@ -1,12 +1,23 @@
-FROM unixtastic/git-ssh-server
+FROM alpine:3.10
```
tobybellwood: Can we get this to 3.11 - the other images are mostly running on 3.11 now, alternatively, we could extend on amazeeio/commons to remove the need to manually increment?

```suggestion
FROM alpine:3.11
```

smlx: Good point, will bump the version.

smlx: version bumped

tobybellwood (approved):
approved - all other images are now at 3.11 too

PR #1552, opened 2020-01-16
 Update the SSH public key in Lagoon UI.

### LINKED ISSUE #1471: SSH key management for users in UI
A user should be able to manage their SSH keys via the Lagoon UI

justinlevi:
Adds a minimal ui to manage developer SSH Keys via the lagoon UI.

# Checklist

Managing sshkeys required graphql mutations to be run directly on the API which wasn't as user friendly as it could be. 

# Closing issues
`closes #1471`


services/ui/src/components/SshKeys/AddSshKey.js
```diff
@@ -0,0 +1,123 @@
...
+            return <div>Adding SSH Key...</div>;
+          }
+
+          return (
+            <div className="addNew">
+
+              { error ? <div className="error">{error.message.replace('GraphQL error:', '').trim()}</div> : "" } 
+
+              <div>
+                <label>SSH Key Name</label>
```
rocketeerbkw: Is there a reason there's no `htmlFor` attribute here? Accessibility is covered by `aria-labelledby` but I think this label wouldn't be click-selectable.

services/ui/src/components/SshKeys/AddSshKey.js
```diff
@@ -0,0 +1,123 @@
...
+                <input
+                  aria-labelledby="addSshKeyName"
+                  id="sshKeyName"
+                  name="sshKeyName"
+                  label='SSH Key Name'
```
rocketeerbkw: I don't think `label` is a valid `input` attribute

services/ui/src/components/SshKeys/AddSshKey.js
rocketeerbkw: it's added for `htmlFor="sshKeyType"` so I wasn't sure if this was intentionally left off

services/ui/src/components/SshKeys/AddSshKey.js
```diff
@@ -0,0 +1,123 @@
...
+              </div>
+
+              <div>
+                <label htmlFor="sshKeyType">SSH Key Type</label>
+                <select
+                  id="sshKeyType"
+                  name="sshKeyType"
+                  onChange={handleChange}
+                  aria-labelledby="addSshKeyType"
+                  label='SSH Key Type'
```
rocketeerbkw: I don't think `label` is a valid `input` attribute

services/ui/src/components/SshKeys/AddSshKey.js
rocketeerbkw: and, now that i'm looking closely, the aria label `addSshKeyName` doesn't match any IDs that I see, so that isn't working as intended either (I think)

services/ui/src/components/SshKeys/AddSshKey.js
```diff
@@ -0,0 +1,123 @@
...
+                />
+              </div>
+
+              <div>
+                <label htmlFor="sshKeyType">SSH Key Type</label>
+                <select
+                  id="sshKeyType"
+                  name="sshKeyType"
+                  onChange={handleChange}
+                  aria-labelledby="addSshKeyType"
```
rocketeerbkw: This is redundant with the `htmlFor` attribute on the label, and the ID here doesn't match any other element

services/ui/src/components/SshKeys/AddSshKey.js
```diff
@@ -0,0 +1,123 @@
...
+                  {['SSH_RSA', 'SSH_ED25519'].map(value => (
+                    <option key={value} value={value}>
+                      {value}
+                    </option>
+                  ))}
+                </select>
+              </div>
+
+              <div>
+                <label>SSH Key</label>
```
rocketeerbkw: Same comments re `htmlFor`, `aria-labelledby` and `label`

rocketeerbkw (approved):
Some minor non-functional potential issue (didn't test in-browser). Could easily be handled in a follow-up issue though.

Schnitzel:
@justinlevi 
can you have a look at the suggestions of @rocketeerbkw? and if you agree fix&commit them?
thanks

justinlevi:
All issues should be addressed in https://github.com/justinlevi/lagoon/commit/bc9701022b7768307b073f32c731aca00a492d37

PR #1554, opened 2020-01-17
Update New Relic PHP Agent to 9.6.0.255 and pin Xdebug to 2.6.0 for PHP 7.0

### LINKED ISSUE #1543: Update New Relic PHP Agent to 9.6.0.255
**Describe the bug**
Newrelic PHP Agent has been released - https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/php-agent-950252

seanhamlin:
# Checklist

Updated the version of the PHP Agent to be the latest stable release. https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes

Version 9.4 added support for PHP 7.4, and at the moment New Relic is in those images. So now it will actually work.

Had to also downgrade Xdebug to 2.6.0 on the PHP 7.0 images so that they would actually build.

# Closing issues

Closes #1543`.

### DIFF
--- a/images/php/fpm/Dockerfile
+++ b/images/php/fpm/Dockerfile
@@ -35,8 +35,11 @@ COPY 00-lagoon-php.ini.tpl /usr/local/etc/php/conf.d/
 COPY php-fpm.d/www.conf /usr/local/etc/php-fpm.d/www.conf
 COPY ssmtp.conf /etc/ssmtp/ssmtp.conf
 
-# Defining Versions - https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/ / https://docs.newrelic.com/docs/agents/php-agent/getting-started/php-agent-compatibility-requirements
-ENV NEWRELIC_VERSION=9.3.0.248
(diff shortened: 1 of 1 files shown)


smlx:
> It is strongly recommended that users use an explicit version number in their pecl install invocations to ensure proper PHP version compatibility (PECL does not check the PHP version compatiblity when choosing a version of the extension to install, but does when trying to install it).
> 
> For example, memcached-2.2.0 has no PHP version constraints (https://pecl.php.net/package/memcached/2.2.0), but memcached-3.1.4 requires PHP 7.0.0 or newer (https://pecl.php.net/package/memcached/3.1.4). When doing pecl install memcached (no specific version) on PHP 5.6, PECL will try to install the latest release and fail.
> 
> Beyond the compatibility issue, it's also a good practice to ensure you know when your dependencies receive updates and can control those updates directly.

https://hub.docker.com/_/php

.. so I guess we should be pinning _all_ the pecl packages?

dasrecht (approved):
Builds locally and on CI, looks good to me

PR #1563, opened 2020-01-21
Harbor k8s

cdchris12:
This PR integrates Harbor into our native k8s implementation as the default repository to be used for Lagoon and for Lagoon local.

# Checklist
  - See #1680 


docker-compose.yaml
```diff
@@ -455,63 +455,204 @@ services:
       lagoon.rollout: daemonset
   harborclair:
     image: ${IMAGE_REPO:-lagoon}/harborclair
+    hostname: harborclair
```
Schnitzel, line 562: do we really need this hostname defined? in my understanding this is not needed, applies to all services

services/api/src/clients/harborClient.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+const got = require('got');
+const R = require('ramda');
+
+const { HARBOR_ADMIN_PASSWORD } = process.env;
+
+const lagoonHarborRoute = R.compose(
+  R.defaultTo('http://172.17.0.1:8084'),
```
Schnitzel, line 8: can't the default be just `harbor-nginx:8080`? We want to run as much within the docker-compose networking instead of having the api connect to an "external" port.

services/api/src/clients/harborClient.js
```diff
@@ -0,0 +1,22 @@
...
+  R.find(R.test(/harbor-nginx/)),
+  R.split(','),
+  R.propOr('', 'LAGOON_ROUTES'),
+)(process.env);
+
+// Test to see if we have a local harbor or not
+// curl 172.17.0.1:8084; if 200: use that. if not, use harbor-nginx?
```
Schnitzel: I don't see such a test?

services/api/src/resources/environment/harborSetup.js
```diff
@@ -0,0 +1,136 @@
...
+
+const lagoonHarborRoute = R.compose(
+  R.defaultTo('http://172.17.0.1:8084'),
+  R.find(R.test(/harbor-nginx/)),
+  R.split(','),
+  R.propOr('', 'LAGOON_ROUTES'),
+)(process.env);
+
+const createHarborOperations = (sqlClient /* : MariaSQL */) => ({
+  addEnvironment: async (openshiftProjectName, environmentID) => {
```
Schnitzel: we create a new harbor project for each Lagoon Environment? I don't think that's necessary? a harbor Project per Lagoon Project should be enough?

services/api/src/resources/environment/harborSetup.js
```diff
@@ -0,0 +1,136 @@
...
+
+    // Get new harbor project's id
+    try {
+      const res = await harborClient.get(`projects?name=${openshiftProjectName}`)
+      var harborProjectID = res.body[0].project_id
+      logger.debug(`Got the harbor project id for project ${openshiftProjectName} successfully!`)
+    } catch (err) {
+      // 409 means project already exists
+      // 201 means project created successfully
+      if (err.statusCode == 404) {
```
Schnitzel: and what means 404? :)

services/api/src/resources/environment/harborSetup.js
```diff
@@ -0,0 +1,136 @@
...
+      // 201 means project created successfully
+      if (err.statusCode == 409) {
+        logger.info(`Unable to create the harbor project "${openshiftProjectName}", as it already exists in harbor!`)
+      } else {
+        console.log(res)
+        logger.error(`Unable to create the harbor project "${openshiftProjectName}" !!`, err)
+      }
+    }
+
+    // Get new harbor project's id
```
Schnitzel: it gets not only new harbor project ids? but also existing ones?

services/api/src/resources/environment/harborSetup.js
```diff
@@ -0,0 +1,136 @@
...
+          "environment": environmentID,
+        }),
+      );
+      logger.debug(`Environment variable INTERNAL_REGISTRY_URL for ${openshiftProjectName} created!`)
+
+    } catch (err) {
+      logger.error("Initial var already set!!")
+    }
+
+    try {
```
Schnitzel: is there a specific reason to do this in two `try` blocks?

services/api/src/resources/environment/harborSetup.js
```diff
@@ -0,0 +1,136 @@
...
+        Sql.insertEnvVariable({
+          "name": "INTERNAL_REGISTRY_PASSWORD",
+          "value": harborTokenInfo.token,
+          "scope": "INTERNAL_CONTAINER_REGISTRY",
+          "environment": environmentID,
+        }),
+      );
+      logger.debug(`Environment variable INTERNAL_REGISTRY_PASSWORD for ${openshiftProjectName} created!`)
+    } catch (err) {
+      logger.error("Something went wrong setting env vars!", err)
```
Schnitzel: if you want to get the error, you need to output it like:
```
logger.error("Error while setting up harbor environment variables for ${openshiftProjectName}, error: ${err}")
```

services/api/src/typeDefs.js
```diff
@@ -56,6 +56,7 @@ const typeDefs = gql`
     RUNTIME
     GLOBAL
     CONTAINER_REGISTRY
+    INTERNAL_CONTIANER_REGISTRY
```
Schnitzel: `INTERNAL_CONTAINER_REGISTRY`

Schnitzel (changes requested):
looks very very nice! did a review with what is here already

docker-compose.yaml
cdchris12: From my research, this variable tells the created docker-container what its hostname should be. This is useful if a container needs to talk to itself. As far as I can tell, it shouldn't hurt anything to specify the container's hostname.

services/api/src/clients/harborClient.js
cdchris12: Because the docker-compose harbor is running in a different docker network than the k3d docker network, we have to use this forwarded port to allow access between the two networks.

services/api/src/resources/environment/harborSetup.js
cdchris12: This was resolved by switching to a single harbor project per lagoon project.

services/api/src/resources/environment/harborSetup.js
cdchris12: That is correct; we want to know if any harbor project exists for this lagoon project name

services/api/src/resources/environment/harborSetup.js
cdchris12: This should definitely only be one try/catch block; will update

services/api/src/resources/environment/harborSetup.js
cdchris12: That was a copypasta error; resolved with next push ;)

services/api/src/resources/environment/harborSetup.js
cdchris12: Updated.

services/api/src/typeDefs.js
cdchris12: Good catch!

services/api/src/clients/harborClient.js
cdchris12: This was a comment from back when I was still trying to figure out the logic for making this work. Since we've changed the way this all works since then, this is no longer relevant.

services/api/src/resources/project/harborSetup.js
```diff
@@ -0,0 +1,128 @@
...
+      logger.debug(`Got the harbor project id for project ${lagoonProjectName} successfully!`)
+    } catch (err) {
+      if (err.statusCode == 404) {
+        logger.error(`Unable to get the harbor project id of "${lagoonProjectName}", as it does not exist in harbor!`)
+      } else {
+        logger.error(`Unable to get the harbor project id of "${lagoonProjectName}" !!`)
+      }
+    }
+
+    console.log(harborProjectID)
```
rocketeerbkw: if it's useful to have this in the logs, then can we remove this which would just print a number (?) with no context and add it to the debug() on line 52?

e.g., `logger.debug(`Harbor project id for ${lagoonProjectName}: ${harborProjectId}`)`

services/api/src/resources/project/harborSetup.js
```diff
@@ -0,0 +1,128 @@
...
+            {
+              resource: `/project/${harborProjectID}/repository`,
+              action: "push"
+            }
+          ]
+        }
+      })
+      var harborTokenInfo = res.body
+      console.log(harborTokenInfo)
+      logger.debug(`Robot was created for Harbor project ${lagoonProjectName} !`)
```
rocketeerbkw: if `harborTokenInfo` is useful to have in logs can we add it to the debug line below so that the logged values have some context?

services/api/src/resources/project/harborSetup.js
```diff
@@ -0,0 +1,128 @@
...
+    } catch (err) {
+      // 409 means project already exists
+      // 201 means project created successfully
+      if (err.statusCode == 409) {
+        logger.error(`Unable to create a robot account for harbor project "${lagoonProjectName}", as a robot account of the same name already exists!`)
```
rocketeerbkw: This probably isn't an error, since an existing account is still acceptable. Maybe `warn()` or `debug()` instead?

rocketeerbkw (changes requested):

rocketeerbkw (approved):
LGTM, will merge when I'm done merging master into kubernetes

Issue #1572, opened 2020-01-22
Upgrade Elasticsearch and Kibana to 7.3.2


vincenzodnp:
Elasticsearch 7.3.2 and Kibana 7.3.2 were released.
There is also Opendistro plugin available for ES 7.3.2

smlx:
https://github.com/amazeeio/lagoon/issues/1501

I thought this sounded familiar!

PR #1580, opened 2020-01-23
1572 upgrade elasticsearch kibana opendistro

### LINKED ISSUE #1571: cluster.routing.allocation.disk.threshold_enabled default set to false
**Describe the bug**
Elasticsearch, as reported in the official documentation, has a `disk allocation decider` which controls where allocate new shards or where relocate shards from a node in case of disk usage above a threshold.
The default value in ES 7.5 is `true` but for some reason, Opendistro sets it to `false`.

```
grep cluster.routing.allocation.disk.threshold_enabled /usr/share/elasticsearch/config/elasticsearch.yml 
cluster.routing.allocation.disk.threshold_enabled: false
```

**To Reproduce**
1. Deploy Elasticsearch
2. Grep cluster.routing.allocation.disk.threshold_enabled  on the configuration file

**Expected behavior**
`cluster.routing.allocation.disk.threshold_enabled`  set to `true`

vincenzodnp:
This PR fixes an Elasticsearch issue (#1571) inherited by `opendistro` default's configuration, that diverges from the elasticsearch default configuration. 
By changing the default value of the variable (`cluster.routing.allocation.disk.threshold_enabled)`, elasticsearch will have the default behavior, of balancing shards across all the nodes based on disk allocation.

Moreover, the PR updates `Elasticsearch`, `Kibana` to `7.3.2` and `opendistro_security` plugin to `1.3.0`

# Checklist

- `cluster.routing.allocation.disk.threshold_enabled` is set to false, causing ES to do not balance shards based on disk allocation which causes some nodes to be fuller than others.  
- Upgrade Elasticsearch, Kibana, and Opendistro_security to a newer version.

# Closing issues
closes #1571 #1572

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -467,8 +467,8 @@ $(build-services-galera):
 # Dependencies of Service Images
 build/auth-server build/logs2email build/logs2slack build/logs2rocketchat build/logs2microsoftteams build/openshiftbuilddeploy build/openshiftbuilddeploymonitor build/openshiftjobs build/openshiftjobsmonitor build/openshiftmisc build/openshiftremove build/rest2tasks build/webhook-handler build/webhooks2tasks build/api build/cli build/ui: build/yarn-workspace-builder
 build/logs2logs-db: build/logstash__7
-build/logs-db: build/elasticsearch__7.1
-build/logs-db-ui: build/kibana__7.1
+build/logs-db: build/elasticsearch__7
+build/logs-db-ui: build/kibana__7
 build/logs-db-curator: build/curator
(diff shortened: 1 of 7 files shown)


Schnitzel:
there is already an elasticsearch 7.5, so we should upgrade `elasticsearch:7` to that, and maybe release an earlier version of elasticsearch if opendistro is not that far yet.

tobybellwood:
I've un-milestoned this for now - let's have a regroup around how best to publish and use the elastic images - there's a combination of lagoon usage and customer usage that we need to untangle a bit

PR #1585, opened 2020-01-24
Updating Varnish default `CACHE_SIZE` to 500M

dasrecht:
The Default memory allowance can lead to content delivery issues due to a full cache. We're bumping the cache size of varnish to 500M 

# Checklist


seanhamlin (commented):
I agree this is a sensible step in the right direction. One thing I have found that also works wonders is disabling the caching of 'large' binary files (e.g. `.docx`, `.zip` etc). Having an upper limit of say 10MB to which if you are over forces a cache miss would be far hugely beneficial for some sites.

```
  if ( beresp.http.Content-Length ~ "[0-9]{8,}" ) {
    set beresp.do_stream = true;
    return (pass);
  }
```

Unsure if you want to scope creep this PR, or whether this should be a follow up PR.

dasrecht:
Oh that's a great idea! Totally fine with having this in this PR as long as we make it distinguishable in the changelog

seanhamlin:
Ah good point. Perhaps it should be a follow up PR then, then it will get its own changelog entry.

Raised #1613 to ensure we don't lose track of this.

seanhamlin (approved):

PR #1621, opened 2020-01-30
[WIP] Prune stale environments.

### LINKED ISSUE #1555: Prune unused non-production environments (codename 'Arnie')
## Describe the feature

At the moment there exists in Lagoon an 'auto-idler' to which will spin down non-production pods when there is no traffic to the environment in a given time period (around 4 hours from memory).

This feature is taking this concept 1 step further and removes the environment, reclaiming any PVs and generally freeing up the cluster from the burden of looking after unused environments.

It is not uncommon on some projects to have extremely high non-production environment limits, up to 50 is not uncommon to see in use on a single site. Most of these branches/PR environments just lie around, doing nothing except using up auto-idler time, and PV claims.

The logic _could_ work something like this:

* Is the environment considered **non-production**?
* Has there been **no deployments in the last X days** (e.g. 90 days)
* If yes to both, then remove the environment

On an existing large cluster, I did a Kibana query, and this number represents around 15% of all projects deployed to OpenShift. This would represent a tangible increase in overall cluster resources.

## What can users do to prevent an environment being removed?

There are many ways to avoid the environment being terminated:

* Push a commit to the appropriate branch or PR
* Trigger a redeployment (e.g. through GraphQL, or the Lagoon CLI)
* Perhaps even a button in the Lagoon UI to force the above

Potentially we could also safe guard certain patterns as well, e.g. `/^dev(elop)?/` or `/test/` from the Arnie script.

## Extra for experts

Have the Arnie script configurable per cluster, even to the point of disabling it altogether if needs be. This way each customer with a dedicated cluster can set what is sensible for them. Perhaps 90 days could be the default however.

steveworley:
- Adds a new service that will make a Lagoon API request to remove the environment.

# Closing issues
Closes #1555

### DIFF
--- a/services/environment-pruner/.lagoon.yml
+++ b/services/environment-pruner/.lagoon.yml
@@ -1,0 +1,141 @@
+apiVersion: v1
+kind: Template
+metadata:
+  creationTimestamp: null
+  name: lagoon-openshift-template-cli
+parameters:
+  - name: SERVICE_NAME
+    description: Name of this service
+    required: true
+  - name: SAFE_BRANCH
+    description: Which branch this belongs to, special chars replaced with dashes
+    required: true
+  - name: SAFE_PROJECT
+    description: Which project this belongs to, special chars replaced with dashes
+    required: true
+  - name: BRANCH
+    description: Which branch this belongs to, original value
+    required: true
+  - name: PROJECT
+    description: Which project this belongs to, original value
+    required: true
+  - name: LAGOON_GIT_SHA
+    description: git hash sha of the current deployment
+    required: true
+  - name: SERVICE_ROUTER_URL
+    description: URL of the Router for this service
+    value: ""
+  - name: OPENSHIFT_PROJECT
+    description: Name of the Project that this service is in
+    required: true
+  - name: REGISTRY
+    description: Registry where Images are pushed to
+    required: true
+  - name: DEPLOYMENT_STRATEGY
+    description: Strategy of Deploymentconfig
+    value: "Rolling"
+  - name: SERVICE_IMAGE
+    description: Pullable image of service
+    required: true
+  - name: CRONJOBS
+    description: Oneliner of Cronjobs
+    value: ""
+objects:
+- apiVersion: v1
+  kind: DeploymentConfig
+  metadata:
+    creationTimestamp: null
+    labels:
+      service: ${SERVICE_NAME}
+      branch: ${SAFE_BRANCH}
+      project: ${SAFE_PROJECT}
+    name: ${SERVICE_NAME}
+  spec:
+    replicas: 1
+    selector:
+      service: ${SERVICE_NAME}
+    strategy:
+      type: ${DEPLOYMENT_STRATEGY}
+    template:
+      metadata:
+        creationTimestamp: null
+        labels:
+          service: ${SERVICE_NAME}
+          branch: ${SAFE_BRANCH}
+          project: ${SAFE_PROJECT}
+      spec:
+        volumes:
+          - name: lagoon-sshkey
+            secret:
+              defaultMode: 420
+              secretName: lagoon-sshkey
+        containers:
+        - image: ${SERVICE_IMAGE}
+          name: ${SERVICE_NAME}
+          envFrom:
+          - configMapRef:
+              name: lagoon-env
+          env:
+            ## LAGOON_GIT_SHA is injected directly and not loaded via `lagoon-env` config
+            ## This will cause the cli to redeploy on every deployment, even the files have not changed
+            - name: LAGOON_GIT_SHA
+              value: ${LAGOON_GIT_SHA}
+            - name: SERVICE_NAME
+              value: ${SERVICE_NAME}
+            - name: CRONJOBS
+              value: ${CRONJOBS}
+            - name: JWTSECRET
+              valueFrom:
+                secretKeyRef:
+                  name: jwtsecret
+                  key: JWTSECRET
+            - name: LOGSDB_ADMIN_PASSWORD
+              valueFrom:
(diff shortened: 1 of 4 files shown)


services/environment-pruner/prune.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+# set -e -o pipefail
+
+# Create a JWT admin token to talk to the API.
+API_ADMIN_JWT_TOKEN=$(./create_jwt.sh)
+BEARER="Authorization: bearer $API_ADMIN_JWT_TOKEN"
+
+CUTOFF_DATE=$(date +'%Y-%m-%d %H:%M:%S' --date '-90 days')
```
tobybellwood: should this `-90` be hardcoded, or definable in a variable?

steveworley: Ideally a variable :) the other thing I'm not 100% certain about is if just triggering a delete mutation is enough to remove the environment (it's how we've been doing it) but there might be some oc magic that is needed as well.

services/environment-pruner/prune.sh
```diff
@@ -0,0 +1,75 @@
...
+
+      echo "$OPENSHIFT_URL - $PROJECT_NAME: handling environment $ENVIRONMENT_NAME"
+
+      if [ "$ENVIRONMENT_NAME" = "$PRODUCTION_BRANCH" ]; then
+        # This should only ever remove non-prods.
+        echo "$PROJECT_NAME: $ENVIRONMENT_NAME is the production environment"
+        continue;
+      fi
+
+      Skip if no openshift project exists.
```
seanhamlin: should be a comment yeah?

services/environment-pruner/prune.sh
```diff
@@ -0,0 +1,75 @@
...
+
+    echo "$project" | jq -c '.environments[]' | while read environment
+    do
+      ENVIRONMENT_OPENSHIFT_PROJECTNAME=$(echo "$environment" | jq -r '.openshiftProjectName')
+      ENVIRONMENT_NAME=$(echo "$environment" | jq -r '.name')
+      UPDATED_DATE=$(echo "$environment" | jq -r '.updated')
+
+      echo "$OPENSHIFT_URL - $PROJECT_NAME: handling environment $ENVIRONMENT_NAME"
+
+      if [ "$ENVIRONMENT_NAME" = "$PRODUCTION_BRANCH" ]; then
```
seanhamlin: string comparison in bash is weird. a single equals always makes me twitch.

seanhamlin:
Love this @steveworley ! I salute you.

services/environment-pruner/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    && npm config set unsafe-perm true \
+    && npm -g install jwtgen
+
+COPY create_jwt.sh prune.sh /
+
+ENV JWTSECRET=super-secret-string \
+    JWTAUDIENCE=api.dev \
+    PROJECT_REGEX=".+"  \
+    ROUTER_LOG_INTERVAL=4h \
+    POD_RUN_INTERVAL=14400 \
```
Schnitzel, line 16: this environment variable is never used?

services/environment-pruner/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    ELASTICSEARCH_URL="http://logs-db-service:9200"
```
Schnitzel, line 17: is never used

services/environment-pruner/.lagoon.yml
```diff
@@ -0,0 +1,141 @@
...
+              memory: 10Mi
+    test: false
+    triggers:
+    - type: ConfigChange
+- apiVersion: batch/v1beta1
+  kind: CronJob
+  metadata:
+    name: cronjob-remote
+  spec:
+    schedule: "*/15 * * * *"
```
Schnitzel: I think this is a bit too agressive to run this every 15min?

services/environment-pruner/.lagoon.yml
```diff
@@ -0,0 +1,141 @@
...
+    suspend: true # we're running cronjobs now via the lagoon cronjob system
```
Schnitzel: this will actually never run the conjob, not sure this is what we want?

Schnitzel (changes requested):
a nice start, thanks @steveworley for this.

Just a question overall: Can we somehow make sure that this is not enabled by default? Such a service might makes sense for some usecases, but it could also lead into very very confused users of Lagoon.

Also I think we need to create a system to disable this pruning on a project/environment base, like we do for the `autoIdle`, see https://github.com/amazeeio/lagoon/search?q=autoidle&type=Code

services/environment-pruner/.lagoon.yml
seanhamlin: Agree, when looking at a timespan measured at 90 days, running it once a day seems OK.

steveworley:
Thanks! I just wanted to get a commit up and a starting point validated to make sure it was a direction everyone was happy with :)

> Also I think we need to create a system to disable this pruning on a project/environment base, like we do for the `autoIdle`, see https://github.com/amazeeio/lagoon/search?q=autoidle&type=Code

Agreed - I have some local changes that looks to update the Lagoon schema definition for projects/environments to allow pruning.

services/environment-pruner/prune.sh
steveworley: Whoops! Yeah - I went a bit crazy with my uncommenting while I was testing.

tobybellwood:
There are two angles for consideration here - the management of environments might be better managed by the Lagoon administrator, and this task shouldn't require a dedicated service. Going to close here whilst we rethink

PR #1634, opened 2020-02-05
Use DBaaS operator

shreddedbacon:
# Checklist

This PR allows for lagoon to leverage the DBaaS operator instead of using the ansible service broker (ASB). It introduces a new service type called `dbaas-shared` which can be used instead of the older `mariadb-shared` which uses the ASB.

It still allows for the ASB to be used if `mariadb-shared` is defined as the lagoon.type in the `.lagoon.yml` file.

It will prefer the `dbaas-shared` type if undefined, before falling back to `mariadb-shared` then `mariadb-single` as before.

# Closing issues
none

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -493,7 +493,9 @@ service-images += ssh
 
 # Images for local helpers that exist in another folder than the service images
 localdevimages := local-git \
-									local-api-data-watcher-pusher
+									local-api-data-watcher-pusher \
+									local-dbaas-provider
+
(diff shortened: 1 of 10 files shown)


shreddedbacon:
@tobybellwood I'm not sure if it should be labelled `2-build-deploy` or `5-operators-provisioning`, or both

Schnitzel:
superseeded by https://github.com/amazeeio/lagoon/pull/1657

PR #1641, opened 2020-02-06
Add helmchart for python

### LINKED ISSUE #1596: Image - Python2 + Python3 Helm Chart - create
Currently, there is only a `python-ckandatapusher`, this should be made more generic for python in general, the ckandatapusher image could just be deployed as `lagoon.type: python`, as it is really only changes in the image itself that are needed, not the template

twardnw:
closes #1596


tobybellwood:
- Updated helmchart to match node template
- added tests to kubernetes and control-k8s

tobybellwood:
This chart sets the port to 8800.  If we were to want it configurable - someone is welcome to make it so - but the tests currently have flask set to use 8800.

PR #1644, opened 2020-02-07
Add group env_var keycloak scopes/perms

rocketeerbkw:
Adds keycloak permission checks for group environment variables that only allow group maintainers to view/add them

### DIFF
--- a/services/api/src/resources/env-variables/resolvers.ts
+++ b/services/api/src/resources/env-variables/resolvers.ts
@@ -203,6 +203,43 @@ const addEnvVariableToEnvironment = async (
   return R.prop(0, rows);
 };
 
+// Stub/Example that implements add permission check
+const addEnvVariableToGroup = async (
+  root,
+  {
+    input: {
+      id, type, typeId, name, value, scope: unformattedScope,
+    },
+  },
+  {
+    sqlClient,
+    hasPermission,
+  },
+) => {
+  await hasPermission('env_var', 'group:add', {
+    group: `${typeId}`,
+  });
+
+  const scope = envVarScopeToString(unformattedScope);
+
+  const {
+    info: { insertId },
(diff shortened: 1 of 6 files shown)


services/api/src/resources/env-variables/resolvers.ts
```diff
@@ -38,7 +30,6 @@ const getEnvVarsByProjectId = async (
   if(args.withGroupEnvVars){
     const projectGroups = await models.GroupModel.loadGroupsByProjectId(pid);
```
rocketeerbkw, line 31: This will allow all users to see all env vars for all groups for a project, regardless of a member is part of a group or not. We either need to rewrite this section so that is uses the same logic as here https://github.com/amazeeio/lagoon/blob/master/services/api/src/resources/group/resolvers.ts#L63

OR, better IMO is we don't surface the group level env vars in the project. Instead, i've added `envVariables` to the group and they can be retrieved like

```graphql
project:projectByName(name: "credentialstest-project1"){
    id
    name
    groups {
      id
      name
      ... on Group {
        envVariables {
          id
          name
          scope
          value
        }
      }
    }
  }
```

since the groups are all added in openshiftbuilddeploy anyway, I don't think it's too much of a problem to do an extra query, or modify the current one to get them via groups

services/api/src/resources/env-variables/resolvers.ts
```diff
@@ -187,7 +203,44 @@ const addEnvVariableToEnvironment = async (
   return R.prop(0, rows);
 };
 
-const deleteEnvVariable = async (
+// Stub/Example that implements add permission check
```
rocketeerbkw, line 206: whatever mutation ends up getting created for adding group env vars, the code in here should work

twardnw (approved):

PR #1650, opened 2020-02-10
Update the shared cleanup script to support multiple ASB configs, also deal with the read replica support

seanhamlin:
Update the shared cleanup script to

* support multiple ASB configs
* deal with the read replicas

Also, there now is additional checks for whether each database is likely in use, and a SQL statement printed that you can use to drop the databases in question.

This has been tested to work on multiple RDS clusters.

# Checklist


smlx (approved):
LGTM and most importantly it's been tested against RDS already

PR #1662, opened 2020-02-14
Add group level environment variables

### LINKED ISSUE #1538: group-level environment variables
Extend the API so that environment variables can be provided at a group level

twardnw:
# Checklist

# Closing issues
Closes #1538


services/api/src/resources/env-variables/resolvers.ts
```diff
@@ -21,18 +13,35 @@ const envVarScopeToString = R.cond([
...
     sqlClient,
     hasPermission,
+    models,
   },
 ) => {
   await hasPermission('env_var', 'project:view', {
     project: pid,
   });
-
+  let queries = [];
```
rocketeerbkw: Since you added the env variables as part of the group itself, and lagoon is internally using that method, can we get rid of this change to `getEnvVarsByProjectId` all-together and only get project env vars from the project field? It seems odd, that this would work, but the `environment` doesn't have a similar `withGroupEnvVars`/`withProjectEnvVars` argument.

services/api/src/typeDefs.js
```diff
@@ -329,7 +333,7 @@ const typeDefs = gql`
     """
     Environment variables available during build-time and run-time
     """
-    envVariables: [EnvKeyValue]
+    envVariables(withGroupEnvVars: Boolean): [EnvKeyValue]
```
rocketeerbkw: IMO we should not modify this field. More explanation in previous comment

services/openshiftbuilddeploy/src/index.js
```diff
@@ -250,6 +256,13 @@ const messageConsumer = async msg => {
       buildconfig.spec.strategy.customStrategy.env.push({"name": "PROMOTION_SOURCE_ENVIRONMENT","value": promoteSourceEnvironment})
       buildconfig.spec.strategy.customStrategy.env.push({"name": "PROMOTION_SOURCE_OPENSHIFT_PROJECT","value": openshiftPromoteSourceProject})
     }
+    if (!R.isEmpty(projectOpenShift.groups)) {
+      projectOpenShift.groups.forEach(group => {
+        if (!R.isEmpty(group.envVariables)){
+          buildconfig.spec.strategy.customStrategy.env.push({"name": "LAGOON_GROUP_VARIABLES", "value": JSON.stringify(group.envVariables)})
```
rocketeerbkw, line 280: If there were two groups with env vars for a project, then wouldn't this loop produce duplicates environment names in the build config? How does openshift handle that, does the last one win or do they get merged?

rocketeerbkw (changes requested):

rocketeerbkw:
For 1.4.0, we'll merge in the `kubernetes` branch to master, then this PR will need to be updated to also inject the group environment variables into kubernetes containers

services/openshiftbuilddeploy/src/index.js
Schnitzel: yes I think the last one wins

Schnitzel:
@twardnw 
can you please check the comments by brandon and fix the merge conflict?

services/openshiftbuilddeploy/src/index.js
twardnw: correct

twardnw:
@rocketeerbkw , Michael and I looked over the failures here, it seems to be something relating to the change to typescript? Should we switch it back to standard JS to get it merged and then update later?

twardnw:
I should expand, we're specifically seeing an error when the projects are created in the API (via the `populate-api-data` files ) showing
```Error while setting up harbor environment variables for ci-test1, error: TypeError: Sql.insertEnvVariable is not a function```
but that function _does_ exist, the only difference we can see is the env-variables resolver and sql files are converted to typsecript, when nothing else in the API appears to be there yet?

Schnitzel:
@rocketeerbkw can you help us here convert this to TypeScript?

tobybellwood:
This implementation is very out of date and would require a significant rewrites if we thought it was worth implementing.

PR #1671, opened 2020-02-19
Variables from API escaped incorrectly in build-args

### LINKED ISSUE #1667: build environment variables get escaped incorrectly
**Describe the bug**
When a user is adding a build environment variable e.g. a private key starting with `-----BEGIN PUBLIC KEY----- `

Lagoon deployments fail due to issues with the escaping around the dashes

`--build-arg TEST_VARIABLE=-----BEGIN PUBLIC 'KEY-----\n`

shreddedbacon:
# Checklist

This was an interesting one. Variables defined in the API when being constructed as `build-args` using `jq` were not escaped correctly which lead to build-deploy errors.
```
BUILD_ARGS+=($(echo $LAGOON_PROJECT_VARIABLES | jq -r '.[] | select(.scope == "build" or .scope == "global") | "--build-arg \(.name)=\(.value)"'))
```

I noticed we already successfully pass in a key as a build arg, and this works fine.
```
BUILD_ARGS+=(--build-arg LAGOON_SSH_PRIVATE_KEY="${SSH_PRIVATE_KEY}")
```
But when the ones in the API are pulled out and dumped using `jq` it did some shenanigans around it, which I'm not really sure how, but made it show up like the following with the errant single quotes.
```
--build-arg MY_PUB_KEY=-----BEGIN PUBLIC 'KEY-----
```
This would cause a build to fail at the first container image build step, as the arguments passed to docker build got all sorts of upset.

But the `LAGOON_SSH_PRIVATE_KEY` would show up with the entire name and value wrapped in single quotes
```
--build-arg 'LAGOON_SSH_PRIVATE_KEY=-----BEGIN OPENSSH PRIVATE KEY-----'
```

This PR re-crafts the way the build-args are created for variables defined in the API to be more like the ones that already exist.

The arg defined in the API now shows like the following, which matches how others are presented
```
--build-arg 'MY_PUB_KEY=-----BEGIN PUBLIC KEY-----'
```

Update: PR also includes #1643 changes from OC builddeploy into Kubectl builddeploy

# Closing issues
closes #1667


Schnitzel:
nice work @shreddedbacon ! can you also update the tests that we are actually testing for multiline and spaced variables?

Schnitzel:
@shreddedbacon can you have a look at this? as k8s has landed in master there are now some updates needed.

shreddedbacon:
@Schnitzel I've added the changes to the kuberenetes builddeploy, I also incorporated the reduced verbosity fixes to bring it inline with oc builddeploy

Schnitzel:
@shreddedbacon 
looks like the k8s test fails :( probably because we only add the env variables to `local-dev/api-data/02-populate-api-data-openshift.gql` but there is also `local-dev/api-data/03-populate-api-data-kubernetes.gql` now :)

PR #1682, opened 2020-02-21
Use UTF-8 charset in api-db connection

### LINKED ISSUE #1681: Mojibake in Lagoon API JSON
**Describe the bug**
The Lagoon API sometimes returns incorrectly encoded JSON.
The specific string I noticed the issue on is this one returned from the Lagoon API (slightly edited):

moji0

**To Reproduce**
To reproduce the issue you can edit a database table directly to add a multibyte utf-8 character to a project title, and then list projects using the lagoon cli.

E.g. add "…" to a project tile:

moji1

And then list projects using the `lagoon-cli`:

moji3

**Expected behavior**
There are two problems with this. The first is the mojibake, and the second is that JSON is defined to be UTF-8 encoded, which the string shown in the first screenshot isn't.

I'd expect:

* Lagoon to handle non-ascii glyphs cleanly, and preferably _only_ UTF-8 throughout.
* the JSON library used by the Lagoon API not to encode invalid JSON.

**Screenshots**
See above.

**Additional context**
https://en.wikipedia.org/wiki/Mojibake

smlx:
# Checklist

The [default character set](https://mariadb.com/kb/en/setting-character-sets-and-collations/) used by MariaDB is latin1, a single-byte encoding scheme. Lagoon currently doesn't set the charset on the DB connection, so it uses the default latin1.

This means that if a multibyte UTF-8 character was encoded and written to the DB, then read back and decoded as latin1, it would cause mojibake.

Taking the string from #1681  and reversing the encoding/decoding shows that it has likely been mis-decoded as latin1:
moji2

This PR sets the charset of the database connection to UTF-8 to avoid this issue.

In this example I'm listing the project from #1681 using `lagoon-cli`:

Before:
moji3

After this PR:
moji4

# Closing issues
Closes #1681

### DIFF
--- a/services/api/src/clients/sqlClient.ts
+++ b/services/api/src/clients/sqlClient.ts
@@ -9,12 +9,13 @@ export const getSqlClient = () => {
     user: 'api',
     password: 'api',
     db: 'infrastructure',
+    charset: 'utf8',
   });
 
   sqlClient.on('error', error => {
     logger.error(error.message);
   });
-  
+
   return sqlClient;
 };
 


rocketeerbkw:
@smlx is there a need to correct any data that's already stored, or is it just a connection config issue?

smlx:
hi @rocketeerbkw good question

project names seem to be protected from non-ascii chars by the API:
```
lagoon -l local add project -p foo… -g git://example.com/foo.git -S 2 -E prod
Error: graphql: Only lowercase characters, numbers and dashes allowed for name!
```

however environment names are not:
```
$ lagoon -l local list environments -p foo
ID	NAME	DEPLOYTYPE	ENVIRONMENT	ROUTE 
15	foo…	branch    	development	none	
```

The issue I saw originally in #1681 was in an environment name, but it was in an old lagoon installation that was migrated from pre-RBAC. So maybe the migration script connected to the DB using utf8 to write the environment names or something..

In my testing this PR _will_ cause issues for fields that have non-ascii glyphs in them.

so you're right, this needs more consideration before merging. and I guess cleaning up existing DB entries? :thinking:

Schnitzel:
@smlx 
there are entrypoints that are running when we update the DB:
https://github.com/amazeeio/lagoon/tree/master/services/api-db/docker-entrypoint-initdb.d

we could add a check in there to see if there is any wrong characters in the DB and throw an error that it should be handled manually by the admins?

smlx:
I did a bit of reading on this, and I think there's a way to fix it cleanly going forward, without breaking existing lagoon installations/DBs.

Rather than the patch that I've got in this PR which sets the charset on the connection, the default server charset and collation should be set at the server level via:
```
SET character_set_server = 'utf8mb4';
SET collation_server = 'utf8mb4_unicode_ci';
```
This can be done in the DB update scripts that @Schnitzel linked above.

What this should do is _only_ set the new encoding for _new_ databases. Otherwise it will fallback to whatever encoding is used by the existing database. So effectively it will mean that new Lagoon installations will be `utf8`, while old ones will remain `latin1` to avoid breaking anything.

When I get a chance I'll do some testing to make sure this works as expected.

PR #1685, opened 2020-02-25
Fix #1684 by setting system:image-puller to default serviceaccount

### LINKED ISSUE #1684: Promotion deployment fails because permissions to pull source images from registry
**Describe the bug**
Promotion deployment sometimes fails because of a lack of permissions in pulling images from the registry.  
Lagoon gives the destination "lagoon-deployer" service account access to the source namespace as "system:image-puller" as this is needed for the build to have access to the docker images. But the actual pods are running with the "default" serviceaccount which does not have access to the images and therefore a possible pull fails.
This behavior happens when pods of the promoted environment are scheduled in a node different from the one the source environment is running.

**To Reproduce**
1. Promote environment to production
2. Check if pods are scheduled on a different node
3. See error: `Error response from daemon: unauthorized: authentication required `

**Expected behavior**
The promoted environment should start by using images of the source environment

vincenzodnp:
Sometimes promotion deployment fails and it's necessary to change manually the DeploymentConfig because the promoted environment has no sufficient permissions to pull source images.
If the pods of the promoted environment are scheduled on the same node of the source environment, the deployment doesn't fail. If they are scheduled on another node, the deployment fails with `Error response from daemon: unauthorized: authentication required`
The error is due to a lack of permissions because pods run with the `default` serviceaccount and it has not `system:image-puller` permission set.
This PR fixes by assigning the right permissions to the serviceaccounts. 

# Checklist

This PR fixes the promotion deployment

# Closing issues
closes #1684


services/openshiftbuilddeploy/src/index.js
```diff
@@ -388,7 +389,7 @@ const messageConsumer = async msg => {
       if (err.code == 404) {
         logger.info(`${openshiftProject}: RoleBinding ${openshiftProject}-lagoon-deployer-image-puller in ${openshiftPromoteSourceProject} does not exists, creating`)
         const promotionSourceRolebindingsPost = Promise.promisify(openshift.ns(openshiftPromoteSourceProject).rolebindings.post, { context: openshift.ns(openshiftPromoteSourceProject).rolebindings })
-        await promotionSourceRolebindingsPost({ body: {"kind":"RoleBinding","apiVersion":"v1","metadata":{"name":`${openshiftProject}-lagoon-deployer-image-puller`,"namespace":openshiftPromoteSourceProject},"roleRef":{"name":"system:image-puller"},"subjects":[{"name":"lagoon-deployer","kind":"ServiceAccount","namespace":openshiftProject}]}})
+        await promotionSourceRolebindingsPost({ body: {"kind":"RoleBinding","apiVersion":"v1","metadata":{"name":`${openshiftProject}-lagoon-deployer-image-puller`,"namespace":openshiftPromoteSourceProject},"roleRef":{"name":"system:image-puller"},"subjects":[{"name":"lagoon-deployer","kind":"ServiceAccount","namespace":openshiftProject},{"name":"default","kind":"ServiceAccount","namespace":openshiftProject}]}})
```
Schnitzel: good start! But I think we need to adapt this a bit. I see two problems:
1. you are using an existing rolebinding that is called `${openshiftProject}-lagoon-deployer-image-puller` but it includes more than what the name suggests (it includes also a role binding for the default serviceaccount)
2. this code is only executed if `${openshiftProject}-lagoon-deployer-image-puller` does not exist yet and therefore it will only be fixed for new promotion projects, even though we currently have promotion projects out there already and we want to fix them as well.

Because of this I suggest to create a new rolebinding to solve the two problems in one.

Schnitzel (changes requested):

services/openshiftbuilddeploy/src/index.js
vincenzodnp: I've created a new RoleBinding: `${openshiftProject}-lagoon-default-image-puller`
See latest commit

vincenzodnp: 746ddd5a78c1aae27f76f22d5dcab5dd2beea6be

Schnitzel (approved):

PR #1690, opened 2020-02-27
KubernetesMisc Service

### LINKED ISSUE #944: Add Kubernetes services
There is currently:
- [x] openshiftbuilddeploy
- [x] openshiftbuilddeployQueue
- [x] openshiftbuilddeploymonitor
- [x] openshiftjobs
- [x] openshiftjobsmonitor
- [x] openshiftmisc
- [x] openshiftremove

which are the main services that has Lagoon to talk to OpenShift

We need to find a way for the same to talk to Kubernetes with either:
- Make them generic for OpenShift and Kubernetes (if possible)
- Create them as well for Kuberenetes

justinlevi:
# Checklist

# Closing issues
Partially closes #944


services/kubernetesmisc/src/handlers/kubernetesBuildCancel.ts
```diff
@@ -0,0 +1,87 @@
...
+const kubernetesBuildCancel = async (data: any) => {
+  const { build: { name: buildName }, project, environment } = data;
+
+  const { project: projectInfo } = await getOpenShiftInfoForProject(project.name);
+  const { url, token } = getUrlTokenFromProjectInfo(projectInfo, project.name);
+  const config: ClientConfiguration = getConfig(url, token);
+  const client = new Client({ config });
+  const { namespace, safeProjectName } = generateSanitizedNames(project, environment, projectInfo);
+
+  const deleteResult = deleteJob(client, namespace, buildName);
```
rocketeerbkw: OpenShift builds are a full "product" and can handle edge cases like cancelling a build that's already failed or completed. Since we're using jobs, we probably should double check that the job is still active right before we cancel it.

I also wonder how we should convey the status to the API. The `kubernetesbuilddeploymonitor` will bail if the job is missing and leave the build in "running" forever. The API does have a `CANCELLED` status type which we could update here. We could also add a log message "cancelled by user/lagoon api."

Edit: I got the deploy/jobs services mixed up, se revised comment above

rocketeerbkw (changes requested):

services/kubernetesmisc/src/handlers/kubernetesBuildCancel.ts
```diff
@@ -0,0 +1,135 @@
...
+      return undefined;
+    } else {
+      logger.error(err);
+      throw new Error();
+    }
+  }
+  return result;
+}
+
+const updateLagoonTask = async (jobInfo, jobStatus, taskId, project, jobName) => {
```
rocketeerbkw: I thought I saw that you deleted this function in a commit? But either way, it's not being used at all, so it can go

services/kubernetesmisc/src/handlers/kubernetesBuildCancel.ts
```diff
@@ -0,0 +1,135 @@
...
+  const config: ClientConfiguration = getConfig(url, token);
+  const client = new Client({ config });
+  const { namespace } = generateSanitizedNames(project, environment, projectInfo);
+
+  // Check that job is still active
+  const jobInfo = await getJobInfo(client, namespace, buildName);
+  if (jobInfo) {
+    logger.error(`Job ${buildName} does not exist, bailing`);
+    await deleteJob(client, namespace, buildName);
+  }
```
rocketeerbkw: I initially thought to check if the status is "active," but since the jobs monitor will (or is supposed to) delete jobs after they're failed/complete, then I guess just checking that the job exists at all is the same as checking if it's active? Can you add a comment here to explain that? It's def a business logic condition that wouldn't be clear unless you were intimately familiar with the whole tasks/jobs system.

services/kubernetesmisc/src/handlers/kubernetesBuildCancel.ts
```diff
@@ -0,0 +1,135 @@
...
+  
+  // Update lagoon deployment to CANCELLED. 
+  await updateDeployment(parseInt(id, 10), {status: 'CANCELLED'})
```
rocketeerbkw: I think this should be in the `if (jobInfo)` condition, since we don't wan't to update the deployment status unless we actually delete the job.

rocketeerbkw (changes requested):

services/kubernetesmisc/src/handlers/kubernetesBuildCancel.ts
justinlevi: 👍

services/kubernetesmisc/src/handlers/kubernetesBuildCancel.ts
justinlevi: 👍

justinlevi:
@rocketeerbkw All Code Review issues should be resolved

rocketeerbkw (approved):

rocketeerbkw:
LGTM, I expect tests will pass again. Please wait til after I merge `master` into `kubernetes` and resolve those conflicts before merging this one in 🙏

PR #1704, opened 2020-03-05
Active/Standby Implementation

### LINKED ISSUE #909: Blue Green Deployments
After some workshops within the team and with some customer feedback we have the following proposal and are seeking feedback from the Lagoon community.

## Summary
Lagoon will provide a Blue/Green Deployment system which allows Developers to switch all custom routes from one production environment to another production environment. This allows:
1. 0-Downtime production deployments: Production environment can be deployed, post-rollout tasks ran and tested before actual customers are visiting the site.
2. Fast rollbacks: Switch the routes back to the previous production environment
3. Deployment aborts: never switch the routes to the new production environment

## Changes in Lagoon
- Allow two Environments be of type `production`
- New GraphQL mutation for setting in production environment `active` while the other one is `standby`
  - this will also switching Custom Routes from the current active to the new active environment
  - Switching the active production environment will do NOT trigger a redeploy of any environments
- New Drush Site alias with `production-active` and `production-standby` (automatically filled with the real environments)
- New `routes.production` key in `.lagoon.yml` where custom routes can be defined for the production environment (only these will be switched by the new GraphQL Mutation)

### Situation
- Two production environments already defined, in this example we call them `blue` `green`, each of them is connected to a branch, also called `blue` and `green` (same system as today, environment names = branch names)
- `.lagoon.yml` contains routes:
    ```
    routes:
      insecure: Redirect
      active:
        - nginx:
          - www.example.com
          - www.example.net
      standby:
        - nginx:
          - standby.example.com
          - standby.example.net
     ```
- `.lagoon.yml` defined post-rollout tasks:
    ```
  post-rollout:
    - run:
        name: if this is production, sync from current active production environment
        command: |
            if [[ "$LAGOON_ENVIRONMENT_TYPE" == "production" ]]; then
                drush -y sql-sync @production-active @self
                drush -y rsync @production-active:%files @self:%files
            fi
        service: cli
        shell: bash
    - run:
        name: drush cim
        command: drush -y cim
        service: cli
    - run:
        name: drush updb
        command: drush -y updb
        service: cli
    - run:
        name: drush cr
        command: drush -y cr
        service: cli
     ```
- we assume that the environment `blue` was already deployed and is set as the `active` production environment
### Deployment Workflow
1. Developer pushes code update into `green` branch
2. Lagoon deploys the `green` environment
  As `green` environment is a `LAGOON_ENVIRONMENT_TYPE=production` environment, drush will run `drush -y sql-sync @production-active @self`, while `production-active` is automatically set to the `blue enviornment`. Therefore syncing the DB and Files from `blue` to `green`
3. After deployment finished, Developer tests the green deployment with accessing it via the automatically created routes like: varnish-myproject-green.us.amazee.io, also gives stakeholders possibility to test the site on these URLS (if needed also custom routes for green could be created, like green.example.com)
4. Testing of green deployment is completed and approved
5. Developer triggers a Production Environment switch via GraphQL API or Lagon UI
6. Lagoon triggers:
6.1. Checks all defined routes with a specific annotation `routes.active` and moves them within Kubernets/OpenShift from the `blue` namespace to the `green` namespace
6.1. Checks all defined routes with a specific annotation `routes.standby` and moves them within Kubernets/OpenShift from the `green` namespace to the `blue` namespace
6.2. Lagoon switches internal environment status 
  from: `blue=active, green=standby`
  to: `green=active, blue=standby`
6.3. Lagoon informs Developer about successful switch

### Rollback Scenario
This assumes that the deployment need to be switched back.
1. Developer triggers another Production Environment switch:
2. Lagoon Triggers:
3.1. Checks all defined routes under `routes.production` and moves them within Kubernets/OpenShift from the `blue` namespace to the `green` namespace
3.2. Lagoon switches internal environment status 
  from: `green=active, blue=standby`
  to: `blue=active, green=standby`
3.3. Lagoon informs Developer about successful switch

### Next Deployment Workflow
Basically same procedure, but instead of pushing into `green` environment/branch, developer pushes into `blue` environment/branch

## Additional ideas:
- Should we prevent Lagoon from deploying a change into the current active production environment? This could force the developers to always use the blue/green deployment system, but maybe could cause some issues if they want for example push a hotfix into the current active production environment without touching the standby environment
- Technically we could have multiple standby production environments, but this maybe makes it too complex to understand? Is there a use case to have multiple standby production environments?

shreddedbacon:
# Checklist

Based on the discussion in #909 this is an initial implementation for Active/Standby (AKA Blue/Green) inside Lagoon.
It uses an operator (https://github.com/amazeeio/dioscuri) to handle the movement of routes between environments in Openshift only currently.

# Closing issues
closes #909


shreddedbacon:
Few things
- [x] Routes in the API do not currently move between environments when triggering an active/standby switch
- [x] No way to currently check if the route migration has failed, the operator needs some additional logic to handle this, working on it.

shreddedbacon:
I created some new options at the project level to store info related to active/standby

These store the current active/standby environments in lagoon, these are set manually initially, but then the mutation to switch them will update them
```
activeProductionEnvironment
standbyProductionEnvironment
```

These store the routes that are actually migrated between the environments as part of the operator. They are updated by the misc task that runs and will update the API on a successful condition of the routemigrate resource
```
activeRoutes
standbyRoutes
```

shreddedbacon:
> I created some new options at the project level to store info related to active/standby
> 
> These store the current active/standby environments in lagoon, these are set manually initially, but then the mutation to switch them will update them
> 
> ```
> activeProductionEnvironment
> standbyProductionEnvironment
> ```
> 
> These store the routes that are actually migrated between the environments as part of the operator. They are updated by the misc task that runs and will update the API on a successful condition of the routemigrate resource
> 
> ```
> activeRoutes
> standbyRoutes
> ```

Just on this, I'm not thinking it makes more sense to just have `productionEnvironment` and `standbyProductionEnvironment` rather than the three of them, then the active/standby switch just flips between these two configurations.

shreddedbacon:
I did some basic UI changes to represent the active/standby environments on the project overview, and also on the environment view page. This probably needs to be done better, but I'm not sure how best to represent these.

Project overview:
image

Environment view:
image

Tests are now in place to do a route migration between two production environments.

shreddedbacon:
In the event that a branch or environment name has conflicts with the default `production-active` or `production-standby` aliases, these are now configurable in the API.

The alias also now default to `lagoon-production` and `lagoon-standby` to make it more obvious they are lagoon environments.

services/openshiftmisc/src/handlers/routeMigration.js
```diff
@@ -0,0 +1,275 @@
...
+  // research says this is because crd is not supported to be patched
+  // `message=the body of the request was in an unknown format - accepted media types include: application/json-patch+json`
+  try {
+    const migrateRoutesDelete = promisify(
+      dioscuri.ns(openshiftProject).routemigrates(openshiftProject).delete
+    );
+    await migrateRoutesDelete({
+      body: {}
+    });
+    await new Promise(resolve => setTimeout(resolve, 10000)); // sleep a bit after deleting
```
rocketeerbkw: This makes me nervous... will the timing ever depend on how active or "busy" the k8s cluster/api is? What's the impact of waiting but the the delete isn't "done" yet?

The alternative is to periodically query the API until the crd is actually deleted. An example of that is in the `openshiftremove` service https://github.com/amazeeio/lagoon/blob/master/services/openshiftremove/src/index.js#L228-L236

services/openshiftmisc/src/handlers/routeMigration.js
```diff
@@ -0,0 +1,275 @@
...
+      await migrateRoutesPost({
+        body: migrateRoutes(openshiftProject, destinationOpenshiftProject)
+      });
+      logger.verbose(`${openshiftProject}: created routeMigration resource`);
+    } catch (err) {
+        logger.error(err);
+        throw new Error();
+    }
+  } catch (err) {
+    await new Promise(resolve => setTimeout(resolve, 1000)); // sleep a bit before creating
```
rocketeerbkw: is there any way to avoid this magic sleep number?

services/openshiftmisc/src/handlers/routeMigration.js
```diff
@@ -0,0 +1,275 @@
...
+    '',
+    'task:misc-openshift:route:migrate',
+    data,
+    `*[${projectName}]* Route Migration between environments *${destinationOpenshiftProject}* started`
+  );
+  // check the route migrate resource for the status conditions, only update lagoon on a completed task
+  var whileCount = 0;
+  var breakLoop = false;
+  try {
+    while (whileCount < 10 && !breakLoop) {
```
rocketeerbkw: This while loop will totally block the node event loop, so the other tasks that this service does will have to wait for the route migration to finish.

Since we are just updating the API at this point, we could move all this code into it's own function, and call it over and over again in a `setTimeout` as needed until it's finished.

rocketeerbkw (changes requested):
Sorry I found a few things to complain about :)

Also a philosophical question: Why do we add the tracking task to the current/active environment? The process is the standby env is the "newest" stuff and we are switching it to "active." They byproduct of that is that the current "active" becomes the new "standby," but that's not the goal. The goal is to make standby become active. To me it would seem like the task is "make the standby env active" and so it would make sense that the task would show up on the standby env (which at the end of the process is now the active env).

Makefile
```diff
@@ -566,7 +567,7 @@ test-services-webhooks: test-services-main
 	IMAGE_REPO=$(CI_BUILD_TAG) docker-compose -p $(CI_BUILD_TAG) up -d webhook-handler webhooks2tasks
 
 # All Tests that use REST endpoints
-rest-tests = rest node features nginx elasticsearch
+rest-tests = rest node features nginx elasticsearch active-standby
```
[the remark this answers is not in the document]
shreddedbacon: I wasn't 100% sure so added it to be safe incase there was some edge in the makefile that I missed. Happy to remove if this entire bit is going away/unsued in future.

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -869,6 +969,11 @@ DELIMITER ;
 
 CALL add_availability_to_project();
 CALL add_production_environment_to_project();
+CALL add_standby_production_environment_to_project();
+CALL add_standby_routes_to_project();
+CALL add_production_routes_to_project();
+CALL add_standby_alias_to_project();
+CALL add_production_alias_to_project();
```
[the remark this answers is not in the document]
shreddedbacon: Can do. Cheers 😄

node-packages/commons/src/tasks.js
```diff
@@ -211,10 +211,41 @@ async function createDeployTask(deployData: Object) {
...
+      var productionEnvironmentsLimit = 2
+
+      // we want to make sure we can deploy the `production` env, and also any envs in the active/standby environment
+      // we may not even need the active/standby and just go to having `production` and `standbyProduction` only
+      if (
+        environments.project.productionEnvironment === branchName
+        || environments.project.standbyProductionEnvironment === branchName
+      ) {
+        // get a list of production environments
+        console.log(environments.project);
```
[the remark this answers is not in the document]
shreddedbacon: Is debug code, leftover from the original snippet that I got from the devenv code here https://github.com/amazeeio/lagoon/blob/master/node-packages/commons/src/tasks.js#L220

services/api/src/resources/deployment/resolvers.js
```diff
@@ -758,6 +763,99 @@ const deployEnvironmentPromote = async (
...
+
+  // active/standby really only should be between production environments
+  await hasPermission('environment', `deploy:production`, {
+    project: destProject.id,
+  });
+
+  // @TODO: if we have permission to deploy production, is this required?
+  // await hasPermission('environment', 'view', {
+  //   project: destProject,
+  // });
```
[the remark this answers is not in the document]
shreddedbacon: Thanks, will adjust this.

services/openshiftmisc/src/handlers/routeMigration.js
shreddedbacon: Yeah, this was a bit I was completely unsure on. I'll scope out the referenced code and adjust.

services/openshiftmisc/src/handlers/routeMigration.js
shreddedbacon: Sounds good, is there an example of this already in the code base that I can use as a base?

node-packages/commons/src/tasks.js
```diff
@@ -211,10 +211,41 @@ async function createDeployTask(deployData: Object) {
...
+          `projectName: ${projectName}, branchName: ${branchName}, existing environments are `,
+          prod_environments,
         );
+
+        if (prod_environments.length >= productionEnvironmentsLimit) {
+          if (prod_environments.find(i => i === branchName)) {
+            logger.debug(
+              `projectName: ${projectName}, branchName: ${branchName}, environment already exists, no environment limits considered`,
+            );
+          } else {
```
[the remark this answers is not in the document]
shreddedbacon: This was when I was testing productionEnvironment to be able to specify branches as regex in the same way the branches field can. I'll clean this up.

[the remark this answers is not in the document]
shreddedbacon: I've just looked over this again as this is meant to be similar enforcement to the development environment limit.

Part of A/S is to enforce a strict 2 production envs only. As currently, you can sort of trick lagoon into being able to deploy multiple production environments.

If I define `productionEnvironment` to be `brancha` and `standbyProductionEnvironment` to be `branchb`, and deploy both of these environments, they both will get created as production environments.

Both of these as part of `openshiftbuilddeploy` get set as `environmentType = 'production'`
https://github.com/amazeeio/lagoon/blob/6da12dfe7d0ae3d00d028cbf3011a46e7adb753d/services/openshiftbuilddeploy/src/index.js#L62-L68

If I then change `standbyProductionEnvironment` to be `branchc` then try and deploy it, it should not build it because brancha and branchb are already running as production `environmentType` in lagoon..

This should be the same as if I tried to deploy a development environment in that if I've deployed 5/5 already, then try to deploy another it will not do it.

I did test this locally to make sure that it works, and it does as it is intended to do. I can throw a test together in ansible and load it as part of this PR.

services/openshiftmisc/src/handlers/routeMigration.js
shreddedbacon: Have updated the PR with code that replaces this while loop

services/openshiftmisc/src/handlers/routeMigration.js
shreddedbacon: Yeah, removed now

shreddedbacon:
@rocketeerbkw I've made adjustments to the code based on your review. If you get a chance, can you have a check over the changes.

node-packages/commons/src/tasks.js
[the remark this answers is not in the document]
rocketeerbkw: I think there's an edge case when an environment is marked as production but its name is not `productionEnvironment` or `standbyProductionEnvironment`. If that were the case then this new code branch gets skipped and it starts processing as if it were a dev environment instead. Who know what actually happens then.

I do agree though, that the only way to hit this edge case is for someone to manually change the environment type using the API. If you've tested all this an it's working then great, I'll leave final decision to you.

rocketeerbkw (approved):
LGTM. If this goes into 1.4.0, it will need to be updated to support kubernetes also.

services/api/src/resources/deployment/resolvers.js
```diff
@@ -719,7 +724,7 @@ const deployEnvironmentPromote = async (
   const sourceEnvironment = R.prop(0, activeEnvironments);
 
   await hasPermission('environment', 'view', {
-    project: sourceEnvironment.project,
+    project: sourceEnvironment.project.id,
```
Schnitzel:
```suggestion
    project: sourceEnvironment.project
```
a bit too over eager replacing :)

Schnitzel (changes requested):

services/api/src/resources/deployment/resolvers.js
shreddedbacon: D'oh! Thanks!

PR #1731, opened 2020-03-20
Added support for Solr7.7 and latest Search API Solr module

### LINKED ISSUE #1557: Update Solr images (inc Solr 7.7 support)
Latest available solr image is 7.5 on https://hub.docker.com/r/amazeeio/solr/ - however <7.7 are EOL according to https://lucene.apache.org/solr/downloads.html

vincenzodnp:
Added support for Solr 7.7 and updated `solr-drupal` images with the latest configuration files from Drupal Solr Search API module

# Checklist

Lagoon now support Solr 7.7 and `solr-drupal` image ships configuration files from latest stable Drupal Solr Search API module

# Changelog Entry
- Added support for Solr 7.7
- solr-drupal image ships Drupal Solr Search API 8.x-3.9 configuration files

# Closing issues
closes #1557


vincenzodnp:
@Schnitzel @dasrecht can you review it?

Schnitzel:
@vincenzodnp 
- how is this image tested? I don't see a test that would use this?
- Should we remove solr7.5? Or do we want our customers still allow to use solr7.5 if solr7.7 is out?
- if we still keep solr7.5 shouldn't we update the drupal config of that one as well?

vincenzodnp:
Hello @Schnitzel 

> * how is this image tested? I don't see a test that would use this?

It passed CI checks not tests

> * Should we remove solr7.5? Or do we want our customers still allow to use solr7.5 if solr7.7 is out?

I've discussed with @dasrecht and we stated to keep 7.5 and start to ask customers to switch to 7.7 since 7.5 is EOL.

> * if we still keep solr7.5 shouldn't we update the drupal config of that one as well?

You're right. Going to update it

Schnitzel:
> we stated to keep 7.5 and start to ask customers to switch to 7.7 since 7.5 is EOL.

okay, but removing the docker image here will not remove it from dockerhub, so if we want to tell customers to not use it anymore we should better remove it here, as continue to have it in Lagoon means that we are continue to support it, which it seems to be not the case.

> It passed CI checks not tests

Not sure what exactly you mean.
My question was referring to the fact that we are not testing this solr7.7 image anywhere during the automated test system. So I would suggest to use this here:
https://github.com/amazeeio/lagoon/blob/396680e1814aa04060d747defdf37488eb6a1dc2/tests/files/drupal8-dockerfiles/php7.4/solr.dockerfile#L3

Schnitzel (changes requested):
see comment above

vincenzodnp:
@Schnitzel I've removed support for Solr7.5
Updated test to include Solr7.7
Updated documentation for Solr 7.7

PR #1737, opened 2020-03-24
Check for a mariadb-shared service instance in the namespace

### LINKED ISSUE #1735: Handle parallel usage of mariadb-dbaas and mariadb-shared
I'm not 100% sure but while looking at the mariadb-dbaas implementation in k8s I think I found an issue int he openshift implementation of mariadb-dbaas.
Basically I think in here: https://github.com/amazeeio/lagoon/blob/1d07b9505b3164564b713513f01f68e9f1e3bc02/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L88-L103

there is a missing check if the project already uses `mariadb-shared` before we check if `mariadb-dbaas` is even existing or not. As it could be that:
1. an environment first has used the `mariadb-shared` because only `mariadb-shared` servicebroker existed on the cluster
2. the cluster admin enables the `mariadb-dbaas` operator
3. on a new deployment the above check sees the existing `mariadb-dbaas` operator and now creates this to create a new `mariadb-dbaas` even though there is an already existing `mariadb-shared` 

Unless the `mariadb-dbaas` and `mariadb-shared` create the exact same DB, Username, password combo?

shreddedbacon:
# Checklist

As #1735 describes, we should check if a `mariadb-shared` using the ASB is already in the project

# Closing issues
closes #1735


images/oc-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -90,8 +90,11 @@ do
     # mariadb-single deployed (probably from the past where there was no mariadb-shared yet) and use that one
     if oc --insecure-skip-tls-verify -n ${OPENSHIFT_PROJECT} get service "$SERVICE_NAME" &> /dev/null; then
       SERVICE_TYPE="mariadb-single"
+    # check if an existing mariadb service instance already exists
+    elif oc -n ipsum-master get serviceinstance "$SERVICE_NAME" &> /dev/null; then
```
Schnitzel:
```suggestion
    elif oc -insecure-skip-tls-verify -n ${OPENSHIFT_PROJECT} get serviceinstance "$SERVICE_NAME" &> /dev/null; then
```

Schnitzel (changes requested):

PR #1738, opened 2020-03-24
Fix issue with elasticsearch buffers overflowing in logs-forwarder

smlx:
The main change that seems to increase performance is switching from the deprecated secure_forward plugin to the built-in forward plugin, which supports TLS.

The advantages of the forward plugin are:

* it's maintained.
* support for fluentd workers so that input performance can be scaled easily.
* creates a new connection per chunk forward event.

On the last point: the old secure_forward plugin held onto its connection which meant that sessions were "sticky" to a pod and effectively all work was being sent to single forwarder pod.  With the forward plugin, the load balancer can do its job of spreading work across all logs-fowarder pods.

The other changes include:

* flush threads are now set to 2 (x4 workers = 8)
* giving the output an @id so it is identifiable in logs
* not using a timekey, so chunk flushing is _only_ size dependent
* storing any tag/time in the same chunk to avoid lots of little chunks
* increasing the flush interval and chunk size
* dropping chunks on buffer overflow
* remove redundant index_name field in elasticsearch output
* make scheme configurable via environment variable (default http)

# Checklist

`logs-forwarder` was not able to write to Elasticsearch fast enough to keep up with the volume of logs. This change seeks to improve the situation.

# Closing issues

n/a

### DIFF
--- a/services/logs-forwarder/.lagoon.multi.yml
+++ b/services/logs-forwarder/.lagoon.multi.yml
@@ -269,9 +269,9 @@ objects:
           <buffer>
             @type file
             path /fluentd/buffer/elasticsearch
-            # buffer params
+            # buffer params (per worker)
             chunk_limit_size 32MB
-            total_limit_size 15GB
+            total_limit_size 5GB
             # flush params
             flush_thread_count 2
             overflow_action drop_oldest_chunk
--- a/services/logs-forwarder/.lagoon.multi.yml
+++ b/services/logs-forwarder/.lagoon.multi.yml
@@ -156,19 +156,25 @@ objects:
   data:
     FLUENT_CONF: |-
       <system>
-        log_level warn
+        log_level info
+        workers 4
       </system>
 
       <source>
-        @type secure_forward
+        @type forward
+        @id forward_in
         @label @LOGS
-        self_hostname "#{ENV['HOSTNAME']}"
-        secure true
+        tag forward
         port 24284
-        shared_key "#{ENV['LOGS_FORWARDER_SHARED_KEY']}"
-        ca_cert_path /fluentd/ssl/ca_cert.pem
-        ca_private_key_path /fluentd/ssl/ca_key.pem
-        ca_private_key_passphrase "#{ENV['LOGS_FORWARDER_PRIVATE_KEY_PASSPHRASE']}"
+        <security>
+          self_hostname "#{ENV['HOSTNAME']}"
+          shared_key "#{ENV['LOGS_FORWARDER_SHARED_KEY']}"
+        </security>
+        <transport tls>
+          ca_cert_path /fluentd/ssl/ca_cert.pem
+          ca_private_key_path /fluentd/ssl/ca_key.pem
+          ca_private_key_passphrase "#{ENV['LOGS_FORWARDER_PRIVATE_KEY_PASSPHRASE']}"
+        </transport>
       </source>
 
       <source>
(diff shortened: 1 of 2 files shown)


services/logs-forwarder/.lagoon.multi.yml
```diff
@@ -246,10 +246,12 @@ objects:
       <label @LAGOON_ELASTICSEARCH>
         <match **>
           @type elasticsearch
+          @id out_elasticsearch
           #with_transporter_log true
           #@log_level debug
           host "#{ENV['ELASTICSEARCH_HOST']}"
           port "#{ENV['ELASTICSEARCH_HOST_PORT']}"
+          scheme https
```
Schnitzel: we need to make this configurable via an ENV variable (while the default is `http`).
If ES is running within the same OpenShift as the logs-forwarder we're connecting directly to the ES service which is running http. Only if we have an external EKS (like in CH1 now) we use https

services/logs-forwarder/.lagoon.multi.yml
```diff
@@ -259,19 +261,17 @@ objects:
...
             total_limit_size 15GB
-            flush_thread_count 8
-            overflow_action block
-            retry_type periodic
-            retry_wait 10s
+            # flush params
+            flush_mode interval
+            flush_interval 60s
+            flush_thread_count 16
+            overflow_action drop_oldest_chunk
```
Schnitzel, line 277: my understanding of having `overflow_action block` is described here: https://docs.fluentd.org/configuration/buffer-section

> block processing of input plugin to emit events into that buffer

basically it means that fluentd tells the inputs (which are fluentds again in logs-collector and external logs-forwarder) to not send data anymore, so it will will the buffers of these fluentds as well.
I think that's a bit safer than just dropping the oldest chunks?

services/logs-forwarder/.lagoon.multi.yml
```diff
@@ -259,19 +261,17 @@ objects:
           ssl_version TLSv1_2
           request_timeout 60s
           slow_flush_log_threshold 90s
-          <buffer tag,time>
+          <buffer>
```
Schnitzel, line 269: mhh interesting, my understanding of https://docs.fluentd.org/output/elasticsearch#index_name-optional is that you need to setup the buffer with time if you use time in the indexname.
but maybe because we use the `record_modifier` on top already, this is not necessary?

Schnitzel (changes requested):

services/logs-forwarder/.lagoon.multi.yml
smlx: Yes that is my understanding. We don't use `index_name`, we just use `target_index_key`.

The filter that adds `index_name` is also unconditional so we don't need the `index_name` fallback.

services/logs-forwarder/.lagoon.multi.yml
smlx: Well I guess this a philosophical debate :smile: 

I think that the chunks should be dropped at the point that the bottleneck is occurring because that makes it easy to identify the point in the log pipeline that the issue is located: it's the only fluentd with a full buffer and a log full of dropped chunk warnings.

If you `block` instead, the entire pipeline backs up which tends to obfuscate the location of the issue because buffers are full everywhere and you may see issues at points in the pipeline several hops removed from the actual bottleneck. You will still end up dropping logs eventually - it just take slightly longer and will be further back in the pipeline.

services/logs-forwarder/.lagoon.multi.yml
smlx: ok, done

PR #1740, opened 2020-03-25
Deleting environment variable permission issue - Hot Fix

### LINKED ISSUE #1472: Deleting environment variable permission issue
**Describe the bug**
As a maintainer/owner in a group, I am allowed to add environment variables to a specific environment, but when I try to delete this variable it fails with `"Unauthorized: You don't have permission to \"delete\" on \"env_var\"."`

**To Reproduce**
1. Add variable 
```
mutation {
  addEnvVariable(input: {
    type: ENVIRONMENT, 
    typeId: 463, 
    scope: RUNTIME, 
    name: "MYVARIABLENAME", 
    value: "MyVariableValue"}
  ) {
    id
  }
} ## return ID:123
```

2. Note successful variable creation

3. Delete variable 
```
mutation {
  deleteEnvVariable(input: {id: 123})
}
```

4. Get permission denied error
```
"Unauthorized: You don't have permission to \"delete\" on \"env_var\"."
```

**Expected behavior**
I should be able to delete variables I created

justinlevi:
# Checklist

The issue stemmed from trying to call the `hasPermission` function without a project id. The project ID was not being set when the environment variable was being added and the query to get the `perms` requires a project ID saved in the database otherwise it will return `NULL`

```
select `project`.`id` as `pid` from `env_vars` left join `environment` on `env_vars`.`environment` = `environment`.`id` left join `project` on `env_vars`.`project` =  `project`.`id` where `env_vars`.`id` = 8
```

https://github.com/amazeeio/lagoon/blob/205acc7c97f89f3c6955e7abfe5fa1c5fae85a55/services/api/src/resources/env-variables/resolvers.js#L175

Tested this by adding/deleting an environment variable using a JWT from a user with the same roles as the issue describes. Everything worked as expected. 

# Closing issues
closes #1740

### DIFF
--- a/services/api/src/resources/env-variables/resolvers.js
+++ b/services/api/src/resources/env-variables/resolvers.js
@@ -156,6 +156,7 @@ const addEnvVariableToEnvironment = async (
       value,
       scope,
       environment: typeId,
+      project: environment.project,
     }),
   );
 
--- a/services/api/src/resources/env-variables/resolvers.js
+++ b/services/api/src/resources/env-variables/resolvers.js
@@ -156,7 +156,6 @@ const addEnvVariableToEnvironment = async (
       value,
       scope,
       environment: typeId,
-      project: environment.project,
     }),
   );
 
--- a/services/api/src/resources/env-variables/sql.js
+++ b/services/api/src/resources/env-variables/sql.js
@@ -50,7 +50,11 @@ const Sql /* : SqlObj */ = {
     knex('env_vars')
       .select({ pid: 'project.id' })
       .leftJoin('environment', 'env_vars.environment', '=', 'environment.id')
-      .leftJoin('project', 'environment.project', '=', 'project.id')
+      .leftJoin('project', function () {
+        this
+          .on('env_vars.project', '=', 'project.id')
+          .orOn('environment.project', '=', 'project.id')
+      })
       .where('env_vars.id', id)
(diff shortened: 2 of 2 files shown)


rocketeerbkw (changes requested):
I think you're on the right track for what the problem is, but this solution will have negative side effects. All environment variables are stored in a single table (`env_vars`) and we allow environment variables to be set at different "levels": `project`,` environment` and soon [`groups`](https://github.com/amazeeio/lagoon/pull/1662). The way the "level" is stored is by having a diff column in the db table for each level. So if you add an ID to the project column, you're telling lagoon to apply that environment variable for every environment in that project and that's not what should happen in the `addEnvVariableToEnvironment` resolver.

You're probably correct that the permissions check in `deleteEnvVariable` is wrong but I'd guess either the SQL query is totally wrong ([`selectPermsForEnvVariable`](https://github.com/amazeeio/lagoon/blob/master/services/api/src/resources/env-variables/sql.js#L49)), or the return result of that query is not what the code expects (maybe it's returning multiple rows and we [only look at the first one](https://github.com/amazeeio/lagoon/blob/master/services/api/src/resources/env-variables/resolvers.js#L178)?)

justinlevi:
rewrite to use conditional join

rocketeerbkw (approved):

PR #1753, opened 2020-03-31
Add webhooks for all harbor projects when created

cdchris12:
# Checklist

This commit adds the ability for kubernetesbuilddeploy to automatically add a webhook for each project when that project is created. These webhooks will be triggered when a container security scan within Harbor either fails or completes.


services/api/src/resources/project/harborSetup.js
```diff
@@ -121,6 +121,34 @@ const createHarborOperations = (sqlClient /* : MariaSQL */) => ({
     } catch (err) {
       logger.error(`Error while setting up harbor environment variables for ${lagoonProjectName}, error: ${err}`)
     }
+
+    // Set webhooks for Harbor Project
+    if (lagoonHarborRoute === 'http://172.17.0.1:8084'){
+      var webhookAddress = 'http://172.17.0.1:7777'
+    } else {
+      var webhookAddress = "https://hooks.lagoon.amazeeio.cloud/"
+    }
```
Schnitzel: I think this could be done a bit better:
1. instead of using `http://172.17.0.1:7777` we can use `http://webhook-handler:3000` as harbor and the webhook-handler are locally in the same docker network you can talk to the services with their DNS names and their internal ports.
2. The discovery of the public Hook WebHook URL can be fully automated via checking the env variables, see an example:
https://github.com/amazeeio/lagoon/blob/b6117c46d9fbc335b0dd0126a633904a70548486/services/api/src/clients/keycloakClient.js#L21-L26

services/api/src/resources/project/harborSetup.js
```diff
@@ -121,6 +121,34 @@ const createHarborOperations = (sqlClient /* : MariaSQL */) => ({
...
+    try {
+      var res = await harborClient.post(`projects/${harborProjectID}/webhook/policies`, {
+        body: {
+          targets: [
+            {
+              type: "http",
```
Schnitzel, line 138: does this work as `http` if the URL is actually an `https` link?

Schnitzel (changes requested):

services/api/src/resources/project/harborSetup.js
cdchris12: Yes. When created by the UI, this is always set as `http`, regardless if the route is using `http` or `https`.

services/api/src/resources/project/harborSetup.js
cdchris12: I agree with you on how to set the default for this variable. About the container address; we already use the 172.17.0.1 notation on line 7, which why I opted for it here. I'll test with `webhook-handler` to verify it works as expected.

cdchris12: The `webhook-handler:3000` works as expected. The only problem with that is that we aren't loading the webhook-handler for most of the CI tests. Should I write some testing logic which includes these webhooks as well?

cdchris12: After sleeping on it, I realized that we can't really test this just yet because we do not yet have a webhook handler for these Harbor webhooks as of yet.

PR #1755, opened 2020-03-31
Updated docs for lagoonizing a repo with a patches directory.

thom8:
Current documentation refers to adding a `patches` directory if using `drupal8-composer-mariadb`. As this has now been disabled by default we can remove this from the documentation steps.

https://github.com/amazeeio/lagoon/blob/master/docs/using_lagoon/drupal/drupal8-composer-mariadb/cli.dockerfile#L7

Documentation is provided in the `cli.dockerfile` to enable the use of the `patches` directory.


AlannaBurke (approved):
This is fine, but I'm wondering if we point people to the dockerfiles in our documentation - I know we have links to them, but I don't feel like we really emphasize "hey, these have info in them that you should be reading." What do you think?

Schnitzel:
@AlannaBurke yea, we could definitely do a better job in pointing people to the right places, wanna add these pointers in here? or do you want to create a separate issue?

AlannaBurke:
@Schnitzel let's make a separate issue

thom8:
This is a bit of an edge case IMO we should encourage Drupal users to post patches to D.O and reference from there.

@AlannaBurke we could create a separate page for using local patches with lagoon but I don't think it needs to be in the primary docs for lagooni(s|z)ing a project since it isn't that common.

AlannaBurke:
@thom8 I totally agree - I don't think we need a page about this particular thing, I just realized that we have instances like this where we essentially have documentation in the dockerfiles, and we aren't pointing people to it, so i want to go and take a look at that and make sure that if there is info that we need to be bringing forward into the documentation, that we're doing that!

PR #1773, opened 2020-04-06
Expose Lagoon tag/version in UI and API

### LINKED ISSUE #1384: Lagoon UI: Display the currently running tag
**Describe the feature**

It would be handy to know what version of Lagoon is running in the UI. Currently the only way to know what version you are running is buried deep inside OpenShift.

Perhaps link to the release notes as well for the released tag for bonus points.

Perhaps something like

image

shreddedbacon:
# Checklist

Being able to know which version of Lagoon is currently running via the UI or API is handy when knowing which mutations/queries are supported, or quickly checking which version is running to know what feature set is available.

This PR adds the following query, which just exposes the `LAGOON_VERSION` environment variable.
```
query lagoonVersion {
  lagoonVersion
}
```
With the response:
```
{
  "data": {
    "lagoonVersion": "v1.4.0"
  }
}
```

And adds the footer to pages like so
image

# Closing issues
closes #1384

### DIFF
--- a/services/api/src/resolvers.js
+++ b/services/api/src/resolvers.js
@@ -4,6 +4,10 @@ const GraphQLDate = require('graphql-iso-date');
 const GraphQLJSON = require('graphql-type-json');
 
 const {
+  getLagoonVersion,
+} = require('./resources/lagoon/resolvers');
+
+const {
   getDeploymentsByEnvironmentId,
   getDeploymentByRemoteId,
   addDeployment,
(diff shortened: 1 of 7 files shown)


shreddedbacon:
The only thing with this PR is that it relies on `LAGOON_VERSION` being defined appropriately in the deployment for the `api` or `ui` services in openshift/kubernetes, rather than defaulting to what is in the image itself.

Case is if a hot-fix image rolls out for the `api` or `ui` services, this variable might have a different value in the image if it isn't configured in the deployment.

Issue #1289, opened 2019-10-10
Add the ability to provide metadata to Lagoon projects/environments


tobybellwood:
Currently env variables can be provided to projects or environments - but these can be complex and unwieldy for some purposes.

There could also exist the ability to provide (and maintain) a list of more lightweight tags/topics for projects - potentially similar in scope to GitLab's tag_list implementation (https://docs.gitlab.com/12.3/ee/api/projects.html) or GitHub's topics (https://developer.github.com/v4/object/topic/)

These tags could be implemented as an array - allowing specific tags to be added or removed (and updated), exposed via the GraphQL API and filterable.

For the case of projects synced from GitLab or GitHub - these could be synced from the repos across to the project.  Whether environments could hold different tags to projects is probably a further conversation.

seanhamlin:
This would be useful for a number of usecases that I can think of:

* Knowing whether a project is considered `Real` (and not a `Test` project for instance)
* Knowing what technology a project uses (e.g. `Drupal7`, `Drupal8`, `Laravel`, `NodeJS` or `Static`)
* Knowing whether a project is considered `High profile`
* Knowing is a project is `SaaS`, `SaaS+` or `PaaS` (GovCMS specific)
* Knowing is a site is impacted by `Covid-19` and will receive a lot more traffic then usual.

Having the ability to integrate this into the GraphQL API would be great. Agree with @tobybellwood that being able to synchronise the Gitlab tags would also be marvellous.

Current API feature list

* Add a tag to a namespace
* Remove a tag from a namespace
* Get a list of tags for a namespace
* Get a list of namespaces that are tagged with a tag
* Get a list of namespaces that are not tagged with a tag

The Lagoon UI for now could just display the tag (read only) to make this a bit easier.

Bonus points would allow you to filter the project list by tag.

tobybellwood:
added in #1775

PR #1775, opened 2020-04-06
Added support to arbitrary project metadata key/value pairs

stooit:
# Checklist

At present categorising projects within Lagoon is unwieldy (env vars) and require external systems to collate and manage queries (e.g there is no way to directly add or query by arbitrary metadata).

This PR adds support for adding any number of key/value metadata pairs and querying where any project contains a specific key or key/value pair.

Sponsored by [GovCMS](https://www.govcms.gov.au/) and [Salsa Digital](https://www.salsadigital.com.au/)

From the docs added:
## Maintaining project metadata
Project metadata can be assigned using arbitrary key/value pairs. Projects can then be queried by the associated metadata; for example you may categorise projects by type of software, version number, or any other categorisation you may wish to query on later.
### Add/update metadata on a project
Updates to metadata expect a key/value pair. It operates as an `UPSERT`, meaning if a key already exists the value will be updated, otherwise inserted.
You may have any number of k/v pairs stored against a project.
```graphql
mutation {
  updateProjectMetadata(
    input: { id: 1,  patch: { key: "type", value: "saas" } }
  ) {
    id
    metadata
  }
}
```
### Query for projects by metadata
Queries may be by `key` only (e.g return all projects where a specific key exists) or both `key` and `value` where both key and value must match.
All projects that have the `version` tag:
```graphql
query projectsByMetadata {
  projectsByMetadata(metadata: [{key: "version"] ) {
    id
    name
  }
}
```
All projects that have the `version` tag, specifically version `8`:
```graphql
query projectsByMetadata {
  projectsByMetadata(metadata: [{key: "version", value: "8"] ) {
    id
    name
  }
}
```
### Removing metadata on a project
Metadata can be removed on a per-key basis. Other metadata key/value pairs will persist.
```graphql
mutation {
  removeProjectMetadataByKey (
    input: { id: 1,  key: "version" }
  ) {
    id
    metadata
  }
}
```

# Closing issues
Partial #1289, no UI


services/api/src/resources/project/resolvers.js
```diff
@@ -193,6 +193,68 @@ const getProjectByName = async (
...
+) => {
+
+  let where;
+  try {
+    await hasPermission('project', 'viewAll');
+
+    where = whereAnd([
+      args.createdAfter ? 'created >= :created_after' : '',
+      args.gitUrl ? 'git_url = :git_url' : '',
+    ]);
```
Schnitzel: you don't accept `createdAfter` and `gitUrl` as parameters for the `projectsByMetadata` Query in GraphQL, so we don't need to check for them here.

services/api/src/resources/project/resolvers.js
```diff
@@ -193,6 +193,68 @@ const getProjectByName = async (
...
+      return [];
+    }
+
+    const userProjectIds = await models.UserModel.getAllProjectsIdsForUser({
+      id: keycloakGrant.access_token.content.sub,
+    });
+
+    where = whereAnd([
+      args.createdAfter ? 'created >= :created_after' : '',
+      args.gitUrl ? 'git_url = :git_url' : '',
```
Schnitzel: not needed

services/api/src/resources/project/resolvers.js
```diff
@@ -193,6 +193,68 @@ const getProjectByName = async (
...
+      inClause('id', userProjectIds),
+    ]);
+  }
+
+  for (const meta of args['metadata']) {
+
+    let key = meta['key'];
+    let val = meta['value'];
+
+    var q = `JSON_CONTAINS(metadata, '"`+val+`"', '$.`+key+`')`;
```
Schnitzel, line 229: wow, never heard of `JSON_CONTAINS` in MySQL? Crazy world :D

services/api/src/typeDefs.js
```diff
@@ -340,6 +341,10 @@ const typeDefs = gql`
     Which groups are directly linked to project
     """
     groups: [GroupInterface]
+    """
+    Metadata key/values stored against a project
+    """
+    metadata: String
```
Schnitzel: there is a JSON field in GraphQL, see example:
https://github.com/amazeeio/lagoon/blob/8d95f023c12139b3672ce38582f5b0ebe9481cf2/services/api/src/typeDefs.js#L614
I think we can use that one?

services/api/src/util/db.js
```diff
@@ -4,6 +4,8 @@ const camelcaseKeys = require('camelcase-keys');
 const R = require('ramda');
 const snakecase = require('./snakeCase');
 const snakecaseKeys = require('snakecase-keys');
+const logger = require('../logger');
```
Schnitzel: I don't think this is needed in here?

Schnitzel (changes requested):

Schnitzel:
wow, epic work @srowlands <3
one thing I'm not sure about the whole thing, we're adding a new GraphQL Query `getProjectsByMetadata`, but there is also the Query `allProjects` already that we could just extend that it supports to be filtered by Metadata.
I think this goes a bit into the architecture of GraphQL and might be a bit an esoteric question, but my gut feeling tells me that as a user of the GraphQL I would rater use an already existing query and just pass it more data to it.
Happy to hear the opinion of others though.

seanhamlin:
❤️ this Stu! The only thing I would add is some simple graphQL tests to prove that the functionality works as designed.

e.g. 

* Add a key/value pair
* Query for it (assert can find 1 project)
* Remove it
* Query for it (assert cannot find any projects)

This can be picked up by the team as well, if you are slammed (let me know).

services/api/src/resources/project/resolvers.js
stooit: ha yeah, sort-of relational/nosql hybrid

stooit:
🙏 Thanks for the prompt review + feedback!

Have resolved the minor bits and pieces, keen to hear feedback on whether this should be rolled into the `allProjects` query too.

Agree on tests, I can circle back to this next mid next week unless someone picks that up in the meantime

Schnitzel (approved):

PR #1776, opened 2020-04-07
Update project logic for Harbor projects

cdchris12:
# Checklist

This PR updates the API logic used when creating and deleting a project in Lagoon so that it will work seamlessly with Harbor to add and remove those projects from Harbor when they are created or deleted within Lagoon.


services/api/src/resources/project/harborSetup.js
```diff
@@ -151,6 +151,57 @@ const createHarborOperations = (sqlClient /* : MariaSQL */) => ({
...
+
+    // Delete harbor project
+    try {
+      var res = await harborClient.delete(`projects/${harborProjectID}`);
+      logger.debug(`Harbor project ${lagoonProjectName} deleted!`)
+    } catch (err) {
+      // 400 means the project id is invalid
+      // 404 means project doesn't exist
+      // 412 means project still contains repositories
+      logger.info(`Unable to delete the harbor project "${lagoonProjectName}", error: ${err}`)
```
Schnitzel: let's make this an `logger.error`

services/api/src/resources/project/harborSetup.js
```diff
@@ -151,6 +151,57 @@ const createHarborOperations = (sqlClient /* : MariaSQL */) => ({
...
+      const res = await harborClient.get(`projects?name=${lagoonProjectName}`)
+      var harborProjectID = res.body[0].project_id
+      logger.debug(`Got the harbor project id for project ${lagoonProjectName} successfully!`)
+    } catch (err) {
+      if (err.statusCode == 404) {
+        logger.warn(`Unable to get the harbor project id of "${lagoonProjectName}", as it does not exist in harbor!`)
+        return
+      } else {
+        logger.error(`Unable to get the harbor project id of "${lagoonProjectName}", error: ${err}`)
+      }
```
Schnitzel: you are catching an error here, but the code will continue below, is that intentionally? is the code below able to run without knowing the harborProjectID?

Schnitzel (changes requested):

services/api/src/resources/project/harborSetup.js
cdchris12: Good catch. Trying to continue past not being able to find this value will lead to the API crashing.

services/api/src/resources/project/resolvers.ts
```diff
@@ -475,6 +475,10 @@ export const deleteProject: ResolverFn = async (
     logger.error(`Could not delete default user for project ${project.name}: ${err.message}`);
   }
 
+  const harborOperations = createHarborOperations(sqlClient);
+
+  const harborResults = await harborOperations.deleteProject(project.name)
```
Schnitzel:
```suggestion
  // @TODO discuss if we want to delete projects in harbor or not
  //const harborOperations = createHarborOperations(sqlClient);

  //const harborResults = await harborOperations.deleteProject(project.name)
```

Schnitzel (changes requested):

PR #1784, opened 2020-04-10
Disable caching of large binary files in Varnish

### LINKED ISSUE #1613: Disable caching of large binary files in Varnish
**Describe the bug**

In order to protect the soon to be 500MB of Varnish cache (see #1585), Varnish should avoid caching static files that are > 10MB. This is because they are likely going to be served from nginx very quickly, so the overhead of Varnish storing them is pointless.

**Expected behavior**

Large binary files are not cached in Varnish. Perhaps there could be a new environment variable to make this configurable as well.

seanhamlin:
# Checklist

Don't cache large files, likely to use up all of Varnish's memory when they can be served by Nginx easily. Note, there are 2 commits, one is just code style (no functional changes), and the other is what needs reviewing.

# Closing issues
Closes #1613


images/varnish-drupal/drupal.vcl
```diff
@@ -147,23 +151,27 @@ sub vcl_recv {
      return (pipe);
   }
 
-  # We only try to cache  GET and HEAD, other things are passed
+  # Large binary files are passed.
+  if (req.url ~ "\.(msi|exe|dmg|zip|tgz|gz|pkg)") {
```
smlx: Is this going to catch anything that isn't already caught by the `Content-Length` check? I mean, it will catch very small binaries, but does it matter if these are cached?

seanhamlin: Yeah, this is mainly to prevent the files < 10MB that are pure static binary files. These files are fairly pointless to have in Varnish cache.

smlx: OK, makes sense. Should you put a `$` at the end of the regex to make it a bit more specific?

smlx: @seanhamlin ping

seanhamlin: 👍

images/varnish-drupal/drupal.vcl
```diff
@@ -330,6 +329,13 @@ sub vcl_backend_response {
     set beresp.http.Expires = "" + (now + beresp.ttl);
   }
 
+  # Files larger than 10 MB get streamed.
+  if (beresp.http.Content-Length ~ "[0-9]{8,}") {
+    set beresp.do_stream = true;
+    set beresp.uncacheable = true;
+    set beresp.ttl = 0s;
```
smlx: I think this value could result in request serialisation as per https://book.varnish-software.com/4.0/chapters/VCL_Basics.html#vcl-backend-response. Maybe it should be e.g. 120s?

smlx (changes requested):

images/varnish-drupal/drupal.vcl
seanhamlin: Great point, I forgot the hit-for-pass should be cached.

smlx (approved):

smlx:
LGTM

PR #1787, opened 2020-04-10
Drush 9 site alias for sql-sync

### LINKED ISSUE #1786: sql-sync fails with Drush 9
**Describe the bug**
The `drush sql-sync` task fails on Drupal 8 with Drush 9. See https://lagoon.readthedocs.io/en/latest/using_lagoon/drupal/drush9/ which requires aliases to be prefixed as `@lagoon.ALIAS`

```
In SqlSyncCommands.php line 77:
                                                            
  Error: no alias record could be found for source @master  
```

**To Reproduce**
1. Go to Tasks'
2. Click on tasks
3. Scroll down to sql-sync
4. Run
5. See it errors

mglaman:
# Checklist

This ensures sql-sync uses the `@lagoon` prefix.

Closes #1786

### DIFF
--- a/services/api/src/resources/task/resolvers.js
+++ b/services/api/src/resources/task/resolvers.js
@@ -403,11 +403,19 @@ const taskDrushSqlSync = async (
     project: destinationEnvironment.project,
   });
 
+  const command =
+  'drush_version=$(drush version --format=list) && \
+if [ ${drush_version%.*.*} -gt "8" ]; then \
+  drush -y sql-sync @lagoon.' + sourceEnvironment.name + ' @self; \
+else \
+  drush -y sql-sync @' + sourceEnvironment.name + ' @self; \
+fi';
+
   const taskData = await Helpers(sqlClient).addTask({
     name: `Sync DB ${sourceEnvironment.name} -> ${destinationEnvironment.name}`,
     environment: destinationEnvironmentId,
     service: 'cli',
-    command: `drush -y sql-sync @${sourceEnvironment.name} @self`,
+    command,
     execute: true,
   });
 


Schnitzel:
@mglaman 
thank you! Unfortunately we still have many people using Drush 8, which do not require the `lagoon.` prefix.
Maybe we could do something like we do for the cache clear, where we check which Drupal version is in use:
https://github.com/amazeeio/lagoon/blob/8753ec17303d4a13d2f31317ca4a7539c0396d06/services/api/src/resources/task/resolvers.js#L328-L337

we would need to check for the drush version, but it could be very similar.

mglaman:
I'll check that out! I knew it couldn't be a drastic change, especially if there were Drupal 7 sites. I'll work on an updated PR

mglaman:
@Schnitzel made a check on the Drush version, looks like this gets the proper version if called via Drush launcher or Drush directly. I use `-gt` to support Drush 10 as well.

services/api/src/resources/task/resolvers.js
```diff
@@ -403,11 +403,19 @@ const taskDrushSqlSync = async (
     project: destinationEnvironment.project,
   });
 
+  const command =
+  'drush_version=$(drush version --format=list) && \
+if [ ${drush_version%.*.*} -gt "8" ]; then \
+  drush -y sql-sync @lagoon.' + sourceEnvironment.name + ' @self; \
```
mglaman, line 409: Couldn't use template literals, since we have bash variables.

mglaman:
:( I don't know how to grok the pipeline to see what went wrong

tobybellwood:
we've fixed this in Lagoon now!

PR #1794, opened 2020-04-16
Destructive Action Button Color

### LINKED ISSUE #1748: UI - Destructive Actions should have a different Color
In the UI we have a few destructive actions e.g.

- Deleting an environment
- Deleting a SSH Key

those Buttons should be in a different color to underline that it's a destructive action.

shreddedbacon:
# Checklist

Destructive actions are currently blue, friendly looking buttons. They should not be blue.

image

# Closing issues
closes #1748


rocketeerbkw (changes requested):
A whole new component for a delete button feels wrong. The two buttons have the same functionality just with different colors.

I don't know what the "standard" is but material design, for example, uses `variant` and `color` to control styling. I think here we can add a color prop that defaults to `blue` and can accept `red`.

Also, there is another delete button in the [billing modifiers](https://github.com/amazeeio/lagoon/blob/master/services/ui/src/components/BillingModifiers/AllBillingModifiers.js#L72)

shreddedbacon:
@rocketeerbkw I've added `variant` as a prop. Defaults to blue.

rocketeerbkw (approved):

PR #1798, opened 2020-04-16
Remove legacy CLI image

smlx:
# Checklist

This image is no longer in use by amazee.io, having been superseded by https://github.com/amazeeio/lagoon-cli

# Closing issues
n/a


rocketeerbkw (changes requested):
Some more stragglers (found by regex search `cli\b`):

- https://github.com/amazeeio/lagoon/blob/master/jest.config.js#L4,L10
- https://github.com/amazeeio/lagoon/blob/master/package.json#L13
- https://github.com/amazeeio/lagoon/blob/master/.vscode/settings.json#L4

Also, did you just `rm` the `cli` folder? I wonder if we should also explicitly remove the node packages and clean up the shared yarn workspace `node_modules` folder. Some quick googling comes up blank with the correct way to "yarn delete workspace" though. Not sure if it's worth it to `cd cli && yarn remove <list> < every> <package> <here>` before deleting the cli directory. It might also be possible to just run `yarn install` with the `cli` folder gone?

smlx:
> Some more stragglers (found by regex search cli\b)

Ah thanks! I'll remove those..

> Also, did you just rm the cli folder?

Yes...

> I wonder if we should also explicitly remove the node packages and clean up the shared yarn workspace node_modules folder. Some quick googling comes up blank with the correct way to "yarn delete workspace" though. Not sure if it's worth it to cd cli && yarn remove <list> < every> <package> <here> before deleting the cli directory. It might also be possible to just run yarn install with the cli folder gone?

This is not an ecosystem I'm familiar with.. happy to take any expert advice.

Will `yarn remove cli` before deleting the directory work to update the workspace?

rocketeerbkw:
I'm no expert but I found a support channel in the yarn discord.

If you run `yarn` (with no arguments) in the root folder of the repo, it should detect that the cli workspace is gone and update the `yarn.lock` accordingly.  I'd expect a diff of that file to show a bunch of package versions being removed.

`yarn remove cli` doesn't work because `cli` isn't an actual package

smlx:
Yarn seemed to run correctly, :crossed_fingers: for CI

rocketeerbkw (approved):
Awesome!

Issue #1726, opened 2020-03-16
Pre-rollout tasks cause deployments to fail on newly created Lagoon environments.


cdchris12:
**Describe the bug**
When creating a new environment via the first push, if a `pre-rollout` task is defined, the build will fail. This is due to the fact that pre-rollout tasks are, by their nature, to be executed _before_ the built containers are rolled out. Thus, the attempted command fails with an error message similar to:
```
Error from server (NotFound): pods "null" not found
```

**To Reproduce**
1. Create a new branch of an already existing project, and ensure a `pre-rollout` task is defined in `lagoon.yml`.
2. Push the new branch in order to create a Lagoon build.
3. View logs once build fails.

**Expected behavior**
In the case that any pre-rollout task(s) cannot be completed due to the specified pods not existing because this is the first deploy to a Lagoon environment, the pre-rollout task(s) should simply be skipped.

Schnitzel:
mhh, this should be handled by https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/scripts/exec-pre-tasks-run.sh#L17-L19
but I think https://github.com/amazeeio/lagoon/pull/1529 broke that, as `$POD` is now not empty, but filled with `null`. 
I think @smlx can help us here?

rocketeerbkw:
I fixed this when I merged `master` into `kubernetes` in https://github.com/amazeeio/lagoon/commit/3168fffcb5234e6148dae14ae8ac38a29f32ad15#diff-aa5d30ee04ad43ee9ef99d2befa04f3c

PR #1802, opened 2020-04-20
Redeploy pods if only the configmap is changed

### LINKED ISSUE #1728: Updating a project's environment variables doesn't actually update those variables within running containers
**Describe the bug**
Whenever an environment variable is created or updated within a project, any running pods are not redeployed. This is due to the fact that the env vars are changed in the `lagoon-env` configmap, which is referenced by the pod. Since this configmap is not actually a part of the pod that's changing, the pod hasn't actually changed and isn't redeployed as a result.

**To Reproduce**
1. Create a new environment variable with a `global` scope
2. See that the newly created environment variable is not reflected in any of the project's containers.

**Expected behavior**
When environment variables are updated, containers should be redeployed accordingly to reflect any updates to those variables.

shreddedbacon:
# Checklist

Whenever an environment variable is created or updated within a project, any running pods are not redeployed. This is due to the fact that the env vars are changed in the lagoon-env configmap, which is referenced by the pod. Since this configmap is not actually a part of the pod that's changing, the pod hasn't actually changed and isn't redeployed as a result.

This PR aims to do some initial checks to see if only the configmap has changed, and redeploys any pods if it has.

Added bonus: ~It can check if this is the first time a deployment has occurred and won't try to run any `pre-rollout` tasks.~ Lagoon will now check for the existence of two new VARs during a deployment to see if it should run pre or post rollout tasks instead. These need to be defined in the API as `true`.
```
LAGOON_PREROLLOUT_DISABLED
LAGOON_POSTROLLOUT_DISABLED
```

# Closing issues
closes #1728 #1804

### DIFF
--- a/docs/using_lagoon/lagoon_yml.md
+++ b/docs/using_lagoon/lagoon_yml.md
@@ -111,6 +111,11 @@ Common uses for post-rollout tasks include running `drush updb`, `drush cim`, or
 * `shell`
   * Which shell should be used to run the task in. By default `sh` is used, but if the container also has other shells \(like `bash`, you can define it here\). This is useful if you want to run some small if/else bash scripts within the post-rollouts. \(see the example above how to write a script with multiple lines\).
 
+Note: If you would like to temporarily disable pre/post-rollout tasks during a deployment, you can set either of the following environment variables in the API at the project or environment level \(see how on [Environment Variables](environment_variables.md)\).
+
+* `LAGOON_PREROLLOUT_DISABLED=true`
+* `LAGOON_POSTROLLOUT_DISABLED=true`
+
 ## Routes
 
 ### `routes.autogenerate.enabled`
--- a/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
+++ b/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
@@ -324,7 +324,7 @@ do
     # STATEFULSET="${SERVICE_NAME}-galera"
     # SERVICE_NAME="${SERVICE_NAME}-maxscale"
     ## noone should be using `mariadb-galera`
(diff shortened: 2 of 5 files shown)


shreddedbacon:
This will definitely need some :eyes: over it. I haven't fully tested it locally yet.

rocketeerbkw:
>Added bonus: It can check if this is the first time a deployment has occurred and won't try to run any pre-rollout tasks.

Why? What is the problem this addresses? I didn't find an issue and never heard this getting discussed anywhere so I'm just unaware of the background. Are there no use cases where a pre-rollout on first deploy is useful? Could we accomplish the same thing by injecting a `FIRST_DEPLOYMENT` environment variable and then use that in the pre-rollout task itself (so projects can opt-out)?

shreddedbacon:
> > Added bonus: It can check if this is the first time a deployment has occurred and won't try to run any pre-rollout tasks.
> 
> Why? What is the problem this addresses? I didn't find an issue and never heard this getting discussed anywhere so I'm just unaware of the background. Are there no use cases where a pre-rollout on first deploy is useful? Could we accomplish the same thing by injecting a `FIRST_DEPLOYMENT` environment variable and then use that in the pre-rollout task itself (so projects can opt-out)?

Yeah I commented here sort of describing why and came to the same conclusion of a VAR. It has happened occasionally for a few PaaS customers.

https://github.com/amazeeio/lagoon/issues/1804#issuecomment-617515490

But basically, pre-rollout tasks won't run if the pods they need to run in don't exist. So if a pre-rollout task is defined, and it is a first deployment, it will fail until it is removed.

shreddedbacon:
I just found this #1726 so I guess the actual issue of pre-rollout tasks failing was already resolved.
Updated PR to remove my check.

Issue  #1804 of pre-rollout tasks failing in an unrecoverable way, being user controllable without having to ask an administrator for help could be handy, and might have helped with that particular issue. 

I updated the PR to include these new VARs to check for when a deployment occurs. Lagoon will run tasks unless these are set to `true`
```
LAGOON_PREROLLOUT_DISABLED
LAGOON_POSTROLLOUT_DISABLED
```

shreddedbacon:
Closing this as #1923 is a bit simpler

PR #1806, opened 2020-04-21
Add Harbor Migration Script

cdchris12:
This PR adds a script which can be used to create Harbor projects, create Harbor robot accounts, and add the corresponding environment variables to a project which does not currently have a Harbor project associated with it.

# Checklist

### DIFF
--- a/services/api/src/migrations/2-harbor/harborSync.ts
+++ b/services/api/src/migrations/2-harbor/harborSync.ts
@@ -18,13 +18,12 @@ import { query, prepare } from '../../util/db';
     // Get project's env vars
     var envVars = await query(sqlClient, `SELECT * FROM env_vars WHERE project='${project.id}';`);
     var hasVars = false;
-    if (envVars != null) {
-      // Check for Harbor variables
-      for (var j = 0; j < envVars.length; j++) {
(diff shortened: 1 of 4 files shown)


services/api/src/migrations/2-harbor/harborSync.ts
```diff
@@ -0,0 +1,42 @@
...
+  logger.info('Syncing Harbor projects with Lagoon projects');
+  // Get a list of all Lagoon projects
+  const projects = await projectHelpers(sqlClient).getAllProjects();
+  for(var i=0; i < projects.length; i++) {
+    var project = projects[i];
+
+    // Get project's env vars
+    var envVars = await query(sqlClient, `SELECT * FROM env_vars WHERE project='${project.id}';`);
+    var hasVars = false;
+    if (envVars != null) {
```
rocketeerbkw: Did you test this specific condition? Our sql library is weird and I don't remember if returning 0 results actually returns `null` or if the `!=` is coercing the actual result (I think an object?) to `null` and so it's true by accident.

cdchris12: Good catch; this was an artifact from my earlier testing attempts. I removed this logic, and the script still functions as intended.

rocketeerbkw (approved):

Issue #1818, opened 2020-04-28
"Make build" does not complete and gets stuck when building services/ssh Docker image


josephgultekin:
**Describe the bug**
"Make build" does not complete and gets stuck when building services/ssh Docker image

**To Reproduce**
1. git clone https://github.com/amazeeio/lagoon.git
2. make build
3. build does not complete because Ubuntu tzdata module prompts for user input for TZ selection when being installed. With docker build --quiet arg, this is never displayed and build never completes.

**Expected behavior**
Build to complete all image builds and going back to OS shell command prompt

**Screenshots**
see attached

**Additional context**
n/a

![Screenshot from 2020-04-28 11-25-13](https://user-images.githubusercontent.com/2167226/80438709-259e3f80-8948-11ea-8431-566c8b8e8d5b.png)

xantrix:
I got the same issue, good catch! thanks for the PR :tada:

smlx:
This was closed in the PR above.

PR #1819, opened 2020-04-28
Fix for issue #1818. Pin ubuntu image version as 18.04 in ssh/Dockerfile

josephgultekin:
Please see related issue #1818 
Was able to reproduce both on MacOS and Ubuntu 18.04 VM with Docker 20.x

See attached screenshot of completed build after proposed fix was applied.

Screenshot from 2020-04-28 11-45-45


smlx (approved):
This is fine, but I'll add an issue to the backlog to convert this image to alpine.

PR #1830, opened 2020-04-30
Re-implement JWT generation script using pyjwt

smlx:
# Checklist

I was cleaning up local branches and came across this patch.. I thought I already submitted a PR months ago but I guess not. I think this was from AIOSEC-3.

This is a rewrite of the token generation script in python instead of bash. The advantage is that it makes the script much easier to understand and takes advantage of a widely used library (https://github.com/jpadilla/pyjwt).

# Closing issues
n/a


Schnitzel:
we have 3 other places where we generate jwts in bash: 

https://github.com/amazeeio/lagoon/blob/lagoon.sh-prefix/local-dev/api-data-watcher-pusher/create_jwt.sh
https://github.com/amazeeio/lagoon/blob/lagoon.sh-prefix/services/storage-calculator/create_jwt.sh
https://github.com/amazeeio/lagoon/blob/lagoon.sh-prefix/services/auto-idler/create_jwt.sh

they all install nodejs to use jwtgen
I guess it would make sense to also replace these with the pyjwt?

services/storage-calculator/create_jwt.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+
+import os
+import jwt
+
+payload = {'role': 'admin', 'iss': 'storage-calculator',
```
smlx, line 6: The `sh` script used `iss`/`sub`: `auto-idler` here, I changed it to `storage-calculator`. Not sure if that will break.. does the API do any checks on this field?

Schnitzel: no, should be fine

PR #1843, opened 2020-05-05
Improve harbor migration script

### LINKED ISSUE #1834: Improve Harbor Migration Script
- add check if an INTERNAL_REGISTRY_URL is currently set
  - if it is set to another harbor url than the default lagoon one, don't change anything at all (lagoon should assume that this project uses a custom internal registry)
  - if it is set to the default  lagoon harbor, we continue
  - if nothing is set, we continue
- if robot account in harbor already exists, create a new one (with a new hash, so leave the old one intact)
- forcefully update the 3 INTERNAL_REGISTRY_* env variables in lagoon api with the new robot account infos

cdchris12:
# Checklist

This PR substantially overhauls the Harbor project creation process, allowing for a complete Harbor integration within Lagoon.

# Closing issues
Closes #1834


services/api/src/resources/project/harborSetup.js
```diff
@@ -24,133 +25,293 @@ const lagoonWebhookAddress = R.compose(
...
-    try {
-      const res = await harborClient.get(`projects?name=${lagoonProjectName}`)
-      var harborProjectID = res.body[0].project_id
-      logger.debug(`Got the harbor project id for project ${lagoonProjectName} successfully!`)
-    } catch (err) {
-      if (err.statusCode == 404) {
-        logger.error(`Unable to get the harbor project id of "${lagoonProjectName}", as it does not exist in harbor!`)
-      } else {
-        logger.error(`Unable to get the harbor project id of "${lagoonProjectName}" !!`)
+async function removeRobot(sqlClient /* : MariaSQL */, harborClient /* : harborClient */, lagoonProjectName /* : string */, harborProjectID /* : string */, robotName /* : string */) {
```
Schnitzel: this code is never used? I guess it can be removed then?

services/api/src/resources/project/harborSetup.js
```diff
@@ -24,133 +25,293 @@ const lagoonWebhookAddress = R.compose(
...
-        sqlClient,
-        Sql.insertEnvVariable({
-          "name": "INTERNAL_REGISTRY_URL",
-          "value": lagoonHarborRoute,
-          "scope": "INTERNAL_CONTAINER_REGISTRY",
-          "project": projectID,
-        }),
-      );
-      logger.debug(`Environment variable INTERNAL_REGISTRY_URL for ${lagoonProjectName} created!`)
+async function removeAllRobots(sqlClient /* : MariaSQL */, harborClient /* : harborClient */, lagoonProjectName /* : string */, harborProjectID /* : string */) {
```
Schnitzel: this code is never used? I guess it can be removed then?

services/api/src/migrations/2-harbor/harborSync.ts
```diff
@@ -5,33 +5,66 @@ import createHarborOperations from '../../resources/project/harborSetup';
...
+      } else if (envVars[j].name == "INTERNAL_REGISTRY_PASSWORD") {
+        hasPass = true;
       }
     };
 
-    if (! hasVars) {
+    // Filter for projects using this Lagoon's Harbor route
+    if (localLagoonURL == lagoonHarborRoute) {
+      logger.info('Lagoon project: ', project.name, " is using this Lagoon's Harbor; syncing new robot account for it.");
+      await harborClient.addProject(project.name, project.id, `${project.name}-${timestamp}`);
```
Schnitzel: I don't understand why we can run `addProject` with a robotName injected as attribute and not, as both cases you do the same:
generate the robot based on the timestamp. Just first time you do it directly in `harborSync.ts` (for the case of `localLagoonURL == lagoonHarborRoute`) and the second time for the case of `localLagoonURL == ""` it happens directly within `harborClient.addProject`.

In my understanding we could always generate the robot name in `harborClient.addProject` no?

Schnitzel (changes requested):

services/api/src/migrations/2-harbor/harborSync.ts
cdchris12: This convention is used here so that we can continue to use the same `harborClient.addProject` function both here in the migration script and in the kubernetesBuildDeploy system. The k8sBuildDeploy system will pass in a different value for the initially created robot name, so this supports both use cases.

services/api/src/resources/project/harborSetup.js
cdchris12: Eventually, we will need to handle the removal of older robot accounts; this function will be relevant when we want to implement that removal functionality.

services/api/src/resources/project/harborSetup.js
cdchris12: Eventually, we will need to handle the removal of older robot accounts; this function will be relevant when we want to implement that removal functionality.

services/api/src/resources/project/harborSetup.js
Schnitzel: okay, but then please let's not ship code that is not used anywhere, remove the code from this PR, and add it to another branch that we can then use again when the time is right for the removal functionality.

services/api/src/resources/project/harborSetup.js
Schnitzel: okay, but then please let's not ship code that is not used anywhere, remove the code from this PR, and add it to another branch that we can then use again when the time is right for the removal functionality.

cdchris12: Understood. Function removed and stored in the `add_harbor_robot_removal` branch.

services/api/src/resources/project/harborSetup.js
cdchris12: Understood. Function removed and stored in the `add_harbor_robot_removal` branch.

services/api/src/migrations/2-harbor/harborSync.ts
Schnitzel: > The k8sBuildDeploy system will pass in a different value for the initially created robot name

can you show me where this happens?

cdchris12: I am entirely incorrect about this. :(  That functionality would be here, if it existed: https://github.com/amazeeio/lagoon/blob/b7f2525a1bf893f15c1bbc3929ecbc7b5c890b11/services/api/src/resources/project/resolvers.js#L361

I'll update this accordingly so that it always generates a new robot name.

Schnitzel: 👍

PR #1846, opened 2020-05-06
Use labels on kubernetes jobs to not block builds on cron/backups/tasks

### LINKED ISSUE #1652: kubernetesdeployqueue: only check for lagoon build jobs
currently kubernetesdeployqueue checks for any jobs that are running in the namespace:
https://github.com/amazeeio/lagoon/blob/f020b6321d5581e3adcb6a4d0e056ce48e7abde2/services/kubernetesdeployqueue/src/index.js#L82-L91

this can cause build to be delayed unnecessary if there are other non-build jobs also running (cronjobs and also backups use jobs as well)

therefore kubernetesdeployqueue should only check if there are lagoon build jobs running and ignore others

rocketeerbkw:
# Checklist

Adds a `lagoon.job.type` label for builds and tasks so that the kubernetes deploy queue service can block only on other builds for that namespace.

# Closing issues
closes #1652

### DIFF
--- a/services/kubernetesbuilddeploy/src/index.js
+++ b/services/kubernetesbuilddeploy/src/index.js
@@ -134,6 +134,9 @@ const messageConsumer = async msg => {
       "metadata": {
           "creationTimestamp": null,
           "name": buildName,
+          "labels": {
+            "lagoon.job.type": "build",
+          }
       },
       "spec": {
         "backoffLimit": 0,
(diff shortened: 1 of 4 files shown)


services/kubernetesbuilddeploy/src/index.js
```diff
@@ -134,6 +134,9 @@ const messageConsumer = async msg => {
       "metadata": {
           "creationTimestamp": null,
           "name": buildName,
+          "labels": {
+            "lagoon.job.type": "build",
```
Schnitzel: can we use the label `lagoon.sh/jobType` ? (see https://github.com/amazeeio/lagoon/pull/1835)

Schnitzel (changes requested):

services/kubernetesbuilddeploy/src/index.js
rocketeerbkw: Done!

PR #1852, opened 2020-05-07
First iteration of Lagoon Version Update Helper 

dasrecht:
Helper script that allows easier updating of the Versions that are used in the Dockerfiles

It takes a more centralized approach to update the ENV variables that are spread across several Dockerfiles. This way we can update a new version in the helper and run the helper against the lagoon repository and managing the versions in a more centralized way. I'ts a first iteration - I could see the code doing manual update branches and commits while running trough the update in the future.

# Checklist


smlx:
The `Makefile` contains similar information such as `DEFAULT_ALPINE_VERSION` which is passed as a docker `--build-arg`. Could we use that mechanism for these kinds of variables? Or does this this method using Ansible have some other advantages?

dasrecht:
true, i was putting it into a helper as the helper actually changes the dockerfiles for updates - As keeping track of updates is a bit a chore this shold make it easier.

I think the `DEFAULT_ALPINE_VERSION` has other implications to the build process. So i'd keep those two things seperate.

dasrecht:
@tobybellwood @smlx if we can get that merged it would be great - I just updated the versions to reflect the latest versions

- Drush from 8.3.2 > 8.3.5
- Update Composer 1.10.6 > 1.10.7
- Update Newrelic 9.9.0.260 > 9.10.1.263

tobybellwood (approved):
Looks ok to me.  It'd be nice to automate PRs for updates to these versions, but it's a first step

PR #1891, opened 2020-05-15
Run mysql_upgrade on existing databases after upgrade

### LINKED ISSUE #1890: Need to run a mysql_upgrade after mariaDB updates
As we have now added an update to MariaDB 10.4 (from 10.2), we need to run mysql_upgrade on all MariaDB server pods to avoid errors, and allow us the ability to upgrade between versions. 

This should probably be run on an entrypoint - but we'd need to isolate the cases where it needs to be run to avoid performance implications.

mysql_upgrade is included in the mariadb-server-utils package, so that will need to be added too.

tobybellwood:
This PR adds a mysql_upgrade command if there is an existing database.

Definitely open to any attempts to make it nicer!

# Checklist

# Closing issues
Closes #1890

### DIFF
--- a/images/mariadb/Dockerfile
+++ b/images/mariadb/Dockerfile
@@ -39,6 +39,7 @@ RUN \
     mariadb \
     mariadb-client \
     mariadb-common \
+    mariadb-server-utils \
     net-tools \
     pwgen \
     tzdata \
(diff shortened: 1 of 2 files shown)


tobybellwood:
In particular, it'd be great if there was a more efficient way to put the daemon up for the upgrade and bring down again, or check whether it's even needed without the daemon up?

Schnitzel (approved):

PR #1892, opened 2020-05-18
Auto-idler cronjobs

shreddedbacon:
# Checklist

Currently the auto-idler cronjobs that get defined fail to run in a k8s cronjob
```
$ /lagoon/cronjob.sh /idle-clis.sh
2020-05-18T04:52:03Z CRONJOB: /idle-clis.sh
jq: error (at <stdin>:1): Cannot iterate over null (null)
```
It seems a bit odd to have these run in a k8s cronjob anyway when the pod is always running, this PR moves them back to being defined in the template.


seanhamlin:
+1 from me. This would allow the auto idler to function after a Lagoon upgrade.

PR #1901, opened 2020-05-21
Alpha/MVP Grafana API integration

justinlevi:
# Checklist

Adds a simple grafana setup and provisions the datasource and a simple dashboard.

# Closing issues
https://github.com/amazeeio/lagoon/issues/1900


Schnitzel:
we should probably rename "grafana" to something else, as we're not creating a grafana base image here, but rather a lagoon service that uses grafana

tobybellwood:
We won't be implementing this functionality - Grafana should be implemented with a full prometheus stack

PR #1917, opened 2020-05-27
Ensure project and environment labels exist on OpenShift projects.

### LINKED ISSUE #1888: Add labels to lagoon managed openshift projects
Currently finding the namespaces for a project and/or environment requires assuming the pattern is `[project]-[environment]`. This might not work when environments are derived from branches with slashes in them, or if the `--openshiftProjectPattern` was non-standard.

### Desired solution

Add the following labels to openshift projects/namespaces.

- `lagoon/project` - project machine name
- `lagoon/environment` - environment name
- `lagoon/branch` - branch (if available)

You could then use these labels in queries using the `oc` binary.

```
$ oc get ns --selector=lagoon/project=foo
NAME                     STATUS    AGE
foo-develop      Active    1y
foo-master       Active    2y
foo-pr-736       Active    191d
foo-production   Active    2y
```

```
$ oc get ns --selector=lagoon/project=foo,lagoon/environment=production
NAME                     STATUS    AGE
foo-production   Active    2y
```

nicksantamaria:
# Checklist

1. Adds the following labels to openshift projects.
    * `lagoon.sh/project` - project machine name
    * `lagoon.sh/environment` - environment name

There should be consistency between kubernetes and openshift namespace labels now.

# Closing issues
Closes #1888

### DIFF
--- a/services/openshiftbuilddeploy/src/index.js
+++ b/services/openshiftbuilddeploy/src/index.js
@@ -326,7 +326,19 @@ const messageConsumer = async msg => {
     if (err.code == 404 || err.code == 403) {
       logger.info(`${openshiftProject}: Project ${openshiftProject}  does not exist, creating`)
       const projectrequestsPost = Promise.promisify(openshift.projectrequests.post, { context: openshift.projectrequests })
-      await projectrequestsPost({ body: {"apiVersion":"v1","kind":"ProjectRequest","metadata":{"name":openshiftProject},"displayName":`[${projectName}] ${branchName}`} });
+      await projectrequestsPost({
+        body: {
+          "apiVersion":"v1",
+          "kind":"ProjectRequest",
+          "metadata": {
+            "name":openshiftProject,
+            "labels": {
+              "lagoon.sh/project": safeProjectName,
+              "lagoon.sh/environment": safeBranchName
+            }
+          },
+          "displayName":`[${projectName}] ${branchName}`
+        } });
     } else {
       logger.error(err)
       throw new Error


nicksantamaria:
I couldn't find any existing documentation which described using the `--namespace` flag on `kubectl` or `oc`. If there is a place I can add some notes about using label selectors please point me in the right direction.

nicksantamaria:
A quick and dirty bash script for those who want to add these labels to existing projects. It requires `lagoon` and `oc`, both authenticated.

```bash
#!/usr/bin/env bash

while read PROJECT; do
  echo "Updating namespace labels for ${PROJECT}"
  while read ENVIRONMENT; do
    echo "- ${ENVIRONMENT}"
    PATCH="{\"metadata\": {\"labels\": {\"lagoon.sh/project\": \"${PROJECT}\", \"lagoon.sh/environment\": \"${ENVIRONMENT}\"}}}"
    oc patch ns -p "${PATCH}" "${PROJECT}-${ENVIRONMENT}"
  done < <(lagoon list environments --no-header --project="${PROJECT}" | awk '{print $2}')
done < <(lagoon list projects --no-header | awk '{print $2}')
```

seanhamlin:
I like it.

nicksantamaria:
Tests are still failing - but its not clear to me what the problem is. Can someone point me in the right direction?

Schnitzel:
@nicksantamaria 
I think the problem is that the unsanitized branch names does not fit the requirements of k8s for valid label values:
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

this is actually why we add the unsanitized into the pods as annotations instead of labels as annotations are less strict, example:
https://github.com/amazeeio/lagoon/blob/master/images/kubectl-build-deploy-dind/helmcharts/nginx/templates/_helpers.tpl#L69-L75

so we probably need to do the same: add branch names as annotations to namespaces

the test fails because the cronjob test uses the branch name `branch/cronjob` and so it fails.
The namespace test itself uses just a dash: https://github.com/amazeeio/lagoon/blob/master/tests/tests/features-kubernetes.yaml#L19
which is probably not good in catching all the special branch characters that are allowed in git, maybe you want to update that line with at least a `/` ?

nicksantamaria:
Hmm ok, in that case I think the branch label will not be useful. I'll just remove it rather than move it to an annotation. The reason being annotations don't add any functional value to `kubectl` and `oc`.

PR #1954, opened 2020-06-13
Update curl to `7.70.0-r0` in the PHP images

seanhamlin:
# Checklist
- [x] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [x] PR title is ready for changelog and subsystem label(s) applied

This updates the version of curl to `7.70.0-r0` to allow the PHP images to build again.

# Proof this works

```
$ make clean
$ make -j4 build/php__7.4-cli
docker build --quiet --build-arg LAGOON_VERSION=development --build-arg IMAGE_REPO=lagoon --build-arg ALPINE_VERSION=3.11 -t lagoon/commons -f images/commons/Dockerfile images/commons
sha256:70c0910409dc9822d0a52f46ace18c86661a7af1a7b55165cfca78e9aa42a615
touch build/commons
docker build --quiet --build-arg LAGOON_VERSION=development --build-arg IMAGE_REPO=lagoon --build-arg PHP_VERSION=7.4  --build-arg PHP_IMAGE_VERSION=7.4 --build-arg ALPINE_VERSION=3.11 -t lagoon/php:7.4-fpm -f images/php/fpm/Dockerfile images/php/fpm
sha256:a72fe0e1b70dad8c31aa067f8ddee2ee0994c9044e4cf49310031af8ce6bdae4
touch build/php__7.4-fpm
docker build --quiet --build-arg LAGOON_VERSION=development --build-arg IMAGE_REPO=lagoon --build-arg PHP_VERSION=7.4  --build-arg PHP_IMAGE_VERSION=7.4 --build-arg ALPINE_VERSION=3.11 -t lagoon/php:7.4-cli -f images/php/cli/Dockerfile images/php/cli
sha256:33306dac7cf00083d07a713c365db63d546132c88882cf0692da23a82e1da4d5
touch build/php__7.4-cli
```

Works again 🎉 

# Closing issues
Closes #1949

PR #1955, opened 2020-06-13
Add a new environment variable to control New Relic browser monitoring.

### LINKED ISSUE #1951: Use environment Variable to disable NewRelic Browser
Currently we only allow the enable/disable of NewRelic via an environment variable.

Enabling NewRelic also enables the Browser Instrumentation which could in some cases be an unwanted additional request sites do (Also as those scripts need to be referred to in the GDPR and Cookie Management Solutions)

As there's a low number of customers actively using the Browser Instrumentation on lagoon (as far as i can tell) we could go 2 ways.

- Leaving it Enabled by default and have an environment variable that allows easy disabling of the 
- Disabling it by default and needing the environment variable to allow it (this would also lead to less requests sites do whereas the impact should be minimal - but it will follow good privacy-first guidelines to not enable the feature by default)

seanhamlin:
# Checklist

Adds a new environment variable in the PHP images `NEWRELIC_BROWSER_MONITORING_ENABLED` to which defaults to `true` (existing behaviour). Setting this to `false` will prevent the additional JS being inserted into the HTML, and will disable NewRelic Browser monitoring (APM will be fine).

Likely this PR will fail until #1954 is merged.

What this changes the configuration to (running in a container):

```
php:/app$ grep -A0 -B8 NEWRELIC_BROWSER_MONITORING_ENABLED /usr/local/etc/php/conf.d/newrelic.disable
; Setting: newrelic.browser_monitoring.auto_instrument
; Type   : boolean
; Scope  : per-directory
; Default: true
; Info   : Enables or disables automatic real user monitoring ("auto-RUM").
;          When enabled will cause the agent to insert a header and a footer
;          in HTML output that will time the actual end-user experience.
;
newrelic.browser_monitoring.auto_instrument = ${NEWRELIC_BROWSER_MONITORING_ENABLED:-true}
```

# Closing issues
Closes #1951

### DIFF
--- a/docs/using_lagoon/docker_images/php-fpm.md
+++ b/docs/using_lagoon/docker_images/php-fpm.md
@@ -88,6 +88,7 @@ Environment variables are meant to contain common information for the PHP contai
 | :--- | :--- | :--- |
 | `NEWRELIC_ENABLED` | `false` | Enable NewRelic performance monitoring, needs `NEWRELIC_LICENSE` be configured. |
 | `NEWRELIC_LICENSE` | \(not set\) | NewRelic license to be used, Important: `NEWRELIC_ENABLED` needs to be set to`true` in order for NewRelic to be enabled. |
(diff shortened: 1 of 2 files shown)


dasrecht (approved):
Thanks a lot! Looks good to me!

PR #1959, opened 2020-06-16
Logging overhaul

smlx:
Logging overhaul

# Checklist
- [x] Affected Issues have been mentioned in the Closing issues section
- [x] Documentation has been written/updated
- [x] PR title is ready for changelog and subsystem label(s) applied

This PR overhauls the way that logs are collected and transferred for storage.

The changes to the existing system include:

* Unify logs collection between both Kubernetes and OpenShift. This infrastructure is compatible with both.
* Remote lagoon installations use mutual TLS as well as regular username/password auth to forward logs to the central logs-concentrator.
* Add an extension point for shipping logs to an external sink on a per-cluster and per log-type (container/application/router) basis.
* Reduce the number of components in the logging system by using fluentd exclusively.
* Make metrics available via a Prometheus metrics endpoint. This facilitates alerting for log flow issues.
* Parse nginx container logs to make them more useful.
* Rely on dynamic templates for Elasticsearch rather than defining our own.

This PR also includes two "logs-tee" services for application and router logs. These are available to assist migration in the case where the new logging system is installed alongside the old one. The idea is that the current endpoints `application-logs.lagoon.svc.cluster.local` and `router-logs.lagoon.svc.cluster.local` are updated to refer to the two "tee" services which will then copy the logs to both the old and the new infrastructure. Once the old infrastructure is decommissioned, the endpoints can be updated to point to the new infrastructure and those "tee" services can be removed via a flag in the helm chart.

Further work required:

* The image tags in the charts currently refer to this branch. Once this PR makes it into a release those can be updated to a version number.
* To address #1364 we need the environment type added as a tag. See PR #1960.
* The current container/application/router log infrastructure should be removed once this is merged.

# Closing issues

* Closes #1297.
* Closes #1298.
* Closes #1438. This may need some configuration depending on the exact logs required.
* Closes #75 (this may already have been completed).
* Supersedes/closes #1918.
* Supersedes/closes #1343.

PR #1960, opened 2020-06-16
Label namespaces with lagoon.sh/environmentType

### LINKED ISSUE #1364: Refactor Container Logs Elasticsearch Index Name creation to generate less indexes
Currently we just use the Kubernetes Namespace name as the indicator for index names of container logs inside elasticsearch: `container-logs-${record['namespace_name']}-${Time.at(time).strftime("%Y.%m")}`

The problem is that with having many PR environments we create a new index for each PR environment even though they only have a couple of hundred log entries. This makes the elasticsearch slow.

Therefore we would like to change the index name to separate between the environment type (where we currently have `production` and `development`):
- `container-logs-${lagoon-projectname}-${lagoon-environmenttype}-${Time.at(time).strftime("%Y.%m")}`

In order to do this we're missing two essential pieces in the payload of the container logs:
- lagoon project name
- lagoon environment type

Fluentd has the possibility to inject the labels of the namespace into the payload of the container logs, therefore the plan is:

- [ ] Inject the Lagoon ProjectName, EnviornmentName, EnvironmentType as a label into the Kubernetes Namespace when we create the namespace, but also when we redeploy (in order to handle namespaces that don't have the labels yet, but also in case the environment type change)
- [ ] create a script that add/updates the labels in all the namespaces created by lagoon
- [ ] Update the Fluentd config https://github.com/amazeeio/lagoon/blob/master/services/logs-forwarder/.lagoon.multi.yml#L223-L228 to check for these new variables and if they exist use them instead of the current way (but provide a fallback to use the old system, plus also for namespaces that are not created by lagoon - like `default`)

smlx:
# Checklist

This PR makes information about Lagoon environment types available in the Kubernetes API. This facilitates, among other things, a fix for #1364.

# Closing issues
In combination with #1959, closes #1364.

### DIFF
--- a/services/kubernetesbuilddeploy/src/index.ts
+++ b/services/kubernetesbuilddeploy/src/index.ts
@@ -318,7 +318,8 @@ const messageConsumer = async msg => {
           "name":openshiftProject,
           "labels": {
             "lagoon.sh/project": projectName,
-            "lagoon.sh/environment": environmentName
+            "lagoon.sh/environment": environmentName,
+            "lagoon.sh/environmentType": lagoonEnvironmentType
           }
         }
       }
--- a/services/openshiftbuilddeploy/src/index.ts
+++ b/services/openshiftbuilddeploy/src/index.ts
@@ -333,7 +333,8 @@ const messageConsumer = async msg => {
             "name":openshiftProject,
             "labels": {
               "lagoon.sh/project": safeProjectName,
-              "lagoon.sh/environment": safeBranchName
+              "lagoon.sh/environment": safeBranchName,
+              "lagoon.sh/environmentType": lagoonEnvironmentType
             }
           },
           "displayName":`[${projectName}] ${branchName}`


smlx:
Converted to draft as I'm not sure the openshift labels are being applied

services/kubernetesbuilddeploy/src/index.ts
```diff
@@ -318,7 +318,8 @@ const messageConsumer = async msg => {
           "name":openshiftProject,
           "labels": {
             "lagoon.sh/project": projectName,
-            "lagoon.sh/environment": environmentName
+            "lagoon.sh/environment": environmentName,
+            "lagoon.sh/environmenttype": lagoonEnvironmentType
```
Schnitzel:
```suggestion
            "lagoon.sh/environmentType": lagoonEnvironmentType
```

services/openshiftbuilddeploy/src/index.ts
```diff
@@ -333,7 +333,8 @@ const messageConsumer = async msg => {
             "name":openshiftProject,
             "labels": {
               "lagoon.sh/project": safeProjectName,
-              "lagoon.sh/environment": safeBranchName
+              "lagoon.sh/environment": safeBranchName,
+              "lagoon.sh/environmenttype": lagoonEnvironmentType
```
Schnitzel:
```suggestion
              "lagoon.sh/environmentType": lagoonEnvironmentType
```
we use already this way in https://github.com/amazeeio/lagoon/blob/6d08071fad9619d9fe2bd54a77f18124a66c9691/images/kubectl-build-deploy-dind/helmcharts/cli/templates/_helpers.tpl#L58

Schnitzel (changes requested):

services/kubernetesbuilddeploy/src/index.ts
smlx: :+1:

services/openshiftbuilddeploy/src/index.ts
smlx: :+1:

smlx:
Setting this as ready for review. Openshift labelling doesn't work, but that's covered in https://github.com/amazeeio/lagoon/issues/1969.

PR #1979, opened 2020-06-21
Adds healthz-php to php fpm

### LINKED ISSUE #2011: Add healthz-php to PHP and Nginx images
[Healthz-php](https://github.com/amazeeio/healthz-php) is an amazee.io project that runs a series of checks on application environments hosted on Lagoon and returns the status of the component services.
It is a more detailed, lighter weight, and faster endpoint for health checks than, for instance, spinning up an entire Drupal site.

The issue described here is making Healthz-php and exposing it at an endpoint `/.lagoonhealthz`

bomoko:
This makes healthz-php available at `/.lagoonhealthz` and provides a fallback for non PHP supported applications.

# Checklist

See #2011 

# Closing issues
Closes #2011


images/nginx/healthcheck/README.md
```diff
@@ -0,0 +1,10 @@
+# Healthcheck
+
+In this directory you'll find two files
+
+- healthz.locations.php.disable
+- healthz.locations
+
+Both are designed to expose a `/healthz` location from the nginx service. The difference being that the `.php.disable` file is used to point to the [healthz-php](https://github.com/amazeeio/healthz-php) application _if_ there is a PHP service attached to this application.
```
seanhamlin:
```suggestion
Both are designed to expose a `/.lagoonhealthz` location from the nginx service. The difference being that the `.php.disable` file is used to point to the [healthz-php](https://github.com/amazeeio/healthz-php) application _if_ there is a PHP service attached to this application.
```

images/php/fpm/Dockerfile
```diff
@@ -3,6 +3,21 @@ ARG PHP_IMAGE_VERSION
...
+# Defining Versions - Composer
+# @see https://getcomposer.org/download/
+ENV COMPOSER_VERSION=1.10.7 \
+  COMPOSER_HASH_SHA256=b94b872729668de5b5fbf62f16ff588d2a23480dda88c0e45cb43b721b75ae29
+
+RUN curl -L -o /tmp/composer https://github.com/composer/composer/releases/download/${COMPOSER_VERSION}/composer.phar \
+    && echo "$COMPOSER_HASH_SHA256 /tmp/composer" | sha256sum \
+    && chmod +x /tmp/composer \
+    && php -d memory_limit=-1 /tmp/composer create-project amazeeio/healthz-php /healthz-php v0.0.3 \
+    && rm /tmp/composer
```
seanhamlin: Move this to an upstream composer build image, then we can avoid the whole download, install step.

seanhamlin (changes requested):
The main thing I think is missing is docs for the new ENV VARS, e.g. `FAST_HEALTH_CHECK` to which can be enabled on a per project basis to speed up uptime checks. The default is for this not to be set.

images/php/fpm/Dockerfile
```diff
@@ -3,6 +3,21 @@ ARG PHP_IMAGE_VERSION
 ARG ALPINE_VERSION
 ARG IMAGE_REPO
 FROM ${IMAGE_REPO:-lagoon}/commons as commons
+
+FROM php:${PHP_IMAGE_VERSION}-fpm-alpine${ALPINE_VERSION} as healthcheckbuilder
```
bomoko: Use composer images (request/suggestion from @seanhamlin )

bomoko: ah, mentioned above. Resolving this.

images/php/fpm/Dockerfile
bomoko: done!

images/nginx/healthcheck/README.md
bomoko: sorted out

images/nginx/docker-entrypoint
```diff
@@ -15,8 +15,23 @@ ep /etc/nginx/*
...
+echo "Setting up Healthz routing"
+if [ ! -z "$NGINX_FASTCGI_PASS" ]; then
+  echo "Setting up Healthz routing - using PHP"
+  cp /etc/nginx/conf.d/healthz.locations.php.disable /etc/nginx/conf.d/healthz.locations
+else
+  echo "Setting up Healthz routing - using Lua as fallback"
+  cp /etc/nginx/conf.d/healthz.locations.lua.disable /etc/nginx/conf.d/healthz.locations
+fi
+
+if [ "$FAST_HEALTH_CHECK" == "on" ]; then
```
Schnitzel: can we use `true` instead of `on`?

images/nginx/healthcheck/healthz.locations
```diff
@@ -0,0 +1,8 @@
+location /.lagoonhealthz {
+ content_by_lua_block {
+  ngx.status = ngx.HTTP_OK;
+  ngx.header.content_type = 'application/json';
+  ngx.say('{"check_nginx":"pass"}');
+  ngx.exit(ngx.OK);
+ }
+}
```
Schnitzel, line 8: is this file supposed to be here?

Schnitzel (changes requested):

Schnitzel:
@bomoko nice work! just a couple of small questions :)

images/nginx/docker-entrypoint
bomoko: Sure! Will change!

images/nginx/healthcheck/healthz.locations
bomoko: Hi @Schnitzel - yeah, I was half and half about including it - it's essentially a duplicate of the Lua one - but I was thinking through edge cases and I was worried that there _may_ be a case where the conditional that copies the "disabled" files doesn't run, but the reference to the `healthz.locationz` file might still be there:

I was wondering what might happen if the entrypoint was overwritten, but since the helper files are there, and the healthz.locations file is referenced _from inside a helper_, you might get the situation where there was a reference to a file that isn't there (since the entrypoint file no longer copies `healthz.locations`)

But then I got really paranoid and though, okay, that's fine, maybe I ship the default as LUA and only have the entrypoint copy the php version across if it's set ... but then I thought "if I do that, I'll be overwriting the LUA default and won't be able to get it back until the container is recreated).

So I went with the, admittedly, non-DRY solution on this - I'm happy to reign in my paranoia though.
The options are:

1 - have the duplication, but be assured of always having a healthz.location file existing (maybe it could be an empty file with some comments in it? I haven't tried this yet)

2 - have the default be Lua (so we're guaranteed a healthz.locations file exists), and only ever overwrite with the php version

3 - don't ship a default healthz.locations file, and accept that if someone (or some script) doesn't explicitly copy over one of the .disabled files, nginx is gonna die.

Which do you think is best?

Schnitzel: @bomoko 
Thanks for outlining your thought process! Yes I agree all versions 1-3 would make sense. I think though that the current state is confusing and looks like a bug, so I'm worried that people that want to understand what happens here are having a hard time to understand it.
therefore I think version 2 is the best, it actually fits to what we already do in the nginx image:
we ship with a static file version for `app.config` in the `nginx` image: https://github.com/amazeeio/lagoon/blob/93490ded5be6ac97f5415995c9d577f917d6ec69/images/nginx/Dockerfile#L37
which is then overwritten by the nginx-drupal: https://github.com/amazeeio/lagoon/blob/93490ded5be6ac97f5415995c9d577f917d6ec69/images/nginx-drupal/Dockerfile#L10

so I think the system of overwriting is a good one.

bomoko: Great! Done. The build failed, so I'll kick off a new one - tested, all seems to be back in order.

PR #1981, opened 2020-06-22
Allow additional options for configuring auto generated routes

### LINKED ISSUE #1807: Allow to disable auto generated routes on Branches
Currently, autogenerate is set to false, it blocks generation of routes across the entire project, #505. Would it be possible to block autogeneration on branches defined in the branches field on the project instead? This would allow for our PR branches to continue to get autogenerated routes, whilst not re-adding the routes to defined branches.

 ```
     {
        "name": "governors",
        "branches": "^(master|develop|production)$"
      },
 
```

### PR DESCRIPTION
**Pull request by shreddedbacon**
# Checklist

Currently autogenerated routes are either enabled or disabled. If disabled, there is no way for any environments to get an autogenerated route. This PR allows for two new options

* `routes.autogenerated.allowPullrequests` which will allow PR environments to get autogenerated routes
* `environments.${env/branch}.autogenerateRoutes` which if set to true, and `routes.autogenerate.enabled = false` will create the routes for this environment.

# Closing issues
closes #1807

### CODE REVIEW

**Comment by Schnitzel** [comment]
I really like the idea behind this, but I think we're going down a rabbithole that we only hardly can get out of it again:

We're also thinking about adding regex support or deploymenttype support (pr vs. branches, etc.) for the `environments:` key (this key currently only allows explicit environment names).

so if we would have the possibility to do this:

```
environments:
  features\/.*:
    cronjobs:
      - name: drush cron
        schedule: "*/15 * * * *"
        command: drush cron
        service: cli
```
^^ would add a cronjob to all environments that match `features/.*`

then we could also do:

```
environments:
  features\/.*:
    autogenerateRoutes: true
```
^^ would enable autogenerated routes for all all environments that match `features/.*`

but I'm not sure how fast we will have this environment regex matching.

@shreddedbacon 
could we maybe for now just add `autogenerateRoutes: true` (or something similar) to the environment section? this means the customer need to set this for each individual branch though. But this would be more future compatible.

shreddedbacon:
> I really like the idea behind this, but I think we're going down a rabbithole that we only hardly can get out of it again:
> 
> We're also thinking about adding regex support or deploymenttype support (pr vs. branches, etc.) for the `environments:` key (this key currently only allows explicit environment names).

Did not know this was going to be the way forward

> ```
> environments:
>   features\/.*:
>     autogenerateRoutes: true
> ```
> 
> ^^ would enable autogenerated routes for all all environments that match `features/.*`
> 
> but I'm not sure how fast we will have this environment regex matching.
> 
> @shreddedbacon
> could we maybe for now just add `autogenerateRoutes: true` (or something similar) to the environment section? this means the customer need to set this for each individual branch though. But this would be more future compatible.

Will refactor, shouldn't be too difficult.

tobybellwood:
The problem comes for this and other issues where the environment name isn't known or is changeable eg feature or PR envs.  To get an auto route for a PR would mean adding it to the lagoon yml before you know what it is...

shreddedbacon:
> The problem comes for this and other issues where the environment name isn't known or is changeable eg feature or PR envs. To get an auto route for a PR would mean adding it to the lagoon yml before you know what it is...

I was going to leave in the `allow_pullrequest` bit in this for this reason.

Issue #1908, opened 2020-05-25
Performance on the Lagoon dashboard is not great with lots of projects


seanhamlin:
**Describe the bug**

When you are a member of hundreds of projects, some screen transitions in the UI take a long time (~20 seconds).

**To Reproduce**
Steps to reproduce the behaviour:
1. Log into Lagoon, ensure you are _not a lagoon admin_, but you can see a lot of projects (> 100).
2. Click on any project name (to go to the project detail screen)
3. Notice how it takes a long time (20+ seconds)

**Expected behaviour**

The Lagoon UI is navigable at this scale. Sub 3 second load times would be ideal.

**The slow GraphQL query**

This can be seen in the browser console, this particular request took 18.33 seconds to return, for a single project. `PROJECTNAME` redacted to protect the Lagoon project name.

```graphql
{"operationName":"getProject","variables":{"name":"PROJECTNAME"},"query":"query getProject($name: String!) {
  project: projectByName(name: $name) {
    id
    name
    branches
    pullrequests
    created
    gitUrl
    productionEnvironment
    standbyProductionEnvironment
    developmentEnvironmentsLimit
    environments {
      id
      name
      created
      updated
      deployType
      environmentType
      openshiftProjectName
      project {
        id
        name
        productionEnvironment
        standbyProductionEnvironment
        __typename
      }
      __typename
    }
    __typename
  }
}
"}
```

Issue #1270, opened 2019-09-27
Switch from mariasql to mariadb


Schnitzel:
Looks like MariaSQL is not maintained anymore and also doesn't build on node 12 anymore, we should switch to https://www.npmjs.com/package/mariadb

blazeyo:
I've checked all the references to _mariasql_ and _sqlClient_ (496 results in 34 files) in the codebase and it turns out that all of them end up in either [query](https://github.com/amazeeio/lagoon/blob/master/services/api/src/util/db.js#L85) or prepare functions from _services/api/src/util/db.js_. This is most likely good news because it should make the swap relatively easy. I'll be experimenting with that now.

One things struck me during this review. If all the ~490 appearances of the text `sqlClient` end up in the same method, does it really make sense to pass them around? It looks like getting the client in the query/prepare methods would let us remove hundreds of lines of code (literally). Would it make sense to do that along with the package swap?

Here's a breakdown of files mentioning sqlClient:
- apolloServer.js
  - first, the client is passed on to `getGrantForKeycloakToken` from auth.js which isn't using it
  - then a new client is passed to `getCredentialsForLegacyToken` thich isn't using it either
  - then another client instance is created and put in the context (returned)
- authMiddleware.js
  - same as above, the client is only instantiated to be passed to `getGrantForKeycloakToken` and `getCredentialsForLegacyToken`. Both don't use it.
- auth.js
  - holds `getGrantForKeycloakToken` and `getCredentialsForLegacyToken`
- pubSub.js
  - the sqlClient is needed for the db/query function
- resolvers/index.js
  - `sqlClient: MariaSQL` used in a comment (flow type definition)
- sync-groups-opendistro-security.ts
  - the sqlClient is passed to opendistroSecurity.js, which passes it to project/helpers.js which passes it to the db/query function
- consumers-to-groups.js
  - instantiates the sqlClient and passes it to db/query and db/prepare
- projects.ts
  - instantiates the sqlClient and passes it to db/query and db/prepare
- reset-project-keys.ts
  - instantiates the sqlClient and passes it to db/query and db/prepare
- user.ts
  - instantiates the sqlClient and passes it to db/query and db/prepare
- backup/resolvers.js
  - uses the environment/helpers.js (which use db/query)
  - calls db/query directly
- deployment/helpers.js
  - uses the environment/helpers.js (which use db/query)
  - calls db/query directly
- backup/resolvers.js
[...]

PR #1983, opened 2020-06-22
Add newrelic to api

rocketeerbkw:
# Checklist

In an effort to get more info into performance issues (see #1908) this PR adds new relic tracking support to the API. We offer it to our amazee.io customers, may as well take advantage of it ourselves too!

This initial work is very basic. [SQL tracing](https://docs.newrelic.com/docs/agents/nodejs-agent/getting-started/compatibility-requirements-nodejs-agent#database) is only supported out of the box with the mysql/mysql2 packages. That's different than our planned switch to mariadb (#1270). There is no OOTB support for apollo/graphql, so the only data we currently get is timing info.

Instrumenting of external requests (keycloak) also looks to be somewhat bare at this time.

To enable in production, add an environment variable `NEW_RELIC_LICENSE_KEY`

# Closing issues
n/a

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -267,6 +267,10 @@ services:
       - keycloak
     ports:
       - '3000:3000'
+    # Uncomment for local new relic tracking
+    # environment:
+    #   - NEW_RELIC_LICENSE_KEY=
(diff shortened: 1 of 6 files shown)


justinlevi (approved):
This looks fine to me

PR #1989, opened 2020-06-23
Fix notification of error message to rocketchat/slack

### LINKED ISSUE #1971: Add helpers script to check routes' TLS status

vincenzodnp:
Fix the `check_acme_routes.sh` script's notification of error message in slack/rocketchat 

# Checklist

# Closing issues
closes #1971


helpers/check_acme_routes.sh
```diff
@@ -199,15 +200,15 @@ function notify_customer() {
...
+	# json Payload
+	JSON=\'"{\"channel\": \"$CHANNEL\", \"text\": \"${MESSAGE}\"}"\'
 
+	echo -e "Sending notification into ${CHANNEL}"
 	# Execute curl to send message into the channel
 	if [[ $DRYRUN = true ]]; then
 		echo "DRYRUN on \"$NOTIFICATION\" curl -X POST -H 'Content-type: application/json' --data "$JSON" "$WEBHOOK""
 	else
-		curl -X POST -H 'Content-type: application/json' --data "$JSON" "$WEBHOOK"
+		eval "curl -X POST -H 'Content-type: application/json' --data "${JSON}" ${WEBHOOK}"
```
Schnitzel: uh, I don't think I can merge this, see https://stackoverflow.com/questions/17529220/why-should-eval-be-avoided-in-bash-and-what-should-i-use-instead
eval in bash is very very bad, happy to sit together to see how we can make this run without eval

Schnitzel (changes requested):

helpers/check_acme_routes.sh
vincenzodnp: I removed eval and tested.

vincenzodnp:
@Schnitzel can you have a review?

helpers/check_acme_routes.sh
```diff
@@ -18,9 +22,15 @@ COMMAND=${1:-"help"}
 # Set DRYRUN variable to true to run in dry-run mode
 DRYRUN="${DRYRUN:-"false"}"
 
+# Set DEBUG variable to true, to print echo messages
```
Schnitzel: this also enables BASH debugging on line `9`, so I'm not sure it makes sense to use the same env variable for echo messages and also bash debugging

helpers/check_acme_routes.sh
```diff
@@ -18,9 +22,15 @@ COMMAND=${1:-"help"}
...
+DEBUG="${DEBUG:-"false"}"
+
 # Set a REGEX variable to filter the execution of the script
 REGEX=${REGEX:-".*"}
 
+# Set NOTIFYONLY to true if you want to only notify customers about their failed routes
```
Schnitzel: please describe what a "failed route" is, you don't use the term "failed" anywhere else in this script, so it's not really clear what this env variable really does.

helpers/check_acme_routes.sh
```diff
@@ -105,7 +115,9 @@ function create_routes_array() {
 	# Get the list of namespaces with broker routes, according to REGEX
 	for namespace in $(oc get routes --all-namespaces|grep exposer|awk '{print $1}'|sort -u|grep -E "$REGEX")
 	do
-		PROJECTNAME=$(oc get project "$namespace" -o=jsonpath="{.metadata.labels.lagoon\.sh/project}")
+		#PROJECTNAME=$(oc get project "$namespace" -o=jsonpath="{.metadata.labels.lagoon\.sh/project}")
```
Schnitzel: please remove any debugging or old code

helpers/check_acme_routes.sh
```diff
@@ -148,37 +172,46 @@ function check_routes() {
...
-				ocroute=($(echo "$j" | tr ";" "\n"))
-				OCROUTE_NAME=${ocroute[0]}
-				if [[ $DRYRUN = true ]]; then
-					echo -e "DRYRUN oc delete -n $ROUTE_NAMESPACE route $OCROUTE_NAME"
-				else
-					oc delete -n "$ROUTE_NAMESPACE" route "$OCROUTE_NAME"
-				fi
-			done
+			if [[ "$NOTIFYONLY" = "true" ]]; then
+				echo $NOTIFYONLY
```
Schnitzel: please remove debugging things

helpers/check_acme_routes.sh
```diff
@@ -148,37 +172,46 @@ function check_routes() {
 			fi
 
 			echo "$DNS_ERROR"
```
Schnitzel, line 170: is this echo supposed to be here?

helpers/check_acme_routes.sh
```diff
@@ -148,37 +172,46 @@ function check_routes() {
...
+
 	done
 }
 
 # Function to update route's annotation (ie: update tls-amce, remove tls-acme-awaiting-* and set a new one for internal purpose)
 function update_annotation() {
 	echo "Update route's annotations"
-	OCOPTIONS=""
+	OCOPTIONS="--overwrite"
 	if [[ "$DRYRUN" = "true" ]]; then
```
Schnitzel, line 203: sometimes you use `[[ "$DRYRUN" = "true" ]]` or `[[ "$DRYRUN" == true ]]`, please use one way in the whole script.

helpers/check_acme_routes.sh
```diff
@@ -197,21 +230,20 @@ function notify_customer() {
...
-	JSON="'{\"channel\": \"$CHANNEL\", \"text\":\"$MESSAGE\"}'"
-	echo "Sending message $JSON to $CHANNEL"
+	# json Payload
+	PAYLOAD="\"channel\": \"$CHANNEL\", \"text\": \"${MESSAGE}\""
 
+	echo -e "Sending notification into ${CHANNEL}"
 	# Execute curl to send message into the channel
 	if [[ $DRYRUN = true ]]; then
-		echo "DRYRUN on \"$NOTIFICATION\" curl -X POST -H 'Content-type: application/json' --data "$JSON" "$WEBHOOK""
+		echo "DRYRUN on \"$NOTIFICATION\" curl -X POST -H 'Content-type: application/json' --data "$PAYLOAD" "$WEBHOOK""
```
Schnitzel: during the actual curl command you wrap `$PAYLOAD` with `{}` I would suggest to do the same here to make sure the dryrun really shows the same as the acutal script will run.

Schnitzel (changes requested):
hey @vincenzodnp 
good work! but some more things needed to make the script more robust and better to understand.

also can you run another shellcheck over it?

helpers/check_acme_routes.sh
vincenzodnp: I change the comment in `Set DEBUG variable to true, to start bash in debug mode`.
I will remove debug echoes.

helpers/check_acme_routes.sh
vincenzodnp: Changed the comment in:
```
# Set NOTIFYONLY to true if you want to send customers a notification
# explaining why Lagoon is not able to issue Let'S Encrypt certificate for
# some routes defined in customer's .lagoon.yml file.
# If set to true, no other action rather than notification is done (ie: no annotation or deletion)
```

helpers/check_acme_routes.sh
vincenzodnp: removed

helpers/check_acme_routes.sh
vincenzodnp: removed

helpers/check_acme_routes.sh
vincenzodnp: Yes. I added a comment # Print the error on stdout

helpers/check_acme_routes.sh
vincenzodnp: I have already realized this, and I have already fixed. Wondering if you are looking at the latest version I pushed yesterday.

helpers/check_acme_routes.sh
vincenzodnp: :+1:

PR #2020, opened 2020-07-01
Actually list all openshifts defined in the API

shreddedbacon:
# Checklist

Just a minor thing when listing openshifts associated to a lagoon, it doesn't currently list all openshifts, only ones that have projects associated to it.

This change lists all openshifts now, but also crafts a `project_count/projectCount` that is returned too, to quickly see how many projects are on a cluster.

# Closing issues


services/api/src/typeDefs.js
```diff
@@ -265,6 +265,7 @@ const typeDefs = gql`
     sshPort: String
     created: String
     monitoringConfig: JSON
+    projectCount: Int
```
Schnitzel: @shreddedbacon mhh I think this is never resolved? This might be wrong here?

Schnitzel (changes requested):

services/api/src/typeDefs.js
shreddedbacon: Running this on a bare `make api-deployment`
```
query {
  allOpenshifts {
    id
    name
    projectCount
  }
}
```
Returns this
```
{
  "data": {
    "allOpenshifts": [
      {
        "id": 2,
        "name": "ci-local-openshift",
        "projectCount": 20
      },
      {
        "id": 3,
        "name": "kickstart-openshift",
        "projectCount": 1
      },
      {
        "id": 4,
        "name": "ui-openshift",
        "projectCount": 1
      },
      {
        "id": 1002,
        "name": "ci-local-kubernetes",
        "projectCount": 20
      }
    ]
  }
}
```

When returned as part of an `allProjects` query though, it returns null. To get it to return the full project count here would require a crafty sql select for `getOpenshiftByProjectId`
```
query {
  allProjects{
    id
    name
    openshift{
      name
      projectCount
    }
  }
}
```
```
{
  "data": {
    "allProjects": [
      {
        "id": 1,
        "name": "credentialstest-project1-openshift",
        "openshift": {
          "name": "ci-local-openshift",
          "projectCount": null
        }
      },...
    ]
  }
}

```

Schnitzel: mhh, I'm just confused what exactly handles `projectCount` as I cannot find any code that handles this? Is that maybe something in the graphql server?

shreddedbacon: I guess this is where my lack of understanding of JS / graphql server hits a limit. The original query just returns the rows:
https://github.com/amazeeio/lagoon/blob/4438e32128ff46a54d116fafdac2c14bf76e9984/services/api/src/resources/openshift/resolvers.ts#L62

I guess something in graphql server knows how to transform the rows into the `type Openshift`:
https://github.com/amazeeio/lagoon/blob/4438e32128ff46a54d116fafdac2c14bf76e9984/services/api/src/typeDefs.js#L717

If I remove `projectCount: Int` from the typDefs.js, then there was no way to actually see the returned projectcount.

I'm happy to take out the project count from this PR otherwise, I just thought it would be a nice to have in addition to listing all the openshifts.

Schnitzel: > If I remove projectCount: Int from the typDefs.js, then there was no way to actually see the returned projectcount.

mhh, I guess you are right. I'm just a bit worried about introducing something that A) we don't use anywhere else and B) doesn't work all the time.
Probably the correct way would be to add the capability to list all projects under an openshift like so:
```
query allopenshifts {
  allOpenshifts {
    name
    id
    projects {
      name
       id
    }
  }
}
```

and then if somebody is interested in all projects they would just count the project (via jq or so).

But as we're working on updating openshift -> clusters anyway I think we can just remove the `projectCount` and then let @rocketeerbkw and @justinlevi add this "project by openshifts" into their updated Lagoon 2.0 code :)

shreddedbacon: I'll just close this PR then, no point with the work in 2.0 happening.

Schnitzel: @shreddedbacon I like the work for showing all openshifts though! so let's get that one in, let's just leave the `projectCount` piece out :)

Issue #2023, opened 2020-07-02
Remove groups from getOpenShiftInfoForProject call to ensure performance of build queues


dasrecht:
**Describe the bug**
In  https://github.com/amazeeio/lagoon/commit/4438e32128ff46a54d116fafdac2c14bf76e9984 we added the ingressmonitorcontroller information to getOpenShiftInfoForProject as this request is performed a lot during builds we got very slow build run-trough times and high wait times for builds to start/finish.

We should remove the groups for the build related tasks and only query the group information if it's needed for the ingressmonitorcontroller.

PR #2024, opened 2020-07-02
#2023 remove groups from getOpenShiftInfoForProject

Schnitzel:
as it slows down the api by 40x and causes issues all around


twardnw:
Closing in favor of #2028

PR #2031, opened 2020-07-03
Add readinessProbe to CLI containers

### LINKED ISSUE #2017: add readiness check to CLI pods
CLI pods need a readiness check. It was recently found that because of the lack of a readiness check, a command issued via SSH against a CLI pod awaking from idle, the command will run at the same time as the entrypoint scripts, resulting in inconsistent behavior.

One solution raised would be a very high numbered entrypoint script which creates a file used for a readiness check.

twardnw:
# Checklist

# Changelog Entry
* Add an entrypoint which creates a file to be used by a readinessProb
* Add a readinessProbe which checks for `/tmp/ready`

# Closing issues
closes #2017


images/oc-build-deploy-dind/openshift-templates/cli-persistent/deployment.yml
```diff
@@ -135,6 +135,12 @@ objects:
                   memory: 10Mi
                 limits:
                   memory: 8Gi
+              readinessProbe:
+                initialDelaySeconds: 15
```
Schnitzel: let's use much less than 15secs here, having 15 secs here means k8s will wait for 15sec after the container has started before it tries the first time if the container is ready, aka this delays everything by 15 secs.
I think 1 sec is perfectly fine here.

images/php/cli/Dockerfile
```diff
@@ -74,5 +74,8 @@ COPY id_ed25519_lagoon_cli.key /home/.ssh/lagoon_cli.key
 RUN chmod 400 /home/.ssh/lagoon_cli.key
 ENV SSH_AUTH_SOCK=/tmp/ssh-agent
 
+# readinessProbe check
+COPY 999-readiness.sh /lagoon/entrypoints/
```
Schnitzel: Let's add this to the `commons` image where it will then end up in all docker images from lagoon, this solves two issues:
1. we can use the check for `/tmp/ready` also in other places
2. there might be some people that use the `cli` services with another base image than the php-cli Dockerimage, for these people I would at least hope that they use the lagoon commons image as a base and therefore they will also end up having this check inside of them.

images/oc-build-deploy-dind/openshift-templates/cli/deployment.yml
```diff
@@ -112,6 +112,13 @@ objects:
                   memory: 10Mi
                 limits:
                   memory: 8Gi
+              readinessProbe:
+                initialDelaySeconds: 15
+                exec:
+                  command:
+                  - cat
+                  - /tmp/ready
```
Schnitzel: I'm a bit worried that this could break existing cli images that don't have that file yet, can we extend this check that it actually checks if the lagoon version in `/lagoon/version` is `1.8.0` and higher and only then we check for the /tmp/ready file?

Schnitzel (changes requested):

images/oc-build-deploy-dind/openshift-templates/cli/deployment.yml
twardnw: Shouldn't this template only be used on a fresh deploy? Thus the newly running CLI pod would have the file?

Schnitzel: there is a technical possibility that somebody has v1.7 in their cli images defined and therefore this file never exists which means the test will never pass.
So I think we need to add a check for the lagoon version that this is introduced.

images/commons/entrypoint-readiness
```diff
@@ -0,0 +1,19 @@
+#! /bin/bash
```
Schnitzel: can we switch this to just `sh`? not all images have bash installed all the time.

images/commons/entrypoint-readiness
```diff
@@ -0,0 +1,19 @@
...
+function version_check() {
+  test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1";
+}
+
+function file_check() {
+  test -f /tmp/ready
+}
+
+VERSION="${LAGOON_VERSION}:1"
+if version_check $VERSION 1.8.2 || [[ "${LAGOON_VERSION}" = "development" ]]; then
```
Schnitzel: I think if we change this to bash the `[[` needs to be changed as this is a bash feature.

images/commons/lagoon/entrypoints/999-readiness.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
Schnitzel: switch to `sh`

images/php/fpm/Dockerfile
```diff
@@ -18,7 +18,7 @@ ENV LAGOON_VERSION=$LAGOON_VERSION
 
 # Copy commons files
 COPY --from=commons /lagoon /lagoon
-COPY --from=commons /bin/fix-permissions /bin/ep /bin/docker-sleep /bin/
+COPY --from=commons /bin/fix-permissions /bin/ep /bin/docker-sleep /bin/entrypoint-readiness /bin/
```
Schnitzel: mhh we don't need this in the `php` image? could we add this just to the `php-cli` image?

Schnitzel (changes requested):
nice work! just a couple of small things

PR #2033, opened 2020-07-06
Use lagoon task system retrying instead of internal retrying

Schnitzel:
this will not block other builds from being tested and make the retrying more aligned to other services


rocketeerbkw:
This could cause builds to execute out of order, since the tasks system adds delayed messages at the end of the queue. It's practically a revert of cd5a06aa1ce01c8b92594a36810ccba4f43c4d0e.

I think the way to actually fix this is totally abandon the existing task code (for this) and use a queue for each project, so that retries can be added/tracked correctly. Or it could be easier to somehow "get all pending builds" and sort in JS every time a build is being checked?

rocketeerbkw:
Found an issue when trying to fetch pending deployment, so this also fixes #2050

PR #2039, opened 2020-07-07
Patch existing routes with disable_cookies upon deployment.

### LINKED ISSUE #2038: PATCH existing routes with disable_cookies upon deployment
**Describe the bug**

At the moment, if you are to deploy a project with Lagoon, and the routes in OpenShift already exist, then it does not apply some common sense defaults. One of the primary issues faced on older clusters, is that some of the routes were created manually (not with Lagoon), and because of that, that are not configured optimally for Lagoon.

One such annotation is that is hugely beneficial is:

```
Annotations:		haproxy.router.openshift.io/disable_cookies=true
```

And often this is missing on older projects.

**To Reproduce**
1. Deploy a project with Lagoon with no custom domains
2. Manually create a route in OpenShift, missing the `disable_cookies` annotation
3. Deploy the project, this time with the domain added to `.lagoon.yml`
4. Notice how the existing route does not get updated.

**Expected behavior**

Lagoon would PATCH existing resources that is under it's control to ensure that best practice is always adhered too.

seanhamlin:
# Checklist

Patches all existing routes to remove the OpenShift cookies upon the next deployment. This will break any site that relies on this cookie being present (e.g. an HA deployment, that expects to only hit a single pod).

# Closing issues
Closes #2038


Schnitzel:
technically all routes should be already created with `haproxy.router.openshift.io/disable_cookies: 'true'` as this is fixed in the templates and there is no user definable setting to unset this. So we only oc patch the settings that the users actually can change.
plus usually we don’t have customer that have access to openshift UI and create routes themselves
so I guess this is a safe addition to Lagoon, to prevent users from shooting themselves in the foot

PR #2047, opened 2020-07-08
Only run auto-idler service in production Lagoon

### LINKED ISSUE #2044: make sure service idlers don't run unless the lagoon-env is set to production
**Describe the bug**
We had the issue that service idlers were running on a development lagoon and idled production environments

- Service idlers should check on the lagoon-env if the environment is production or development
- Service idlers only run on production environments

e.g. https://github.com/amazeeio/lagoon/blob/master/services/auto-idler/openshift-clis.sh

This will not be needed anymore as soon as @shreddedbacon can roll out the new improved service idler that runs on a per-cluster base instead of the centralized version of the idler we have currently

shreddedbacon:
# Checklist

Idling is not required in development lagoons, so we should disable it.

# Closing issues
closes #2044

### DIFF
--- a/services/auto-idler/idle-clis.sh
+++ b/services/auto-idler/idle-clis.sh
@@ -2,6 +2,8 @@
 
 # set -e -o pipefail
 
+if [ "${LAGOON_ENVIRONMENT_TYPE}" == "production" ]; then
+
 prefixwith() {
   local prefix="$1"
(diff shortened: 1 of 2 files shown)


shreddedbacon:
@Schnitzel can this be merged into the develop branch?

PR #2073, opened 2020-07-14
Add Lagoon labels to OpenShift projects/namespaces

### LINKED ISSUE #1969: Openshift namespace labels aren't applied
**Describe the bug**
Namespaces in OpenShift aren't labelled with the `lagoon.sh/...` labels on project creation.

**To Reproduce**
1. Deploy a project in OpenShift.
2. Inspect the namespace via the k8s API.
3. Observe no `lagoon.sh/...` labels.

**Expected behavior**
`lagoon.sh/...` labels should be added to the namespace.

**Screenshots**
n/a

**Additional context**
Labels on OpenShift namespaces were added in https://github.com/amazeeio/lagoon/pull/1917, but it seems that OpenShift [doesn't support labels on `ProjectRequest` objects](https://github.com/openshift/origin/issues/3819). Instead we will probably have to do this in two steps:
1. Create the project.
2. Patch the namespace.

These labels are important to the Lagoon logging system since they are used to determine project name / type.

smlx:
# Checklist

This PR fixes project/namespace labelling for OpenShift.

# Closing issues
Closes #1969 

Thanks @shreddedbacon for the tip on how to fix this :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -651,8 +651,12 @@ local-registry-up: build/local-registry
 openshift-run-api-tests = $(foreach image,$(api-tests),openshift-tests/$(image))
 .PHONY: $(openshift-run-api-tests)
 $(openshift-run-api-tests): minishift build/oc-build-deploy-dind openshift-test-services-up push-minishift
-		$(eval testname = $(subst openshift-tests/,,$@))
-		IMAGE_REPO=$(CI_BUILD_TAG) docker-compose -p $(CI_BUILD_TAG) --compatibility run --rm tests-openshift ansible-playbook /ansible/tests/$(testname).yaml
+	$(eval testname = $(subst openshift-tests/,,$@))
+	IMAGE_REPO=$(CI_BUILD_TAG) docker-compose -p $(CI_BUILD_TAG) \
+		--compatibility run --rm tests-openshift ansible-playbook \
+		/ansible/tests/$(testname).yaml \
+		--extra-vars \
+		"$$(./local-dev/minishift/minishift --profile $(CI_BUILD_TAG) ssh -- cat /home/docker/.kube/config | jq -rcsR '{kubeconfig: .}')"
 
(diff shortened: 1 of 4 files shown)


Schnitzel:
I think this might not work, as we give the role that lagoon uses to talk to the openshift only permissions to create projectrequests:
https://github.com/amazeeio/lagoon/blob/a5d6d839c62ee1d0dde530342fdeed104ee75c82/openshift-setup/clusterrole-openshiftbuilddeploy.yaml#L17-L24
and only read for projects:
https://github.com/amazeeio/lagoon/blob/a5d6d839c62ee1d0dde530342fdeed104ee75c82/openshift-setup/clusterrole-openshiftbuilddeploy.yaml#L9-L16

we can certainly change this (give access to the role to create projects directly) but then we need to be sure that we mention that in the release notes.

Unfortunatly this is not catched by the CI system, as minishift every useraccount has full access to everything

smlx:
I looked into the permissions a bit more and I think this approach (creating the `projects` directly) is not going to work.

When you create a `projectrequest`, OpenShift automatically adds an `admin` rolebinding in the new namespace for the user that submitted the request.

However when you create a `project` directly the project is created correctly but the `openshiftdeploy` user is _not_ given a rolebinding in the namespace. That means that they cannot interact any further with the namespace. I _could_ give the `openshiftdeploy` user an extra `clusterrolebinding` to enable manual `rolebinding` manipulation but that just starts getting messy.

Going back to the current method of creating a `projectrequest` and then patching the resulting namespace to add labels seems like a better option. However as discussed elsewhere the library we are currently using doesn't support patching a top-level namespace object. :expressionless: 

In conclusion I think that adding namespace labelling for openshift is probably going to be too difficult at this time, so I'll close this PR. Hopefully other ongoing work on the Lagoon deployment process will fix this issue anyway.

PR #2095, opened 2020-07-23
Support for Remote Lagoon build and task controllers

### LINKED ISSUE #2076: [LAGOON ENHANCEMENT PROPOSAL] Implement LagoonBuildDeploy Operator for Kubernetes
Proposal developed by @shreddedbacon 

We propose removing these services from Lagoon and turning them into an operator or controllers.
* `kubernetesbuilddeploymonitor` 
* `kubernetesdeployqueue`
* (existing `openshiftbuilddeploymonitor` may be out of scope for conversion)

Decoupling these services from Lagoon Core will mean that Lagoon Core operates leaner, handing over tasks to controllers that are better suited for performing these tasks, operating on the same clsuters as the tasks.

This would leverage a more native Kubernetes way of interacting with resources. The controllers can watch for changes in resources and perform actions based on these changes. This is far better than how we currently do it using `builddeploymonitor` and reaching out/polling for information and potentially losing connection with the build we were monitoring if the service restarts.

The proposed `LagoonBuildDeploy` operator would run in the destination cluster as part of a Lagoon Remote installation, and it would handle the creation and monitoring of builds.

The operator would be comprised of 2 controllers:
* `LagoonDeploy` - Creation of  the `Kind: Job` or `Kind: Build` objects for Lagoon, using information from a `LagoonBuild` CRD.
* `LagoonMonitor` - Monitors/watches the created `Kind: Job` or `Kind: Build` for status changes and updates the `LagoonBuild` CRD and/or Lagoon API

Lagoon services `(kubernetes|openshift)builddeploy` would now just create a `LagoonBuild` CRD in the destination cluster.

This `LagoonBuild` CRD would contain all the information required to kick off a build. 
Ideally, there should be no requirement for the controller to talk back to the Lagoon API to get any additional information once it receives this CRD. This would reduce the reliance on the API during maintenance when starting a deployment.

The `LagoonDeploy` controller would only mark one `LagoonBuild` CRD as active before it creates the Build/Job, and would only process another one once the active one is finished, making `kubernetesdeployqueue` not required. These would be the only tasks this controller performs.

The `LagoonMonitor` controller would watch for job/build creation. Then it would be responsible for creating the deployment object in the Lagoon API, watching the job, and ultimately updating the associated `LagoonBuild` CRD with the status and log information from the job before updating the Lagoon API.  This controller could potentially undertake more monitor/update tasks down the track.

The workflow would be something like this:
1. User/Webhook triggers a build for an environment
2. Lagoon creates `LagoonBuild` CRD in the destination cluster .
3. `LagoonDeploy` controller sees this CRD get created, creates the namespace if it doesn’t exist yet and can check if there are any running jobs already in the namespace and will only process one CRD at a time.
4. `LagoonMonitor` controller would watch the created builds/jobs and store the resulting status and build/job log in the `LagoonBuild` CRD associated with it.
5. If the status changes, we can POST to Lagoon updating incrementally.
6. `LagoonMonitor` controller would attempt POSTing the resulting data from the `LagoonBuild` CRD to the Lagoon API directly
7. On a successful POST, the `LagoonBuild` CRD may be removed (or the last `n` kept).
8. `LagoonDeploy` would then process the next CRD if one was pending.

# Considerations
This does not necessarily need to replace the proposed services, but could be created as new `activeSystemsX` targets with associated services to operate it. This way it could operate alongside the existing services initially.

To do this would mean creating a new service in lagoon in the same style as `kubernetesbuilddeploy` or `openshiftbuilddeploy` that would create the CRD in the destination cluster for the operator/controllers to handle.

Benefits of this is that only the `activeSystemsDeploy` target needs to be created, as the existing `activeSystemsTask/Misc/Remove` for kubernetes/openshift would be able to be used.

# Notes
Cancellation of builds/jobs could still be performed by the `(kubernetes|openshift)remove` services as normal I believe, as everything else Lagoon would do remains the same.
The operator/controllers have no direct contact with RabbitMQ/broker in Lagoon. Options are to talk directly to the API or post to the webhook handler, and the webhook handler deals with it

shreddedbacon:
# Checklist

Adds initial support for the Remote Lagoon build and task controllers.
Adding new activeSystemsX types:
- lagoon_controllerBuildDeploy
- lagoon_controllerRemove
- lagoon_controllerJobs
- lagoon_controllerMisc

A new service called `controllerhandler` handles the messages that come back from any operations that send messages to the `lagoon-tasks:controller` queue and transforms these into LagoonAPI requests.

Install the Remote Lagoon build and task controllers https://github.com/amazeeio/lagoon-kbd via helm with configuration to talk the Lagoon broker.

#### Before
This is a rough diagram of how the current build deploy system works
image

#### After
And this is how it is when using the operator and associated components.
image

### Work still to do
- [x] Write and add helm chart to install controllers
- [x] Write tests that use the controllers

### Extras
- [x] Integrate the Jobs and Misc tasks that would run in Kubernetes into the controllers too
- [ ] Support openshift (bonus points)

# Closing issues
Partially closes #2076


shreddedbacon:
I changed `xip.io` entries to `nip.io` as I was getting DNS issues intermittently using `xip.io`. 
No issues at all using `nip.io`.

PR #2110, opened 2020-07-27
Update Nginx configuration for Drupal, to bring it inline with best practice. Additional tests too.

### LINKED ISSUE #237: Redirect /index.php/* to /* in Drupal Nginx images
See https://www.drupal.org/forum/support/post-installation/2020-04-06/solved-drupal-8-is-adding-indexphp-in-url for background. In short, Drupal 8 has valid routes for `/index.php/blah` where the friendly URL is `/blah`. This only serves to confuse Google, and the end user.

There is best practice in https://www.nginx.com/resources/wiki/start/topics/recipes/drupal/ to fix this with

```
    # Enforce clean URLs
    # Removes index.php from urls like www.example.com/index.php/my-page --> www.example.com/my-page
    # Could be done with 301 for permanent or other redirect codes.
    if ($request_uri ~* "^(.*/)index\.php/(.*)") {
        return 307 $1$2;
    }
```

We should add this to the base images.

### LINKED ISSUE #2109: Redirect /index.php/* to /* in Drupal Nginx images
See https://www.drupal.org/forum/support/post-installation/2020-04-06/solved-drupal-8-is-adding-indexphp-in-url for background. In short, Drupal 8 has valid routes for `/index.php/blah` where the friendly URL is `/blah`.

There is best practice in https://www.nginx.com/resources/wiki/start/topics/recipes/drupal/ to fix this with

```
    # Enforce clean URLs
    # Removes index.php from urls like www.example.com/index.php/my-page --> www.example.com/my-page
    # Could be done with 301 for permanent or other redirect codes.
    if ($request_uri ~* "^(.*/)index\.php/(.*)") {
        return 307 $1$2;
    }
```

We should add this to the base images.

seanhamlin:
# Checklist

Updates the Nginx configuration for Drupal based sites to bring it closer to https://www.nginx.com/resources/wiki/start/topics/recipes/drupal/

# Closing issues
Closes #2109

### DIFF
--- a/images/nginx-drupal/drupal.conf
+++ b/images/nginx-drupal/drupal.conf
@@ -1,4 +1,4 @@
-### Nginx configuration for Drupal 7 and 8.
+## Nginx configuration for Drupal 7 and 8.
 server {
   include /etc/nginx/conf.d/drupal/server_prepend*.conf;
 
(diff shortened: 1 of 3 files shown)


images/nginx-drupal/drupal.conf
```diff
@@ -7,50 +7,61 @@ server {
...
+    ## Allow "Well-Known URIs" as per RFC 5785
+    location ~* ^/.well-known/ {
+      allow all;
+    }
+
+    ## Block access to "hidden" files and directories whose names begin with a
+    ## period. This includes directories used by version control systems such
+    ## as Subversion or Git to store control files.
+    location ~ (^|/)\. {
+      return 403;
```
smlx, line 46: should this be 404 to match the other directives above?

smlx (approved):
minor query but LGTM and tests++ :rocket:

images/nginx-drupal/drupal.conf
seanhamlin: Potentially, I was on the fence with this one.

images/nginx-drupal/drupal.conf
```diff
@@ -90,34 +101,39 @@ server {
     deny all;
     access_log off;
     log_not_found off;
+    return 404;
```
Schnitzel, line 104: why do we need a specific 404 here? the `deny all` already returns with an `403`? see http://nginx.master.drupal-example-9.test1.amazee.io/patches/asdf

images/nginx-drupal/drupal.conf
```diff
@@ -7,50 +7,61 @@ server {
...
       log_not_found off;
+      return 404;
     }
 
     ## Expiring per default for four weeks and one second, Drupal will overwrite that if necessary
     expires ${NGINX_DEFAULT_EXPIRES:-2628001s};
 
-    ## Disallow access to any dot files, but send the request to Drupal
-    location ~* /\. {
-      try_files /dev/null @drupal;
```
Schnitzel, line 41: removing this will cause issues. Drupal allows URLs that start with a dot: http://nginx.master.drupal-example-9.test1.amazee.io/foobar/.dotfilesforethewin
your proposed regex: `location ~ (^|/)\. {` would match this and block traffic to this.

Schnitzel (changes requested):

tobybellwood:
Can we get the nginx-image parts of this split into a uselagoon/lagoon-images PR, and then the updated tests against the main branch?

tobybellwood:
@seanhamlin do we still want to implement these changes to the drupal conf - if so we'll need to move this across to lagoon-images?

seanhamlin:
Yes this PR is still relevant, but likely will require refactoring for the uselagoon repos.

PR #2119, opened 2020-07-29
Added helper script to delete unused .kibana indexes

### LINKED ISSUE #2118: Helpers script to remove old .kibana indexes
Kibana every time it restarts, creates a new `.kibana_` index and update the alias.
For example the index `.kibana_4` is the index which alias is `.kibana`.

Those indexes are used to keep the kibana configuration and are kept in case of a configuration rollback/restore.
It might be the case to keep just the latest index and delete the old ones.

vincenzodnp:
# Checklist

The aim of the script is to remove old `.kibana_*` indexes and just keep the latest

# Closing issues
closes #2118

### DIFF
--- a/helpers/clean-kibana-indexes.sh
+++ b/helpers/clean-kibana-indexes.sh
@@ -39,6 +39,14 @@ DRYRUN="${DRYRUN:-"false"}"
 
 # Optional set a tenant to check and close/delete indexes
 TENANT="${TENANT:-"*"}"
+if [ "$TENANT" = "*" ]; then
+	REGEX="*"
+# This is a special tenant that requires a special REGEX
+elif [ "$TENANT" = "kibana" ]; then
+	REGEX=""
+else
+	REGEX="_*_$TENANT"
+fi
 
 # Switch context to the selected cluster
 echo -e "You're switching context to ${CLUSTER}"
@@ -63,7 +71,7 @@ cat << EOF |
 IFS=$'\n'
 
 # Generate array with kibana aliases for a specific tenant if specified, otherwise get all aliases
-ELASTICSEARCH_ALIASES=(\$(es-curl GET "_cat/aliases/.kibana_*_$TENANT?h=alias,index"))
+ELASTICSEARCH_ALIASES=(\$(es-curl GET "_cat/aliases/.kibana$REGEX?h=alias,index"))
 
 # For each row, get current alias and index
 for ALIAS_ROW in "\${ELASTICSEARCH_ALIASES[@]}"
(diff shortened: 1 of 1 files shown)


helpers/clean-kibana-indexes.sh
```diff
@@ -0,0 +1,114 @@
...
+	exit 1
+fi
+
+# Set DRYRUN option to "true" to run in a dry-run mode
+DRYRUN="${DRYRUN:-"false"}"
+
+# Optional set a tenant to check and close/delete indexes
+TENANT="${TENANT:-"*"}"
+if [ "$TENANT" = "*" ]; then
+	REGEX="*"
```
Schnitzel: this is not a regex, let's call this what this actually is

helpers/clean-kibana-indexes.sh
```diff
@@ -0,0 +1,114 @@
...
+# For each row, get current alias and index
+for ALIAS_ROW in "\${ELASTICSEARCH_ALIASES[@]}"
+do
+	ALIAS=\$(echo \$ALIAS_ROW|awk '{print \$1}')
+	INDEX=\$(echo \$ALIAS_ROW|awk '{print \$2}')
+	echo "\$ALIAS is the alias for \$INDEX"
+	
+	# Generate an arry with old indexes to remove
+	if [ "\$ALIAS" = ".kibana" ]; then
+		# This fix avoids the "catch-all" .kibana* regex
```
Schnitzel, line 85: you mention to avoid the `"catch-all" .kibana*` but the next line exactly uses that request, this is quite confusing, can you explain better what exactly happens here?

helpers/clean-kibana-indexes.sh
```diff
@@ -0,0 +1,114 @@
...
+set -eu
+
+# Set DEBUG variable to true, to start bash in debug mode
+DEBUG="${DEBUG:-"false"}"
+if [ "$DEBUG" = "true" ]; then
+        set -x
+fi
+
+# Cluster to connect to
+CLUSTER="${CLUSTER:-""}"
```
Schnitzel, line 22: `CLUSTER` is quite confusing here, the cluster of what? where the tenant has their projects? or where the elasticsearch is located? let's call this something better

helpers/clean-kibana-indexes.sh
```diff
@@ -0,0 +1,114 @@
...
+ELASTICSEARCH_POD="${ELASTICSEARCH_POD:-$(kubectl -n "$ELASTICSEARCH_NAMESPACE" get pods --no-headers -o custom-columns=NAME:.metadata.name|head -n1)}"
+echo -e "Selected POD to run commands is $ELASTICSEARCH_POD"
+
+# Generate script dynamically and pipe to the selected POD
+cat << EOF |
+#/bin/bash
+IFS=$'\n'
+
+# Generate array with kibana aliases for a specific tenant if specified, otherwise get all aliases
+ELASTICSEARCH_ALIASES=(\$(es-curl GET "_cat/aliases/.kibana$REGEX?h=alias,index"))
```
Schnitzel: don't you need to escape the `$` here? we do it everywhere else

helpers/clean-kibana-indexes.sh
```diff
@@ -0,0 +1,114 @@
...
+for ALIAS_ROW in "\${ELASTICSEARCH_ALIASES[@]}"
+do
+	ALIAS=\$(echo \$ALIAS_ROW|awk '{print \$1}')
+	INDEX=\$(echo \$ALIAS_ROW|awk '{print \$2}')
+	echo "\$ALIAS is the alias for \$INDEX"
+	
+	# Generate an arry with old indexes to remove
+	if [ "\$ALIAS" = ".kibana" ]; then
+		# This fix avoids the "catch-all" .kibana* regex
+		OLD_INDEXES=(\$(es-curl GET "_cat/indices/.kibana*?h=index" -s |grep -E '^.kibana_?[0-9]$'| grep -v -w \$INDEX))
```
Schnitzel, line 86: can you document how this is different from the other request which fills `OLD_INDEXES`?
In the lower one you excluding the current index via `grep -v -w $INDEX`, but here you match only indexes that match the regex `^.kibana_?[0-9]$` how do you make sure though that the active index is not removed/closed?

Schnitzel (changes requested):
couple of questions and improvements

helpers/clean-kibana-indexes.sh
vincenzodnp: This is how I can define a variable outside the script generated by EOF and pass it to it.
The escaped `$` is because defined variables are used inside the script itself, while REGEX, for example, is defined outside.

helpers/clean-kibana-indexes.sh
vincenzodnp: I called INDEXES

helpers/clean-kibana-indexes.sh
vincenzodnp: So the full snippet is
```
# Generate an arry with old indexes to remove
 84         if [ "\$ALIAS" = ".kibana" ]; then
 85                 # This fix avoids the "catch-all" .kibana* regex
 86                 OLD_INDEXES=(\$(es-curl GET "_cat/indices/.kibana*?h=index" -s |grep -E '^.kibana_?[0-9]$'| grep -v -w \$INDEX))
 87         else
 88                 OLD_INDEXES=(\$(es-curl GET "_cat/indices/\$ALIAS*?h=index" -s |grep -v -w \$INDEX))
 89         fi
```
So for each row of the `ELASTICSEARCH_ALIASES` array, I get the ALIAS and the INDEX.
When we are going to close the `.kibana_*` indexes related to `.kibana` alias, I needed to find a way to avoid the but we had that `.kibana*` wildcard regex, matches **all** the indexes.
So I added this IF statement for the particular case of `.kibana` alias plus the `grep -E` part that does the trick.

helpers/clean-kibana-indexes.sh
vincenzodnp: Kubernetes/Openshift cluster where ES is located.
I changed "Target Kubernetes/Openshift cluster where Elasticsearch is located"
I use it to switch kubectl context

helpers/clean-kibana-indexes.sh
vincenzodnp: The GET gets all .kibana indexes (`*`). From that output I get all the indexes I know are related to `.kibana` alias.
The structure of an index is `.kibana_randomnumber_tenant_progressive-number`. In the case of `.kibana` indexes, they are `.kibana_1`, `.kibana_2`, etc, etc. So with the regular expression of grep, from all the indexes I get only the one related to .kibana alias. From this filtered output, I exclude the current index that is in `INDEX` variable.

vincenzodnp:
@Schnitzel any news about this PR?

helpers/clean-kibana-indexes.sh
Schnitzel: ok, but can you update the comment in the code to make this clearer how exactly this works? right now it looks like a bug.

helpers/clean-kibana-indexes.sh
Schnitzel: please change the actual variable to make it more clearer, not just the comment.

helpers/clean-kibana-indexes.sh
Schnitzel: please document, not comment in here how it works.

smlx:
Hi @vincenzodnp these indices store kibana configuration and I'm not sure if we can just delete them. How do we know that they are unused?

tobybellwood:
should be resubmitted against https://github.com/amazeeio/helper-scripts

PR #2123, opened 2020-07-30
Lagoon 2.0 - API : Removing all Mutations for deleteAll*

justinlevi:
# Checklist

Removing some scary API mutations as they were only being used during CI and local builds. 

# Closing issues
https://github.com/amazeeio/lagoon/issues/2101


local-dev/api-data-watcher-pusher/api-watch-push.sh
```diff
@@ -42,13 +41,7 @@ watch_apidatafolder() {
...
             echo "******* Found changes in gql files in /api-data/, clearing and re-populating"
 
-            if
-                send_graphql_query $clear_gql_file_path;
-            then
-                chsum_clear_prev=$chsum_clear_curr
-            else
-                echo '**** ERROR while clearing, will try again.'
-            fi
+            clear_data=$(/home/clear_api_data.py)
```
rocketeerbkw: I didn't see any messages from this script while testing so I think this method consumes any error messages from the python script, so maybe follow with a `echo $clear_data`. Or maybe there's a way to call the script differently that's allowed to print to stdout?

Also it doesn't look like the keycloak clearing script gets called anywhere? Would that go here?

local-dev/api-data-watcher-pusher/clear_api_data.py
```diff
@@ -0,0 +1,47 @@
...
+notification_slack = "TRUNCATE TABLE notification_slack"
+notification_email = "TRUNCATE TABLE notification_email"
+notification_rocketchat = "TRUNCATE TABLE notification_rocketchat"
+notification_microsoftteams = "TRUNCATE TABLE notification_microsoftteams"
+project_notification = "TRUNCATE TABLE project_notification"
+openshift = "TRUNCATE TABLE openshift"
+project = "TRUNCATE TABLE project"
+ssh_key = "TRUNCATE TABLE ssh_key"
+user_ssh_key = "TRUNCATE TABLE user_ssh_key"
+environment_backup = "TRUNCATE TABLE environment_backup"
```
rocketeerbkw: It looks like there are some duplicate key errors for the backup restores and environment variables still. Let's just truncate all api tables here.

rocketeerbkw (changes requested):

local-dev/api-data-watcher-pusher/Dockerfile
```diff
@@ -1,10 +1,11 @@
...
+FROM python:3-alpine3.11
 
-RUN apk add --no-cache tini bash wget py3-jwt
+RUN apk add --no-cache tini bash wget
+RUN pip install mysql-connector python-keycloak PyJWT
 
 ENV JWTSECRET=super-secret-string \
     JWTAUDIENCE=api.dev
 
 COPY api-watch-push.sh create_jwt.py /home/
```
rocketeerbkw: Sorry, forgot this earlier. I would also add the two new python scripts here.

It might not be strictly necessary, unless there's a use case I don't know of or can't remember? This image only gets run locally or on jenkins, and in both cases I'm pretty sure it uses docker compose, which will mount the local files into the container when it runs (so the scripts will be there).

rocketeerbkw (changes requested):

local-dev/api-data-watcher-pusher/api-watch-push.sh
justinlevi: This should be resolved in 5ae039d239e68beec099fb8041c8712e7dfa8444

local-dev/api-data-watcher-pusher/clear_api_data.py
justinlevi: Should be resolved in 61c790aed43c5f48cd644be3014b31a3724288f2

local-dev/api-data-watcher-pusher/Dockerfile
justinlevi: Should be resolved in 70dd1e7ef50c945be06552e2cbdb5acf8a5a7e9f

rocketeerbkw (approved):

tobybellwood:
would require more work to update than refactor - closing here

PR #2127, opened 2020-07-31
Cache lagoon keycloak realm in CI

smlx:
# Checklist

This reduces keycloak startup time for local testing. On my laptop it goes from 246 seconds to 46 seconds.

# Closing issues
n/a

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -335,6 +335,10 @@ services:
       lagoon.type: custom
       lagoon.template: services/keycloak/.lagoon.app.yml
       lagoon.image: amazeeiolagoon/keycloak:v1-8-2
+    environment:
+      - USE_CACHED_LAGOON_REALM_CI=true
+    volumes:
+      - ./services/keycloak/lagoon-realm-ci.json:/lagoon-realm-ci.json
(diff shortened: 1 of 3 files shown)


Schnitzel:
@smlx 
nice! can you document how we need to update the exported realm in the codebase and when we need to update? Or is that something that is not necessary?

smlx:
hi @Schnitzel I documented in a comment here: https://github.com/amazeeio/lagoon/pull/2127/files#diff-8c9af22e910ccf7cd25a29a92c5faadaR1710-R1714

But appreciate that it might be hard to find, so if there's somewhere better that I could put instructions then please let me know.

And yes, this cached configuration would need to be updated if the keycloak `start.sh` was changed.

Schnitzel:
mhh I guess the most superb version would be:
- use the cached version during local testing
- in CI testing don't use the cached version and run a script that compares the outcome with the actual stored cache in the repo, if there is a difference fail the test.

smlx:
Since I'm doing all testing using the helm charts now, this PR isn't very useful for me. I'll close it but feel free to reopen if it could still be useful.

PR #2128, opened 2020-08-03
Remove duplicate Python versions from images and use it in our tests

tobybellwood:
This PR removes a duplicate Python install from the base images - in some cases the versions were different.

I've also updated the ckan image for the correct libmagic dependency.

This PR also uses the amazeeio python 2.7 image to power the ansible testing.  We can move to 3.7 with some modifications.  We will also need to change our "filters as tests" to upgrade to Ansible 2.9

# Checklist


smlx:
LGTM. I guess we need a follow-up issue to overhaul tests for ansible 2.9 and python 3?

tobybellwood:
yes - shouldn't be too hard IMO

smlx (approved):

PR #2130, opened 2020-08-04
Fix CiActiveStandby gitUrl for local-dev

tobybellwood:
The gitUrl for CiActiveStandby is set to 192.168.42.1 instead of 172.17.0.1 - this causes issues for Linux local-dev - or perhaps because I don't have the full setup for minishift too.  Changing it to match the others fixes the issue.

# Checklist


smlx (approved):
LGTM

PR #2148, opened 2020-08-13
Add random suffix to the bucket name in k8up schedules

### LINKED ISSUE #186: Backup schedules define common bucket names

smlx:
# Checklist

Fixes s3 bucket name collisions.

# Closing issues
Closes: uselagoon/build-deploy-tool#186

### DIFF
--- a/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
+++ b/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
@@ -830,9 +830,9 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
   PRUNE_SCHEDULE=$( /kubectl-build-deploy/scripts/convert-crontab.sh "${NAMESPACE}" "M H(3-6) * * 6")
   TEMPLATE_PARAMETERS+=(-p PRUNE_SCHEDULE="${PRUNE_SCHEDULE}")
 
-  OPENSHIFT_TEMPLATE="/kubectl-build-deploy/openshift-templates/backup-schedule.yml"
   helm template k8up-lagoon-backup-schedule /kubectl-build-deploy/helmcharts/k8up-schedule \
     -f /kubectl-build-deploy/values.yaml \
+    --set projectSecret="$PROJECT_SECRET" \
     --set backup.schedule="${BACKUP_SCHEDULE}" \
     --set check.schedule="${CHECK_SCHEDULE}" \
     --set prune.schedule="${PRUNE_SCHEDULE}" "${HELM_ARGUMENTS[@]}"  > $YAML_FOLDER/k8up-lagoon-backup-schedule.yaml
--- a/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
+++ b/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
@@ -830,9 +830,11 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
   PRUNE_SCHEDULE=$( /kubectl-build-deploy/scripts/convert-crontab.sh "${NAMESPACE}" "M H(3-6) * * 6")
   TEMPLATE_PARAMETERS+=(-p PRUNE_SCHEDULE="${PRUNE_SCHEDULE}")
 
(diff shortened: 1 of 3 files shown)


smlx:
I have no idea how to fix this in OpenShift :shrug:

Schnitzel:
two questions/issues:
- we can only do this for cases where the bucket doesn't already exist or we would immediately create a second bucket
- wouldn't `randAlphaNum` create a new random string every time we run the helmchart? aka it would create a new bucket name on every deployment? :D

smlx:
Thanks for taking a look @Schnitzel!

Both those points are valid. I assumed that the `schedule` was only created on initial deployment, but looking at the code I see that it is actually re-created on every deploy. Why is that? And could we switch it to check for existing an existing `schedule` object before re-creating it?

tobybellwood:
(some discussion about RandAlphaNum recreating every time - seems a known Helm issue - https://github.com/helm/helm/issues/3053)

Do we have a unique name for the project (or identifier for the cluster) that we could use to serve as a permanent suffix (or as a value to add to a hash?

Schnitzel:
> Do we have a unique name for the project (or identifier for the cluster) that we could use to serve as a permanent suffix (or as a value to add to a hash?

I wouldn't use cluster identifier, as projects technically can be in multiple clusters or even move cluster.

Technically the project name is already unique within lagoon, so I'm not exactly sure what we try to fix here? if there are duplicated bucketnames generated we might need to figure out why they are duplicated?

We could add a new field within lagoon that defined the bucketname that should be used, if this field is not set, Lagoon uses the current pattern (`baas-[projectname]`) but if the bucketname is set we could use the defined bucketname. This also would allow to have lagoon set a random suffix to the bucketname for each project that is added to Lagoon, while keeping the current bucketnames existing (so they don't break their backups)?

smlx:
> Technically the project name is already unique within lagoon, so I'm not exactly sure what we try to fix here? if there are duplicated bucketnames generated we might need to figure out why they are duplicated?

There are multiple e.g. `drupal-example` projects across different Lagoon installations which then try to push to a bucket named `baas-drupal-example`.

smlx:
Ok, I added handling for existing buckets / schedule objects.

Schnitzel:
unfortunately this will now break another thing:
we rely on the fact that multiple environments use the same bucketname (to have deduplication across environments).
but each new environment will not create their own bucketname as the `Schedule` does not exist yet.
I think we therefore would need to has something that is the same across multiple projects (the projectname itself wouldn't really work, as this would generate the same hash across different lagoon installations)
but we could use https://github.com/amazeeio/lagoon/blob/master/services/kubernetesbuilddeploy/src/index.ts#L90 which is available via `PROJECT_SECRET` and is created based on the JWTtoken of the lagoon installation (and hopefully is separate between different lagoon installations)

smlx:
Hey @Schnitzel I updated this PR to use the `PROJECT_SECRET` as you described. That's a much simpler change too :smile:

Schnitzel (approved):

Schnitzel:
after thinking about this a bit more I think this still does not cover all cases:
We expect that each environment of a project use the same backup bucket. In this current implementation a new environment will not find a `k8up-lagoon-backup-schedule` object and therefore create a new `baas-*` bucket with the random string, while all other environments could use a predefined bucket.

smlx:
> We expect that each environment of a project use the same backup bucket. In this current implementation a new environment will not find a k8up-lagoon-backup-schedule object and therefore create a new baas-* bucket with the random string, while all other environments could use a predefined bucket.

~Hi @Schnitzel, so the string isn't really random. Instead it's more like a hash based on the project secret. So the hash will be different per project, but shared between environments of the same project.~

Edit: I just realised that the project secret is currently generated with only the JWTSECRET and project name as input. So I added the project ID and deploy target to the input so that it will change in a migrated project.

Edit 2: sorry I just re-read what you wrote and actually understood what you are saying. :facepalm: 
Yes, this is a problem and I don't think there's a way to avoid this. So existing projects may end up with two buckets. What kinds of issues could this cause?

smlx:
I just wanted to note that this PR now changes the way that the `PROJECT_SECRET` is calculated. I don't know if that is going to cause problems? @Schnitzel maybe you could advise?

Also, I believe that https://github.com/amazeeio/lagoon/pull/2355 solves the same basic issue of bucket name collisions. So if that PR is preferred then please close this one.

shreddedbacon:
I think that having a random bucketname generation with the ability to override if required could actually be good, so both PRs as one?

tobybellwood:
to be re-addressed in build-deploy-tool

PR #2149, opened 2020-08-13
Ignore the watchdog table for backups

### LINKED ISSUE #2142: Prebackuppod command fails for enormous tables
**Describe the bug**
Some prebackuppod jobs are failing with this error:

```
mysqldump: Error 3024: Query execution was interrupted, maximum statement execution time exceeded when dumping table `watchdog` at row: 28850830
```

**To Reproduce**
n/a

**Expected behavior**
Backups run successfully.

**Screenshots**
n/a

**Additional context**
Maybe we can exclude `watchdog` from the backup by default? It shouldn't contain anything useful when using `lagoon_logs` anyway, right?

smlx:
# Checklist

Backups can fail for big tables. The table can be enormous and for lagoon will be unused since project will be using the `lagoon_logs` module. Not sure if this is really the best approach though, since `watchdog` might be used for something else in a non-drupal project.

# Closing issues
Closes: #2142

### DIFF
--- a/images/kubectl-build-deploy-dind/helmcharts/mariadb-dbaas/templates/prebackuppod.yaml
+++ b/images/kubectl-build-deploy-dind/helmcharts/mariadb-dbaas/templates/prebackuppod.yaml
@@ -21,6 +21,7 @@ spec:
     && mysqldump --max-allowed-packet=500M --events --routines --quick
     --add-locks --no-autocommit --single-transaction --no-create-db
     --ignore-table=$BACKUP_DB_DATABASE.watchdog
+    --no-create-info
     -h $BACKUP_DB_HOST
     -u $BACKUP_DB_USERNAME
     -p$BACKUP_DB_PASSWORD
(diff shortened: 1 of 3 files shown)


seanhamlin (changes requested):
I think removing tables from backups breaks some assumptions about the backup system (i.e. that you can restore to a backup to fully roll your site back).

If we are to proceed with this, we should look to dump the structure of all tables, and then selectively look to dump the content of all the tables, depending on name.

smlx:
So something like this @seanhamlin ?
```
mysqldump --no-data ... > dump.sql
mysqldump --ignore-table=DB.watchdog ... >> dump.sql
```

seanhamlin:
As long as the second does not dump the structure again, e.g. with `mysqldump --no-create-info`, looks good to me.

smlx:
okay, I've tested the commands locally by dumping a table this way, restoring it into a new DB, and then diffing dumps of the two DBs. As intended, table is created but has no data added.

seanhamlin (approved):
I can't see any valid reason in Lagoon when storing watchdog entries in the DB would be needed. Ideally the `dblog` module in Drupal would be disabled and Kibana used to search logs.

I especially like the multiline dump statements, which are far more readable in the first instance.

Schnitzel:
@smlx this is conflicting now, can you fix that one?

smlx:
@Schnitzel okay, conflicts fixed

PR #2153, opened 2020-08-13
Allow non-admin users to run more queries

rocketeerbkw:
# Checklist

The `allGroups`, `allProjectsInGroup`, and `groupByName` queries were previously checking for `group:viewAll` permissions which is typically granted to admins only. A bugfix in a previous release caused `allGroups -> projects` to return permission denied due to this check.

Now all of these queries will check for user level permissions and fixes the permissions regression.

# Closing issues
AMAZEEIO-3235


rocketeerbkw:
@justinlevi I think you added these resolvers to facilitate billing stuff, can you please double check this doesn't break anything?

wintercreative (approved):
All of the billing tests run as expected and I've manually checked this PR locally and all billing calculations show up as expected in the UI. I don't see anything in the code that would cause any issues so I think we're good to go.

PR #2156, opened 2020-08-17
Add Project/Environment EnvVar overrides for service types

shreddedbacon:
# Checklist

To migrate Openshift Ansible Service Broker `mariadb-shared` provisioned resources to use the newer `mariadb-dbaas`, we need a way to override what may be defined in the `docker-compose.yml` or `.lagoon.yml` files.

It was identified that some may have defined `mariadb-shared` in their files due to an old issue, or past advice to add it to move their sites to use the shared services.

This PR adds an EnvVar check at the project and environment levels to check for `LAGOON_SERVICE_TYPES` in the Lagoon API.

If defined, it will be used to set the service type override, beating what may be defined in the `docker-compose.yml` or `.lagoon.yml` files.

The format that this variable accepts is comma separated values, but each value is itself a colon separated value.

* Example 1: mariadb:mariadb-dbaas
  * This tells any docker-compose services named `mariadb` to use the `mariadb-dbaas` service type, even if the value is hardcoded in the `docker-compose.yml` or `.lagoon.yml` files to be something else.
* Example 2: mariadb:mariadb-dbaas,nginx:nginx-persistent

I'm undecided about putting this in public documentation due to the fact that there is potential to break deployments by changing this. Mainly for databases, other types may be ok.

I could document it with a big !!DANGER!! notice to discourage users from using it though unless they know what the outcome will be.

# Closing issues


rocketeerbkw:
> discourage users from using it

How discouraging should it be? If a user can add an environment variable then presumably they can just edit the `.lagoon.yml` directly, so should this be an admin only thing? It would add complexity but we could maybe add another RBAC scope for environment variables like `add:${name}` then you could potentially enforce that or other variables. Guess you'd need to do the same for `delete:${name}`.

shreddedbacon:
> > discourage users from using it
> 
> How discouraging should it be? If a user can add an environment variable then presumably they can just edit the `.lagoon.yml` directly, so should this be an admin only thing? It would add complexity but we could maybe add another RBAC scope for environment variables like `add:${name}` then you could potentially enforce that or other variables. Guess you'd need to do the same for `delete:${name}`.

Well they can already do this exact thing via docker-compose.yml or .lagoon.yml, this is just another way of doing it. Users could possibly break their own deployments right now by changing between two types.
We already make it known that it is possible to change the type in the docker-compose file and list the types that are available https://lagoon.readthedocs.io/en/latest/using_lagoon/docker-compose_yml/#types

I don't think this particular issue needs the additional RBAC complexity wrapped around it, but I do like the idea of admin only settable variables.

Schnitzel:
@shreddedbacon 
can we also get a test for this? I could imagine that we create a git repo which has the `lagoon.type: none` for an nginx container define, but then via the env variables we change it back to a `node` type and the tests check if the nginx is actually deployed?

shreddedbacon:
> @shreddedbacon
> can we also get a test for this? I could imagine that we create a git repo which has the `lagoon.type: none` for an nginx container define, but then via the env variables we change it back to a `node` type and the tests check if the nginx is actually deployed?

I didn't finish writing the tests today as I got distracted by other priorities. Will continue to write the tests, but may not make it before a release of 1.9, but it would be good to get this in for 1.9

PR #2160, opened 2020-08-20
Added support for injecting Matomo snippets via ngx_http_sub_module.

### LINKED ISSUE #2159: Add ability for nginx images to inject Matomo snippets for website tracking.
Matomo, formerly Piwik, is a free and open source web analytics application which tracks online visits to one or more websites and displays reports on these visits for analysis.

Rather than having to manage code snippets per project it would be nice to be able to inject code snippets into a page via environment variables. This would especially be useful for lagoon users with a large number of projects.

Using the `ngx_http_sub_module` capabilities of NGINX we could provide a the following environment variables :-

`MATOMO_URL` - URL to the Matomo isntallation
`MATOMO_SITE_ID` - Matomo site ID
`MATOMO_TAG_MANAGER_URL` - Optionally, providing support for Matomo tag manager.

Setting these variables on the NGINX container could automatically enable the snippets to be injected into the HEAD section of all HTML responses.

thom8:
Adds an entrypoint script to the base `nginx` image which will generate new NGINX configuration file that injects Matomo tracking snippets. There's 2 different options - a standard installation which requires a Matomo URL `MATOMO_URL` and Site ID `MATOMO_SITE_ID` the other option is providing a URL `MATOMO_TAG_MANAGER_URL` to a Tag manager script that will allow the use of Matomo Tag manager.

If no variables are set no config files are created and there's no affect to the current default behaviours.

# Checklist

# Closing issues
closes #2159

### DIFF
--- a/images/nginx/matomo/80-nginx-matomo-config
+++ b/images/nginx/matomo/80-nginx-matomo-config
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+# Inject Matomo tracking snippets via nginx configuration.
+
+# Single site configuration.
+if [ "$MATOMO_URL" ]; then
+
+    # Add Matomo tracking code to end of the HEAD section.
+    cat <<EOF >> /etc/nginx/conf.d/matomo.conf
+sub_filter '</head>'
+"<!-- Matomo -->
+<script type='text/javascript'>
+var _paq = window._paq = window._paq || [];
+/* tracker methods like 'setCustomDimension' should be called before 'trackPageView' */
+_paq.push(['trackPageView']);
+_paq.push(['enableLinkTracking']);
+(function() {
+    var u='${MATOMO_URL}';
+    _paq.push(['setTrackerUrl', u+'matomo.php']);
+    _paq.push(['setSiteId', '${MATOMO_SITE_ID:-1}']);
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.type='text/javascript'; g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
+})();
+</script>
+<!-- End Matomo Code -->
+</head>";
+sub_filter_once on;
+EOF
+
+fi
+
+# Tag manager configuration.
+if [ "$MATOMO_TAG_MANAGER_URL" ]; then
+
+    cat <<EOF >> /etc/nginx/conf.d/matomo.conf
+sub_filter '<head>'
+"<head>
+<!-- Matomo Tag Manager -->
+<script type='text/javascript'>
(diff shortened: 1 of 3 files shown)


images/nginx/matomo/80-nginx-matomo-config
```diff
@@ -0,0 +1,53 @@
...
+
+    fix-permissions /etc/nginx/conf.d/
+
+fi
+
+# Tag manager configuration.
+if [ ! -z "$MATOMO_TAG_MANAGER_URL" ]; then
+
+    cat <<EOF >> /etc/nginx/conf.d/matomo.conf
+sub_filter '$MATOMO_REPLACE_TAG'
```
tobybellwood: what's this variable for - shouldn't it work the same as the MATOMO_URL snippet and just append to `</head>`?

thom8: Yep reverted, the tag manager recommends adding to the top of the <head> section.

thom8: Thanks for spotting!

images/nginx/matomo/80-nginx-matomo-config
```diff
@@ -0,0 +1,53 @@
...
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.type='text/javascript'; g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
+})();
+</script>
+<!-- End Matomo Code -->
+</head>";
+sub_filter_once on;
+EOF
+
+    fix-permissions /etc/nginx/conf.d/
```
Schnitzel: we don't need fix-permissions in entrypoints scripts, as in the context of an entrypoint the user is already the random user id. fix-permissions is really only useful during a Docker build as the docker build is running as `root` and therefore would create files with inaccessible permissions

images/nginx/matomo/80-nginx-matomo-config
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
+
+# Inject Matomo tracking snippets via nginx configuration.
+
+# Single site configuration.
+if [ ! -z "$MATOMO_URL" ]; then
```
smlx: This can be just
```
[ "$MATOMO_URL" ]
```
and same below.

thom8: thanks @smlx have updated

seanhamlin:
Should this PR be moved to uselagoon?

tobybellwood:
> Should this PR be moved to uselagoon?

ack yes!

thom8:
Moved to lagoon images repo -- https://github.com/uselagoon/lagoon-images/pull/96

PR #2172, opened 2020-08-22
Fix column type for active/standby routes

### LINKED ISSUE #2171: Active/Standby fails to update API if routes too long
**Describe the bug**
When running an active/standby task, if the value of productionRoutes or standbyRoutes is too long, the task will not update the API

**To Reproduce**
1. Trigger active/standby task
2. Wait for it to complete
3. See that the task has not updated in the API
4. Check error logs and see
```

  "errors": [
    {
      "message": "Data too long for column 'production_routes' at row 1",
      "locations": [
        {
          "line": 1170,
          "column": 3
        }
      ],
      "path": [
        "updateProject"
      ]
    }
  ],
  "data": {
    "updateProject": null
  }
}
```

**Expected behavior**
The API should update with the correct values.

shreddedbacon:
# Checklist

The column type was incorrectly defined as varchar(100) when it should be text.
I think there may be another step required to actually modify this though on existing clusters (may need to update this PR with a alter table modify command)?

# Closing issues
closes #2171

### DIFF
--- a/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
+++ b/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
@@ -181,7 +181,7 @@ CREATE OR REPLACE PROCEDURE
         AND column_name = 'production_routes'
     ) THEN
       ALTER TABLE `project`
-      ADD `production_routes` varchar(100);
+      ADD `production_routes` text;
     END IF;
   END;
 $$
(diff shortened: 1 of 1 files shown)


Schnitzel:
@shreddedbacon 
correct we need an actual `alter table` command that runs if the type is wrong, there is an example here:
https://github.com/amazeeio/lagoon/blob/605a1986dce090d2eaafe9866481853f5ace9537/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql#L282-L300

Schnitzel:
@shreddedbacon merged this in, can you create a follow up ticket for creating the `alter table`?

PR #2178, opened 2020-08-31
Pin curl to >7.68 and pull from Alpine v3.12

tobybellwood:
in #1954 we inadvertently unpinned a curl upgrade that was needed (because #1723). 

In this PR I specify a minimum version of curl required, as well as stipulate that it can be downloaded from the Alpine 3.12 repos (instead of edge)

Fun fact - In Alpine package management `>7.68` means 7.68 or greater - https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Add_a_Package

# Checklist

# Closing issues


smlx (approved):
LGTM

PR #2186, opened 2020-09-04
Add platform-viewer role

shreddedbacon:
# Checklist

This adds a new role to keycloak called `platform-viewer`.
The role has only the ability to:
* view logs in elasticsearch
* view projects in lagoon and associated
  * environments
  * deployments
  * tasks
  * backups
  * facts
  * problems
* view notifications
* view groups
* view ssh keys

# Closing issues


services/logs-db/opendistro_security/roles_mapping.yml
```diff
@@ -10,6 +10,7 @@ all_access:
   backend_roles:
     - admin
     - platform-owner
+    - platform-viewer
```
tobybellwood, line 13: wondering whether the `readall_and_monitor` role may be a better fit here?
https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/resources/static_config/static_roles.yml

shreddedbacon: Oh, mayhaps.

shreddedbacon: Turns out that role is no good, few things that it has no permissions to do which it needs.
The only role that really fits is all_access without going down the path of creating a custom role.

tobybellwood: follow-up shows that the `read` role doesn't grant access to all /get functions on indices - particularly aliases :sob: 
https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/resources/static_config/static_action_groups.yml#L83

services/keycloak/start.sh
```diff
@@ -139,6 +139,8 @@ function configure_api_client {
     # Setup platform wide roles.
     /opt/jboss/keycloak/bin/kcadm.sh create roles --config $CONFIG_PATH -r ${KEYCLOAK_REALM:-master} -s name=platform-owner
     /opt/jboss/keycloak/bin/kcadm.sh add-roles --config $CONFIG_PATH -r ${KEYCLOAK_REALM:-master} --rname admin --rolename platform-owner
+    /opt/jboss/keycloak/bin/kcadm.sh create roles --config $CONFIG_PATH -r ${KEYCLOAK_REALM:-master} -s name=platform-viewer
```
rocketeerbkw, line 142: This script runs on every keycloak startup and needs to be idempotent. It also has to handle both use cases of installing a fresh keycloak (new cluster, CI, etc) and updating existing ones (e.g, prod). For those reasons, every function in here is designed to be run only once per keycloak install. And the [order of functions](https://github.com/amazeeio/lagoon/blob/master/services/keycloak/start.sh#L1665) is handled in the main `configure_keycloak` function.

The update mechanism is handled by first checking if some data exists that would've been created by the function and bailing if it is. It works kind of like drupal `update_N` functions but without the nice `hook_install` that can give you a nice fresh start DX. So your changes to `configure_api_client` won't get made to a production keycloak because of [this check](https://github.com/amazeeio/lagoon/blob/master/services/keycloak/start.sh#L98-L102).

All your changes here need to get moved to a new function, check that it will only be run once, and called from `configure_keycloak`. Any changes that are additions, you can just copy/paste, but anytime you modify an existing resource, you'll have to figure out how to do an edit in place (like changing the policies on the `View Task` permission). You can check the other functions for examples for editing resources.

Also, this is not the first time this has come up, so maybe you can add a comment at the top of this file, or in `configure_keycloak` that explains it to save someone else (I also thought I did that, or suggested that already, but can't find any evidence so 🤷).

rocketeerbkw (changes requested):

services/keycloak/start.sh
shreddedbacon: Ahh, did not know this about the startup script.
I'll adjust accordingly. Thanks!

shreddedbacon:
Closing this one off, we can re-evaluate this later on if required.

PR #2188, opened 2020-09-08
Adjust the TLS ACME cleanup script to allow it to run on Mac OSX

seanhamlin:
# Checklist

The version of `awk` that ships with Mac OSX is missing the `-v` arguement, and results in an error. Through `brew` you can install `gawk` to which will work in it's place.

# Closing issues


tobybellwood:
to be resubmitted against https://github.com/amazeeio/helper-scripts

PR #2198, opened 2020-09-11
Add Gitea-specific webhook

tobybellwood:
GitHub push uses head_commit to get the url for the SHA to build.

In Gitea, this hasn't been replicated yet (https://github.com/go-gitea/gitea/pull/5419/files) and it pushes `"head_commit": null,` over the otherwise GitHub-compatible webhook.

This PR replicates the behaviour of the GitHub webhook, but uses a different component to get the URL (similar to the GitLab behaviour).  Because the Gitea webhook also emits a github header, Gitea needs to be processed before GitHub.

Adding gitea compatibility to Lagoon will promote a better end-to-end local development workflow for people getting to grips with Lagoon.

# Checklist


tobybellwood:
leaving as draft whilst I evaluate the relative success of this (expectation low!)
Current discovery on Gitea:
push ok, branch create fails (no commit[0]), branch delete fails (no commit[0]), force push fails (commit[0] is the removed head, not the after)

tobybellwood:
Confirming that commits[0] isn't going to work for GitHub either...
source

tobybellwood:
The final state of this PR is not to change the GitHub webhook - but to add a new Gitea-compatible webhook to Lagoon.

PR #2200, opened 2020-09-11
refactor entrypoint-readiness

Schnitzel:
as https://github.com/amazeeio/lagoon/pull/2196/ has landed we don't need versioncheck anymore

the entrypoint-readiness would never be called if it does not exist.
also making sure that we only check for the ready file if the entrypoint actually exists (prevents issues with modified images)

### DIFF
--- a/images/commons/entrypoint-readiness
+++ b/images/commons/entrypoint-readiness
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-set -eo pipefail
-
 # simple readiness check to see if a late-running entrypoint has completed
 # but only if the entrypoint actually exists
 if [ -f /lagoon/entrypoints/999-readiness.sh ]; then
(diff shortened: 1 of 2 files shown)


images/commons/entrypoint-readiness
```diff
@@ -1,19 +1,9 @@
-#! /bin/sh
+#!/bin/sh
 
-# simple readiness check to see if a late-running entrypoint has completed
-
-function version_check() {
-  test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1";
-}
+set -eo pipefail
```
smlx: Hey @Schnitzel  I don't think `-o pipefail` is available in POSIX shell.

Schnitzel: mhh good question, but we use it here already: https://github.com/amazeeio/lagoon/blob/ad80d05a66f4b2030efe2ba75ee0661b1fb35bd4/images/commons/lagoon/entrypoints/999-readiness.sh#L3

Schnitzel: so we would need to remove it in both places?

PR #2201, opened 2020-09-11
Reduces the amount of auth checks for allProblems query

timclifford:
This PR address a performance issue with the allProblems query GQL resolver by reducing the number of auth checks required to a per project basis, rather than per problem basis.

# Checklist
- [ ] Fixes a server timeout issue that was being thrown during a keycloak auth check for a large query 

# Closing issues
To add

### DIFF
--- a/docs/administering_lagoon/rbac.md
+++ b/docs/administering_lagoon/rbac.md
@@ -124,7 +124,6 @@ Here is a table that lists the roles and the access they have:
 | getProjectByEnvironmentId | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | getProjectByGitUrl | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | getProjectByName | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
-| getAllProblems | problem | viewAll |  | Yes | Yes |  |  |  |  |  |  |
 | addRestore | restore | add | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | updateRestore | restore | update | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | addSshKey | ssh\_key | add | userId | Yes | Yes |  |  |  |  |  | Yes |
--- a/docs/administering_lagoon/rbac.md
+++ b/docs/administering_lagoon/rbac.md
@@ -124,6 +124,7 @@ Here is a table that lists the roles and the access they have:
 | getProjectByEnvironmentId | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | getProjectByGitUrl | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | getProjectByName | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
+| getAllProblems | problem | viewAll |  | Yes | Yes |  |  |  |  |  |  |
 | addRestore | restore | add | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | updateRestore | restore | update | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | addSshKey | ssh\_key | add | userId | Yes | Yes |  |  |  |  |  | Yes |
--- a/services/api/src/resources/problem/helpers.ts
+++ b/services/api/src/resources/problem/helpers.ts
@@ -4,28 +4,28 @@ import { query} from '../../util/db';
 import { Sql } from './sql';
 
 export const Helpers = (sqlClient: MariaClient) => {
-    const getAllProblems = async (projects, args) => {
-      const source = args.source && args.source.map(t => t.toLowerCase() || []);
-      const severity = args.severity && args.severity.map(t => t.toLowerCase() || []);
-      const envType = args.envType && args.envType.map(t => t.toLowerCase() || []);
+  const getAllProblems = async (projects, args) => {
+    const source = args.source && args.source.map(t => t.toLowerCase() || []);
+    const severity = args.severity && args.severity.map(t => t || []);
+    const envType = args.envType && args.envType.map(t => t.toLowerCase() || []);
 
-      const rows = await query(
-        sqlClient,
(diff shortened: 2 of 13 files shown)


docs/administering_lagoon/rbac.md
```diff
@@ -124,6 +124,7 @@ Here is a table that lists the roles and the access they have:
 | getProjectByEnvironmentId | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | getProjectByGitUrl | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | getProjectByName | project | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
+| getAllProblems | problem | viewAll |  | Yes | Yes |  |  |  |  |  |  |
```
bomoko: @timclifford I'm not sure we need this extra permission.

I think the logic you have below that gets the user's project IDs are enough - that way the access to the projects (and their problems) are controlled via a single point (namely, the user's project list) rather than us now having this separate permissions path

services/api/src/resources/problem/resolvers.ts
```diff
@@ -9,17 +9,37 @@ const logger = require('../../logger');
...
+  }
+  catch (err) {
+    if (!keycloakGrant) {
+      logger.warn('No grant available for getAllProblems');
+      return [];
     }
+
+    userProjectIds = await models.UserModel.getAllProjectsIdsForUser({
+      id: keycloakGrant.access_token.content.sub,
+    });
```
bomoko, line 24: After we get these project ids, let's filter out any projects where the following test fails

```
      await hasPermission('problem', 'view', {
          project: project,
      });
```

services/api/src/resources/problem/resolvers.ts
```diff
@@ -9,17 +9,37 @@ const logger = require('../../logger');
...
   let rows = [];
+  let userProjectIds = [];
 
   try {
-    if (!R.isEmpty(args)) {
-      rows = await problemHelpers(sqlClient).getAllProblems(args.source, args.environment, args.envType, args.severity);
-    }
-    else {
-      rows = await query(sqlClient, Sql.selectAllProblems({source: [], environmentId: 0, environmentType: [], severity: []}));
+    await hasPermission('problem', 'viewAll');
```
bomoko: We can remove this permission and this entire try block's logic (after our discussion)

docs/administering_lagoon/rbac.md
timclifford: as discussed, removed this

services/api/src/resources/problem/resolvers.ts
```diff
@@ -9,17 +9,29 @@ const logger = require('../../logger');
...
   try {
-    if (!R.isEmpty(args)) {
-      rows = await problemHelpers(sqlClient).getAllProblems(args.source, args.environment, args.envType, args.severity);
-    }
-    else {
-      rows = await query(sqlClient, Sql.selectAllProblems({source: [], environmentId: 0, environmentType: [], severity: []}));
-    }
+    const userProjectIds = await models.UserModel.getAllProjectsIdsForUser({
+      id: keycloakGrant.access_token.content.sub,
+    });
```
rocketeerbkw, line 24: This isn't a full proof way to return all the projects the current user has access to. Changing this now might not be a blocker but might need to be addressed at some point down the line.

The first issue is the reliance on `keycloakGrant`. This is only passed to the resolver in the case where the user querying the API is logging in via keycloak. We still have a legacy JWT token system that doesn't use keycloak at all, and in that case this will probably throw an error. In the majority if cases, only other system services use that token so I imagine you might not see issues right now with everything being UI based.

The second issue is usage of the API by lagoon admins. Those user accounts aren't added to all the groups, they just get an admin role. So in this case the method will return 0 projects and admins won't be able to see any data. The way we get around this in other resolvers is to create a new scope `viewAll` which we grant to only admins. If a user can `viewAll` projects then just do `select * from` otherwise only load the data the user has access to. For this issue, it looks like you might need to check twice, once for `project viewAll` and then add a new `viewAll` scope for `problem` and check for that also? The best example is probably the [`getAllProjects`](https://github.com/amazeeio/lagoon/blob/main/services/api/src/resources/project/resolvers.ts#L26) resolver.

services/api/src/typeDefs.js
```diff
@@ -767,7 +767,7 @@ const typeDefs = gql`
     """
     Returns all Problems matching given filter (all if no filter defined)
     """
-    allProblems(source: [String], project: Int, environment: Int, envType: [EnvType], identifier: String, severity: [ProblemSeverityRating]): [Problem]
```
rocketeerbkw, line 770: This is technically a backwards compatibility break and would require a major version increase (candidate for lagoon 2?). Since lagoon is open source and the problems stuff is included in core, we should figure how to handle this. Maybe we can mark some part of the API as experimental to avoid holding stuff like this up? It's a similar issue with the billing stuff, were we are breaking stuff technically out of contract but it's ok for now because no one else is using the features.

More of a question/thought for @tobybellwood.

services/api/src/resources/problem/helpers.ts
```diff
@@ -1,61 +1,31 @@
...
+  };
 
-        return await Promise.all(withProjects);
-    };
+  const getSeverityOptions = async () => (
+    R.map(
+      R.prop('severity'),
+        await query(sqlClient, Sql.selectSeverityOptions()),
+      )
+  );
```
rocketeerbkw, line 25: Fun fact, this is a common enough use case that Ramda has a standalone method `pluck`.

```JavaScript

const getSeverityOptions = async () => 
      R.pluck('severity',
        await query(sqlClient, Sql.selectSeverityOptions()),
      );
```

rocketeerbkw (commented):
Just saw a couple things that can be improved with the help of some institutional knowledge (now or later).

services/api/src/resources/problem/resolvers.ts
timclifford: @rocketeerbkw thanks for the detailed response 👍 
@bomoko something for us to discuss:
1. question is, do we address the non-keycloak/legacy auth issue now - or come back to it?
2. we did originally implement a `viewAll` permission for problems, but went against it on the basis that all problems are directly linked to the same project check anyway. Would only a `project viewAll` permission scope be required?

tobybellwood:
@timclifford and @rocketeerbkw is there anything in this PR that would still be of use in the current performance work?

PR #2202, opened 2020-09-13
Kubernetest Object - TypeDef Changes (WIP)

justinlevi:


services/api/src/typeDefs.js
```diff
@@ -278,17 +278,21 @@ const typeDefs = gql`
     uptimeRobotStatusPageId: String
   }
 
-  type Openshift {
+  type Kubernetes {
```
tobybellwood: I thought we'd decided to abstract the type naming to a `DeployTarget` or similar instead of Openshift or Kubernetes, and then specify the type of deploytarget - be it Openshift, Kubernetes or whatever comes next

justinlevi: Yes, based on the conversations here: https://github.com/amazeeio/lagoon/pull/1582 I think you are right. I'll touch base with Brandon on this point tomorrow morning.

Schnitzel: I would still keep the object `Kubernetes`, but on the project itself we use as the key `deployTarget` to reference to which kubernetes we want to deploy too. With that we could in the future also add a new object for another type of cluster, which then also can be referenced under `deployTarget` in the the project.

services/api/src/typeDefs.js
```diff
@@ -461,7 +465,7 @@ const typeDefs = gql`
     """
     Reference to OpenShift Object this Project should be deployed to
     """
-    openshift: Openshift
+    openshift: Kubernetes
```
Schnitzel: we also discussed that we want the possibility to define multiple deploytargets based on the type of environment. So you can define a cluster for production and one for development environment types.
this also means we need to store the actual deployed cluster on the environment itself.

services/api/src/typeDefs.js
```diff
@@ -278,17 +278,21 @@ const typeDefs = gql`
...
   }
 
-  type Openshift {
+  type Kubernetes {
     id: Int
+    """
+    Unique Name
+    """
     name: String
-    consoleUrl: String
```
Schnitzel, line 284: I agree that we should rename this, but currently this removes it? let's call it like `apiUrl`?

services/api/src/typeDefs.js
rocketeerbkw: I thought we decided kubernetes was "the one" so didn't abstract it at all. In order to avoid future breaking changes in case something new comes up, I did add in an interface, but still keeping the single `Kubernetes` object with a field that defines whether it's openshift/k8s under the hood.

services/api/src/typeDefs.js
rocketeerbkw: It was a bit tricky to come up with a solution for this but added a new `DeploymentTargetRule` type that works similar to billing modifiers that I think can handle all the situations that could come up.

services/api/src/typeDefs.js
rocketeerbkw: I thought we could get rid of it because the openshift/k8s services were going away, but forgot we still need it in `ssh` service at least. Renamed to `apiEndpoint`.

services/api/src/typeDefs.js
```diff
@@ -278,16 +283,22 @@ const typeDefs = gql`
...
-    token: String
-    routerPattern: String
-    projectUser: String
-    sshHost: String
-    sshPort: String
     created: String
+    apiType: KubernetesApiType
+    apiEndpoint: string
+    apiToken: string
```
Schnitzel: I think with the new ssh-portal system https://github.com/amazeeio/lagoon/pull/2179 we will not need any apiEndpoint or apiToken stored anymore, as everything happens now inside the cluster itself.
Can you confirm @smlx?

services/api/src/typeDefs.js
```diff
@@ -546,9 +537,13 @@ const typeDefs = gql`
...
     """
-    Name of the OpenShift Project/Namespace this environment is deployed into
+    The DeploymentTarget this environment is deployed into
+    """
+    deploymentTarget: DeploymentTarget
     """
-    openshiftProjectName: String
+    Identifier of this environment in the deployment target
+    """
+    deploymentTargetId: String
```
Schnitzel, line 545: what is this id? the kubernetes namespace?

Schnitzel (commented):
nice! just a couple of questions and we might be able to get rid of the Endpoint and Token all together.

services/api/src/typeDefs.js
tobybellwood: on naming: `deploymentTargetId` seems more like the Id of a deploymentTarget than the namespace within it?

How does this map to the uniqueName components (environmentName.projectName.kubernetesName)

services/api/src/typeDefs.js
```diff
@@ -984,19 +1001,19 @@ const typeDefs = gql`
     patch: UpdateTaskPatchInput!
   }
 
-  input AddOpenshiftInput {
+  input AddKubernetesInput {
```
tobybellwood, line 1009: It would be odd here to use AddKubernetesInput to add an apiType of OPENSHIFT, and maybe just continues the same misnaming?  Why do we not just AddDeployTarget (/Update/Delete the same)

Schnitzel: I don't think so, in the end OpenShift is of type Kubernetes, it's just a special type of Kubernetes. so I think this is correct, you add a kubernetes cluster and tell lagoon that this is an openshift.

While if we would use `AddDeployTarget` you also need a way to tell graphql that you want to add an object `Kubernetes` and also different deployTargets might have different attributes (like a mesosphere cluster might not have a remoteShellHost)

tobybellwood: I'm happy as long as we do have that flexibility - in the end Kubernetes is a type implementing deployTarget.  If each deployTarget type has different required fields, then this is definitely easier :+1: 

As you were

services/api/src/typeDefs.js
Schnitzel: agree, probably better call it something like `environmentTargetId`?

> How does this map to the uniqueName components (environmentName.projectName.kubernetesName)

that's a good question, my understanding is that if we query the API for an environment we don't use the `name` of environment anymore, but instead use the `uniqueName`, as the name could change and having the graphql api define 3 objects that define the environment (name, project, deploymenttarget) is quite heavy.

services/api/src/typeDefs.js
```diff
@@ -518,6 +505,10 @@ const typeDefs = gql`
     """
     name: String
```
Schnitzel: should we change this to `displayName`? As this suggests more that this is now a name for humans and could actually change?

services/api/src/typeDefs.js
rocketeerbkw: yes, this is just the generic version of the `openshiftProjectName`, naming is hard, open to suggstions. It doesn't have any relation to the `uniqueName`

services/api/src/typeDefs.js
rocketeerbkw: and keep `uniqueName` or rename that to `name`? changing what `name` returns is a potentially more confusing break than renaming it though

services/api/src/typeDefs.js
Schnitzel: my inclination would be `environmentDeploymentTargetNamespace` as in the end this would define what it is: the namespace of this environment on the DeploymentTarget, two problems though:
1. it's long (oh well?)
2. not every deploymentTarget might actually have a namespace, so this could cause issues in the future. But at the end we're gonna deploy 100% into kubernetes the next years, so maybe some hard dependencies on kubernetes might not be bad?

services/api/src/typeDefs.js
Schnitzel: I would have sunsetted `name` all together and just use `uniqueName` and `displayName`, we could still have `name` as an attribute on the Environment and just return the `displayName` for now, but mark it as deprecated?

services/api/src/typeDefs.js
rocketeerbkw: hmm.... maybe we don't need this at all anymore? What if we store the environment uniqueName on the k8s namespace (or some other cluster object) instead? Is it enough that lagoon-remote can query for the lagoon environment based on uniqueName  and the lagoon-core doesn't need to know the namespace name?

If that's not possible... probably we can stop relying on the `openshiftProjectName` that's used all over the other mutations and switch to `uniqueName` then this obtuse name doesn't matter as much?

services/api/src/typeDefs.js
justinlevi: Update pushed up including uniqueName and displayName

services/api/src/resolvers.js
```diff
@@ -115,12 +115,12 @@ const {
...
-  addOpenshift,
-  deleteOpenshift,
+  addKubernetes,
+  deleteKubernetes,
   getAllOpenshifts,
   getOpenshiftByProjectId,
-  updateOpenshift,
-  deleteAllOpenshifts,
+  updateKubernetes,
+  deleteAllKubernetes,
```
tobybellwood: This should be removed in #2123

services/api/src/typeDefs.js
tobybellwood: I do think that there should be one name to rule them all, but in the meantime, I think this name `deploymentTargetId` does need to be clearer (that it's not the ID of the deployment target)

services/api/src/typeDefs.js
smlx: Yes, however `ssh-portal` cannot run in the remote cluster until https://github.com/amazeeio/lagoon/issues/2181 is implemented.

services/api/src/typeDefs.js
smlx: I don't have any strong opinion, but could you use `fullyQualifiedEnvironmentName`? That decouples the variable name from the deployment target implementation, and you can define what "fully qualified" means on a per-deployment-target basis.

Obviously this name borrows from [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name).

services/api/src/resolvers.js
justinlevi: reverted this one to make merging w/ 2123 easier

tobybellwood:
A lot of this has been added already - but we're still fighting with openshift/kubernetes naming

PR #2203, opened 2020-09-14
Update URL of renamed lagoon-charts repo

rocketeerbkw:
# Checklist

The repo for lagoon charts was renamed, this updates the URL in the makefile.

# Closing issues
n/a


tobybellwood:
Tbh we could probably also rename the alias from amazeeio to lagoon-charts - doesn't change much, but explains more?

rocketeerbkw:
Are you talking about this part `repo add amazeeio`? If `amazeeio` can be any string then it probably makes sense to change it too.

tobybellwood:
Yup, it's just an internal reference - it can be anything, but would also need to change in the `update` line too

smlx:
FWIW I just used `lagoon` here: https://uselagoon.github.io/lagoon-charts/

Issue #2231, opened 2020-09-30
Compatibility of Lagoon with Harbor 2


Schnitzel:
Seems that Harbor 2 has breaking api changes which Lagoon is not aware yet. 
As we are installing Harbor 2 with our lagoon charts: https://github.com/uselagoon/lagoon-charts/compare/add_harbor#diff-4f3175c54d7aa8d5027622b86f693066R5 we should make sure that Lagoon has compatibility to Harbor

Depending on the changes in Harbor2 it would be great if there is maybe a way to tell Lagoon if the API is v2 or v1, as we might not be able to update all Harbors to v2 when we migrate to Lagoon 2.

tobybellwood:
Lagoon is fully compatible with the latest Harbor releases :partying_face:

PR #2232, opened 2020-09-30
Add Harbor v2 API Compatibility

Schnitzel:
this addresses #2231 to make it just compatible with harbor2 but no more compatible with harbor1.


services/api/src/clients/harborClient.js
```diff
@@ -12,10 +12,13 @@ const lagoonHarborRoute = R.compose(
   R.propOr('', 'LAGOON_ROUTES'),
 )(process.env);
 
+const apiVersion = R.propOr('v1', 'HARBOR_API_VERSION', process.env);
```
Schnitzel: doesn't this mean that if `HARBOR_API_VERSION` is not set the default will be `v1`?
let's make the default `v2` :)

cdchris12: So, we're currently only setting that variable if we deploy the API via Helm. If we want this to be backwards compatible, we would either need to pass that var into this container somehow or leave `v1` as the default.

Schnitzel: I think we're talking about different backwards compatibilities then:
- With Lagoon 2.0 (which is the main branch) we will only be able to deploy lagoon/api via helm, there will be no other way to do this (deploying lagoon via lagoon will not be possible anymore).
- But we might have customers that deploy lagoon/api via helm and want to use an Harbor Version 1 , but that should not be the standard anymore, as we deploy Harbor 2 via our Harbor charts in lagoon-charts repo.
So we can make everything Harbor 2 by default, just it would be cool if we can still support Harbor 1 for some Lagoon installations.

cdchris12: Ah yeah, I was trying to keep this usable for a Lagoon release which would be deployed via Lagoon. Your points are salient; I'll push that change up shortly. 😸

smlx (approved):
I've tested this in the charts repo: :heavy_check_mark:

PR #2237, opened 2020-10-02
Facts api changes

timclifford:
Facts API and UI updates

API
- [x] Adding source and description fields to facts
- [x] New addFacts and deleteFactsFromSource GQL resolvers that allows bulk adding/removing facts

UI 
- [x] Updated UI Facts tab with new source and description fields
image
image


fjgarlin (approved):

tobybellwood:
We'll have to discuss which branch to merge this to.  Lagoon 2 and all development is currently on main.  Master may be used if we need 1.x releases in the meantime.

Schnitzel:
@tobybellwood 
yea, I discussed with the SRE team that we would get this into 1.9.2 and then merge master into main, as this is a rather small change and would help the maintenance teams a lot, I think it's worth doing this.

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -1128,6 +1128,26 @@ CREATE OR REPLACE PROCEDURE
...
+      FROM INFORMATION_SCHEMA.COLUMNS
+      WHERE
+        table_name = 'environment_fact'
+        AND table_schema = 'infrastructure'
+        AND column_name = 'source'
+    ) THEN
+      ALTER TABLE `environment_fact`
+      ADD `source` varchar(300) NOT NULL default ''
+      ADD `description` TEXT NULL DEFAULT ''
+      ALTER TABLE `environment_fact`
```
Schnitzel: @timclifford 
there is a too much `ALTER TABLE` in there, plus also a semicolon missing

Schnitzel (changes requested):

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
timclifford: @Schnitzel added a fix

Schnitzel:
@timclifford 
sorry I think I confused myself, we need two alter tables, see example:
https://github.com/amazeeio/lagoon/blob/dcbb74ce0ee688a123d7564fcfbaa102f91009e4/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql#L274-L277

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -1128,6 +1128,25 @@ CREATE OR REPLACE PROCEDURE
...
+      SELECT NULL
+      FROM INFORMATION_SCHEMA.COLUMNS
+      WHERE
+        table_name = 'environment_fact'
+        AND table_schema = 'infrastructure'
+        AND column_name = 'source'
+    ) THEN
+      ALTER TABLE `environment_fact`
+      ADD `source` varchar(300) NOT NULL default ''
+      ADD `description` TEXT NULL DEFAULT '';
```
Schnitzel:
```suggestion
      ALTER TABLE `environment_fact`
      ADD `source` varchar(300) NOT NULL default '';
      ALTER TABLE `environment_fact`
      ADD `description` TEXT NULL DEFAULT '';
```

timclifford: @Schnitzel https://github.com/amazeeio/lagoon/pull/2237/commits/8d41b7673ca3359137dfb35736c6486e3c9da5f1

PR #2243, opened 2020-10-09
Bugfix: Recent Billing Projects w/o Billing Groups

justinlevi:
This PR fixes an issue with a billing admin page so you can add and change the billing group for a recent project. 

image


tobybellwood:
This was merged to main instead of master - will need to be resubmitted to master for 1.11 @Schnitzel @justinlevi

PR #2267, opened 2020-10-22
API hits calculations for new logging system

### LINKED ISSUE #2266: api hits calculation compatible to new lagoon-logging system
make the api hits calculation compatible with the new lagoon-logging system with still keep backwards compatibility with the old system.

justinlevi:

# Closing issues

Put `closes #2266 `

### DIFF
--- a/services/api/src/models/environment.ts
+++ b/services/api/src/models/environment.ts
@@ -441,7 +441,8 @@ export const EnvironmentModel = (clients) => {
             --> if both are 0, use newResult.average
       */
 
-      for (let i = 0; i < legacyResultCount; i++) {
+      const count = newResultCount > 0 ? newResultCount : legacyResultCount;
+      for (let i = 0; i < count; i++) {
         if (newResultCount !== 0 && newBuckets[i] && newBuckets[i].count && newBuckets[i].count.value !== 0){
           // We have new logging data, use this for total
           total += newBuckets[i].count.value;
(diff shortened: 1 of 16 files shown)


services/api/src/models/environment.ts
```diff
@@ -334,12 +415,22 @@ export const EnvironmentModel = (clients) => {
...
       // 0 hits found in elasticsearch, don't even try to generate monthly counts
-      if (result.hits.total.value === 0) {
+      if (legacyResult.hits.total.value === 0 && newResult.hits.total.value === 0) {
         return { total: 0 };
       }
 
       var total = 0;
 
+      const result = ((legacyResult, newResult) => {
+        if (legacyResult.hits.total.value !== 0){
```
Schnitzel: that's probably not gonna work, we won't exactly switch from the old to the new system exactly on the first of the month. So assuming we switch on the 4th of the month from the legacy to the new system: `legacyResult` still will have some hits and therefore the hits will always be reported for the legacy system, even though on the 5th on the hits arrive in the new system.

see comment below how we could handle this.

Schnitzel: I think we could change it like that:

pseudocode:
```
foreach hourlybucket (#both result buckets should have the exact same amount of buckets)
add to total
   if newResult.bucketcount is not 0 use newResult.bucketcount
   if legacyResult.bucketcount is not 0 use legacyResult.bucketcount
   if newResult.bucketcount is 0 and legacyResult.bucketcount is 0
      --> if newResult.average is not 0, use newResult.average
      --> if legacyResult.average is not 0, use legacyResult.average
      --> if both are 0, use newResult.average 
```

Schnitzel (changes requested):

services/api/src/models/environment.ts
justinlevi: Ahh, good point. Will take this into account and push up some changes

justinlevi:
The latest set of changes have been deployed to the openshift billing environment and prod has been sync'd. We may need to further discuss any discrepancies in project hit calculations. Would like another set of eyes on this as I know this is time sensitive.

PR #2300, opened 2020-10-31
Use image cache and harbor for public images in kubernetes

rocketeerbkw:
# Checklist

This solves two problems:

1. Use the amazee.io image cache for public images in `kubectl-build-deploy-dind`
2. Copies public images into harbor, then use those images in k8s instead of pulling directly from docker hub (or the image cache)

# Closing issues

n/a

### DIFF
--- a/images/kubectl-build-deploy-dind/Dockerfile
+++ b/images/kubectl-build-deploy-dind/Dockerfile
@@ -21,4 +21,6 @@ COPY scripts /kubectl-build-deploy/scripts
 
 COPY helmcharts  /kubectl-build-deploy/helmcharts
 
+ENV IMAGECACHE_REGISTRY=imagecache.amazeeio.cloud
+
 CMD ["/kubectl-build-deploy/build-deploy.sh"]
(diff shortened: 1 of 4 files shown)


rocketeerbkw:
In order to test locally, I ended up modifying `tests/files/drupal9-mariadb/docker-compose.yml` and adding a new service:

```docker-compose
solr:
   image: amazeeio/solr:6.6-drupal
   labels:
     lagoon.type: solr
   expose:
     - "8983"
```

I saw no problems in my local k8s, it looked like a solr deployment and pod were created as I would expect.

I checked all the test files and none of them use external images, they all use a local dockerfile. I'm not sure what the best way to add a test for this new public image copying stuff. I didn't commit the change because I thought there might be an issue with jenkins talking to the same image cache for every build test?

Schnitzel (approved):

Schnitzel:
tested with running "amazeeiolagoon/kubectl-build-deploy-dind:pr-2300" directly inside a kubernetes cluster and that worked perfectly, nice work! thank you.

PR #2236, opened 2020-10-02
stop main branch from building base images

tobybellwood:
This PR does a number of things:
1) It removes from /lagoon the base images that are now built at uselagoon/lagoon-images
2) Non-tagged images are now pushed to the testlagoon dockerhub
3) Tagged images will be pushed to uselagoon dockerhub (and the base images not transferred to lagoon-images can be pushed to amazeeio dockerhub) 
4) It references the :latest tag from the uselagoon dockerhub for base images
5) It references the :latest tag from the testlagoon dockerhub for test images
6) both are overridable with UPSTREAM_REPO and UPSTREAM_TAG (test images may have to be sed'ded if we can't work out where to add the build-arg)

This branch was derived from the lagoon-core branch, so there are a few changes to the test image and api-data-watcher pusher.  I'd suggest that those changes be merged in first?

# Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [x] PR title is ready for changelog and subsystem label(s) applied

PR #2316, opened 2020-11-06
Stop main branch from building base images and publish to use/testlagoon repos

tobybellwood:
This PR supersedes #2236 

It removes the images that are being built in github.com/uselagoon/lagoon-images from building here, and configures the service images to use the `uselagoon/*:latest` images as upstream images.

In addition, it publishes all branch images to testlagoon dockerhub, and main pushes also get a :latest tag
All tagged pushes get pushed to uselagoon dockerhub along with a :latest tag


AlexSkrypnyk:
@tobybellwood 
Has there been an announcement anywhere about moving to http://github.com/uselagoon/lagoon-images repo? I would like to know about such significant changes next time. Thanks

tobybellwood:
There will be very shortly - this is (up to now) a behind-the-scenes change, with a heap of moving parts.  All customer-facing images are still receiving backward-compatible releases to the amazeeio/* namespace on docker hub.

PR #2329, opened 2020-11-13
Fix PHP 7.4 deprecated argument order for implode()

### LINKED ISSUE #2328: PHP 7.4  deprecated argument order for implode()
In PHP 7.4 passing the glue after the pieces (i.e. not using the documented order of parameters) has been deprecated for implode().

In docs/docs/examples/drupal7-mariadb/sites/default/settings.php the deprecated form is used:
`$conf['varnish_control_terminal'] = implode($varnish_hosts, " ");`

KimNyholm:
In PHP 7.4 passing the glue after the pieces (i.e. not using the documented order of parameters) has been deprecated for implode().

In docs/docs/examples/drupal7-mariadb/sites/default/settings.php the deprecated form is used:
`$conf['varnish_control_terminal'] = implode($varnish_hosts, " ");`

Fixes #2328

### DIFF
--- a/docs/docs/examples/drupal7-mariadb/sites/default/settings.php
+++ b/docs/docs/examples/drupal7-mariadb/sites/default/settings.php
@@ -58,7 +58,7 @@
 
     $conf['reverse_proxy'] = TRUE;
     $conf['reverse_proxy_addresses'] = array_merge(explode(',', getenv('VARNISH_HOSTS')), array('varnish'));
-    $conf['varnish_control_terminal'] = implode($varnish_hosts, " ");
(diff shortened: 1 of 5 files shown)


rocketeerbkw:
Thanks for the fix! We also have the reversed arguments in all our test files, would you also care to update those as well? Just search for `implode` in the whole repo.

KimNyholm:
I have changed argument order for implode() in the tests as well, but I will leave the testing to you since I have no idea about how to run the tests.

rocketeerbkw:
Awesome, thank you! Our tests are failing for unrelated reasons, but I don't expect these simple changes to cause any issues itself, will try re-running them.

rocketeerbkw (approved):

PR #2332, opened 2020-11-17
Add projectsByOpenshift API Endpoint

cdchris12:
# Checklist

This PR adds the `projectsByOpenshift` API endpoint, which will show all projects which are hosted on a specified openshift.


services/api/src/resources/project/resolvers.ts
```diff
@@ -187,6 +187,47 @@ export const getProjectByName: ResolverFn = async (
...
+) => {
+  let where;
+  try {
+    await hasPermission('project', 'viewAll');
+  } catch (err) {
+    if (!keycloakGrant) {
+      logger.warn('No grant available for getProjectByOpenshift');
+      return [];
+    }
+  }
```
rocketeerbkw: This is not a sufficient way to verify the user is allowed to view the projects being returned. If the user has `viewAll` permissions, then we don't have to limit the SQL query. In the `else` case, we have to figure out what projects the user _does_ have access to, and limit the SQL query to only return those projects.

See [`getAllProjects`](https://github.com/amazeeio/lagoon/blob/main/services/api/src/resources/project/resolvers.ts#L50-L57) for an example.

rocketeerbkw (changes requested):

services/api/src/resources/project/resolvers.ts
cdchris12: Thanks for the quick review. I've updated the PR accordingly. :)

rocketeerbkw:
There probably needs to be some thought put into what `openshift` permissions are required for this query. There are `openshift:view` and `openshift:viewAll` permissions used in other places and the dependency is that the system needs to know the project first.

In other words, to know if a user has permission to view a specific openshift data, we must first figure out of that user is part of a project that has that openshift linked to it. I don't think this PR addresses that use case either.

If this is only for system engineers maybe it's not worth making available for non-admins to run. You could replace all the permissions try/catch stuff with two checks: `await hasPermission('openshift', 'viewAll');` and `await hasPermission('project', 'viewAll');` to limit the whole query.

cdchris12:
@rocketeerbkw Thanks again for the feedback. :) 

With the latest push, I've added functionality to filter shown projects based on openshift permissions which are derived from the list of projects the user has access to. Hopefully this is what you were looking for here?

rocketeerbkw:
Superseded by #2686

PR #2333, opened 2020-11-17
Use new nginx-ingress stable location

shreddedbacon:
# Checklist

As per https://helm.sh/blog/new-location-stable-incubator-charts/ the helm stable chart location changed, and on november 13 became unusable. This prevent the `stable/nginx-ingress` being installed like it normally does.

This just installs it like other charts we already install as part of testing but uses the new chart location specifically.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -877,13 +877,15 @@ endif
 		--image docker.io/rancher/k3s:$(K3S_VERSION) \
 		--volume $$PWD/local-dev/k3d-registries.yaml:/etc/rancher/k3s/registries.yaml \
 		$(K3D_BTRFS_VOLUME) \
-		-x --no-deploy=traefik \
-		--volume $$PWD/local-dev/k3d-nginx-ingress.yaml:/var/lib/rancher/k3s/server/manifests/k3d-nginx-ingress.yaml
+		-x --no-deploy=traefik
 	echo "$(K3D_NAME)" > $@
 	export KUBECONFIG="$$(./local-dev/k3d get-kubeconfig --name='$(K3D_NAME)')"; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -877,13 +877,15 @@ endif
...
 		$(K3D_BTRFS_VOLUME) \
-		-x --no-deploy=traefik \
-		--volume $$PWD/local-dev/k3d-nginx-ingress.yaml:/var/lib/rancher/k3s/server/manifests/k3d-nginx-ingress.yaml
+		-x --no-deploy=traefik
 	echo "$(K3D_NAME)" > $@
 	export KUBECONFIG="$$(./local-dev/k3d get-kubeconfig --name='$(K3D_NAME)')"; \
 	local-dev/kubectl --kubeconfig="$$(./local-dev/k3d get-kubeconfig --name='$(K3D_NAME)')" apply -f $$PWD/local-dev/k3d-storageclass-bulk.yaml; \
 	docker tag $(CI_BUILD_TAG)/docker-host localhost:5000/lagoon/docker-host; \
 	docker push localhost:5000/lagoon/docker-host; \
+	local-dev/kubectl --kubeconfig="$$(./local-dev/k3d get-kubeconfig --name='$(K3D_NAME)')" --context='$(K3D_NAME)' create namespace nginx-ingress; \
```
smlx, line 886: `helm upgrade` takes a `--create-namespace` argument that does this, might be easier?

shreddedbacon: I'm not fussed. I just copy pasta'd one of the ones below and renamed things.

Don't know how long this will hang around for if we are moving off the deprecated one in the future anyway?

smlx: ah sorry missed that. since the other commands use that approach never mind

smlx (approved):

PR #2334, opened 2020-11-17
Prevent global_tenant access in Kibana

### LINKED ISSUE #2327: Tenant switch in Kibana does not stay permanent or default to user's tennant
For some users, the tenant switch in Kibana seems not to stick and keeps changing. The two possibilities currently known are:
1. Although the global tenant is disabled in the Open Distro configuration, somehow it still appears for some users
2. Global tenant does not appear but the tenant is empty before switching

shreddedbacon:
# Checklist

The way that kibana checks for tenant access eventually falls back to a list of tenants available, which includes `global_tenant`, and will alphabetically sort the available tenants and return them to users when they log in.
This is a problem for users that may have a tenant that is alphabetically after `global_tenant`.

We already disable the global tenant, so this just adds a check in the `getTenantByPreference` that will remove `global_tenant` if global tenant is disabled.

This just takes a copy of this file https://github.com/opendistro-for-elasticsearch/security-kibana-plugin/blob/opendistro-1.6/lib/backend/opendistro_security.js and COPY into the logs-db-ui image that contains the fix (L390-L394)

The latest version of `security-kibana-plugin` is v1.11 and the code that performs this check is completely different, but it may also suffer from the same problem, but would need to be tested/verified if we plan to upgrade.

# Closing issues

closes #2327


smlx:
I don't know if this will work since we are using opendistro upstream images now.

There's a [config option](https://opendistro.github.io/for-elasticsearch-docs/docs/security/access-control/multi-tenancy/#configuration) to disable the global tenant that probably needs to be configured?

shreddedbacon:
> I don't know if this will work since we are using opendistro upstream images now.
> 
> There's a [config option](https://opendistro.github.io/for-elasticsearch-docs/docs/security/access-control/multi-tenancy/#configuration) to disable the global tenant that probably needs to be configured?

Do we not set that already with the upstream images then? 
We did on the images we used to build/run, but global_tenant was still accessible (https://github.com/amazeeio/lagoon/blob/master/services/logs-db-ui/Dockerfile#L38)

So unless we aren't setting it, the bug is still present in upstream opendistro and maybe the bug needs to be raised there?

Edit: so we do disable the global and private tenants still. So the issue is also in the upstream opendistro images.
```
  multitenancy:
    enabled: true
    tenants:
      enable_global: false
      enable_private: false
      preferred:
      - admin_tenant
```

shreddedbacon:
Looking at opendistro 1.10 (which is what we look to be running?) it uses the same file that is in this PR to handle the tenant selection https://github.com/opendistro-for-elasticsearch/security-kibana-plugin/blob/opendistro-1.10/lib/backend/opendistro_security.js.

But 1.11+ it changes entirely as the file is gone entirely, so possible that it is fixed in a newer version too?

smlx:
Yeah it's entirely possible that it's fixed upstream. Looking at this https://github.com/opendistro-for-elasticsearch/security-kibana-plugin/issues?page=1&q=tenant there are a lot of recent fixes.

Schnitzel:
mhh so while this is an epic fix, we don't actually use `logs-db-ui` anymore in production :( it's in here just for historical reasons, but we removed it from the `main` branch.

It's correct that we're running opendistro 1.10 currently in production, so we might need to upgrade to 1.11 in order to test if this is fixed there?

also reminds me of https://github.com/uselagoon/lagoon-charts/issues/109 which we probably should fix soon

shreddedbacon:
This won't work anymore

PR #2355, opened 2020-12-01
BaaS bucket name override support

### LINKED ISSUE #2352: Lagoon should support defining custom baas bucket names
With the ongoing migrations from Openshift to k8s, we've realized that moving our old baas repos from one AWS account to another is not tenable. Thus, we've decided that it would be in our best interest to create new baas repos for migrated projects, due to S3 having globally unique bucket names.

Ideally, there would be a project level environment variable, `baasBucketName`, which would be used by Lagoon to set that project's environments' baas bucket name inside created schedule objects if this env var is present. If the env var is not present, then Lagoon will simply fall back to its current behavior of generating a baas bucket name.

shreddedbacon:
# Checklist

See #2352 for reasoning behind this PR. 

Define `LAGOON_BAAS_BUCKET_NAME` in the API at the project level as a `build` scope (so it doesn't end up in a configmap)

# Closing issues

closes #2352


cdchris12 (approved):
I tested this on my local machine and in PR #2359 on the CI machines. The only issues I found to be causing these tests to not pass were related to our test suite, not this PR. That being said, this functionality lg2m !

PR #2361, opened 2020-12-02
Active standby fixes

shreddedbacon:
# Checklist

Lagoon-charts was updated recently to install dioscuri, this meant that two versions of dioscuri were running and competing for the task.

This PR updates the makefile so that lagoon-remote doesn't install dioscuri.

Dioscuri was also updated to do things slightly differently for when the ingress-nginx admission webhook comes into play, so this PR has references to that branches chart, once confirmed that the tests pass here https://github.com/amazeeio/dioscuri/pull/11 will need to be merged and this PR updated again to use `main` chart instead of `admissionfix`

Additional changes were made to the active/standby tasks in `kubernetesmisc` and `openshiftmisc` to increase the timeout to 60s as sometimes the task could take slightly longer than the original timeout. This timeout isn't a problem when moving to use the lagoon builddeploy controllers though.

# Closing issues


tobybellwood (approved):
LGTM

PR #2367, opened 2020-12-03
Get CI running using helm charts

smlx:
# Checklist

This PR refactors the Jenkins pipeline to get CI running using the Lagoon helm charts. The pipeline now uses [declarative syntax](https://www.jenkins.io/doc/book/pipeline/syntax/#declarative-pipeline) so it's a bit shorter and hopefully easier to understand. It also makes it easier to run a test [matrix](https://www.jenkins.io/doc/book/pipeline/syntax/#declarative-matrix) in future (e.g. multiple k8s versions).

Where the workflow has changed, it now does this:

* Build the images
* Push the images to `testlagoon/*`
* Create a kind cluster
* Copy the built images into the cluster cache (avoid pulling layers down again)
* Run tests in the cluster

The steps following a successful test run should be the same.

Because of the way that the [chart-testing](https://github.com/helm/chart-testing) tool works, we no longer have live log messages in Jenkins. Instead the test job will print a command which sets up a local `$KUBECONFIG` and a ssh port-forward so that you can interact with the `kind` cluster while it's running to tail logs, watch the tests run etc. Check the logs of the `Running tests on kind cluster` stage to see something like this:

Screenshot from 2020-12-03 10-52-50

The things that I'd like special review attention on:

* Does the `Jenkinsfile` logic look correct compared to the old pipeline? Particularly around `SKIP_IMAGE_PUBLISH`, and when to push to `uselagoon/*`?
* Do the other workflows that use the `Makefile` still work? I don't do much with e.g. `docker-compose`, so could someone who uses that functionality confirm that it still works?
* Is there documentation that needs updating?
* Have I missed any test suites that need to be included and run?

# Closing issues

n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -931,7 +931,7 @@ KIND_VERSION = v0.9.0
 GOJQ_VERSION = v0.11.2
 KIND_IMAGE = kindest/node:v1.19.1@sha256:98cf5288864662e37115e362b23e4369c8c4a408f99cbc06e58ac30ddc721600
 TESTS = [features-kubernetes,nginx,active-standby-kubernetes,drupal-php72,drupal-php73,drupal-php74]
-CHARTS_TREEISH = lagoon-test-0.8.1
+CHARTS_TREEISH = lagoon-test-0.8.2
 
 local-dev/kind:
 ifeq ($(KIND_VERSION), $(shell kind version 2>/dev/null | sed -nE 's/kind (v[0-9.]+).*/\1/p'))
--- a/Makefile
+++ b/Makefile
@@ -965,30 +965,21 @@ helm/repos: local-dev/helm
 kind/cluster: local-dev/kind
 	# these IPs are a result of the docker config on our build nodes
 	./local-dev/kind get clusters | grep -q "$(CI_BUILD_TAG)" && exit; \
-		export KUBECONFIG=$$(mktemp) \
+		docker network create kind || true \
+		&& export KUBECONFIG=$$(mktemp) \
 		KINDCONFIG=$$(mktemp ./kindconfig.XXX) \
+		KIND_NODE_IP=$$(docker run --rm --network kind alpine ip -o addr show eth0 | sed -nE 's/.* ([0-9.]{7,})\/.*/\1/p') \
 		&& chmod 644 $$KUBECONFIG \
-		&& curl -sSLo $$KINDCONFIG https://raw.githubusercontent.com/uselagoon/lagoon-charts/$(CHARTS_TREEISH)/test-suite.kind-config.yaml \
-		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.configs."registry.192.168.48.2.nip.io:32443".tls]'  >> $$KINDCONFIG \
-		&& echo '    insecure_skip_verify = true'                                                                    >> $$KINDCONFIG \
-		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."registry.192.168.48.2.nip.io:32080"]'      >> $$KINDCONFIG \
-		&& echo '    endpoint = ["http://registry.192.168.48.2.nip.io:32080"]'                                       >> $$KINDCONFIG \
-		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.configs."registry.192.168.160.2.nip.io:32443".tls]' >> $$KINDCONFIG \
-		&& echo '    insecure_skip_verify = true'                                                                    >> $$KINDCONFIG \
-		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."registry.192.168.160.2.nip.io:32080"]'     >> $$KINDCONFIG \
-		&& echo '    endpoint = ["http://registry.192.168.160.2.nip.io:32080"]'                                      >> $$KINDCONFIG \
-		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.configs."registry.172.25.0.2.nip.io:32443".tls]'    >> $$KINDCONFIG \
-		&& echo '    insecure_skip_verify = true'                                                                    >> $$KINDCONFIG \
-		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."registry.172.25.0.2.nip.io:32080"]'        >> $$KINDCONFIG \
-		&& echo '    endpoint = ["http://registry.172.25.0.2.nip.io:32080"]'                                         >> $$KINDCONFIG \
-		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]'                               >> $$KINDCONFIG \
-		&& echo '    endpoint = ["https://imagecache.amazeeio.cloud", "https://index.docker.io/v1/"]'                >> $$KINDCONFIG \
-		&& echo 'nodes:'                                                                                             >> $$KINDCONFIG \
-		&& echo '- role: control-plane'                                                                              >> $$KINDCONFIG \
-		&& echo '  image: $(KIND_IMAGE)'                                                                             >> $$KINDCONFIG \
-		&& echo '  extraMounts:'                                                                                     >> $$KINDCONFIG \
-		&& echo '  - containerPath: /var/lib/kubelet/config.json'                                                    >> $$KINDCONFIG \
-		&& echo '    hostPath: $(HOME)/.docker/config.json'                                                          >> $$KINDCONFIG \
+		&& curl -sSLo $$KINDCONFIG.tpl https://raw.githubusercontent.com/uselagoon/lagoon-charts/$(CHARTS_TREEISH)/test-suite.kind-config.yaml.tpl \
+		&& envsubst < $$KINDCONFIG.tpl > $$KINDCONFIG \
+		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]'                >> $$KINDCONFIG \
+		&& echo '    endpoint = ["https://imagecache.amazeeio.cloud", "https://index.docker.io/v1/"]' >> $$KINDCONFIG \
+		&& echo 'nodes:'                                                                              >> $$KINDCONFIG \
+		&& echo '- role: control-plane'                                                               >> $$KINDCONFIG \
+		&& echo '  image: $(KIND_IMAGE)'                                                              >> $$KINDCONFIG \
+		&& echo '  extraMounts:'                                                                      >> $$KINDCONFIG \
+		&& echo '  - containerPath: /var/lib/kubelet/config.json'                                     >> $$KINDCONFIG \
+		&& echo '    hostPath: $(HOME)/.docker/config.json'                                           >> $$KINDCONFIG \
 		&& KIND_CLUSTER_NAME="$(CI_BUILD_TAG)" ./local-dev/kind create cluster --config=$$KINDCONFIG \
 		&& echo -e 'Interact with the cluster during the test run in Jenkins like so:\n' \
 		&& echo "export KUBECONFIG=\$$(mktemp) && scp $$NODE_NAME:$$KUBECONFIG \$$KUBECONFIG && KIND_PORT=\$$(sed -nE 's/.+server:.+:([0-9]+)/\1/p' \$$KUBECONFIG) && ssh -fNL \$$KIND_PORT:127.0.0.1:\$$KIND_PORT $$NODE_NAME" \
--- a/Makefile
+++ b/Makefile
@@ -988,8 +988,10 @@ kind/cluster: local-dev/kind
 		&& echo '  - containerPath: /var/lib/kubelet/config.json'                                                    >> $$KINDCONFIG \
 		&& echo '    hostPath: $(HOME)/.docker/config.json'                                                          >> $$KINDCONFIG \
 		&& KIND_CLUSTER_NAME="$(CI_BUILD_TAG)" ./local-dev/kind create cluster --config=$$KINDCONFIG \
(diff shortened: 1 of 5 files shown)


tobybellwood:
working through this now - Jenkins looks great - just going through local ramifications

tobybellwood:
One outstanding issue is that the `build` images are still using the `amazeeiolagoon/kubectl-build-deploy-dind:master` image - not the one from this repo.  Whilst we're waiting on the downstream fix, we can resolve this by setting the overwrite variable in the API and webhooks2tasks deployments `OVERWRITE_KUBECTL_BUILD_DEPLOY_DIND_IMAGE`

My only request would be for a persistent kubeconfig when building locally (to use with Lens/kubectx etc), but again, that can wait - getting these tests in is more important IMO

Otherwise, this looks good to me (pending conflict resolves!) - `make -j8 build && make -j8 kind/test` runs locally nicely.

We will have to do some post-work to get a workable docker-compose stack to test api/ui development with, but it doesn't seem far off, and we've also got some k3s changes to bring across.

smlx:
> One outstanding issue is that the build images are still using the amazeeiolagoon/kubectl-build-deploy-dind:master image - not the one from this repo. Whilst we're waiting on the downstream fix, we can resolve this by setting the overwrite variable in the API and webhooks2tasks deployments OVERWRITE_KUBECTL_BUILD_DEPLOY_DIND_IMAGE

Thanks for finding this, I'll push a fix.

> My only request would be for a persistent kubeconfig when building locally (to use with Lens/kubectx etc), but again, that can wait - getting these tests in is more important IMO

This one is easy!
```
# find the cluster name
./local-dev/kind get clusters
# export kubeconfig
./local-dev/kind export kubeconfig --name ...
```

I don't know where that kind of documentation should live..

smlx:
Ok @tobybellwood I rebased, added the override for the `build-deploy-dind` image, and fixed some other unrelated breakage. Ready for review again.

Makefile
```diff
@@ -925,3 +924,100 @@ ui-development: build/api build/api-db build/local-api-data-watcher-pusher build
...
+		&& echo -e 'Interact with the cluster during the test run like so:\n' \
+		&& echo "export KUBECONFIG=\$$(mktemp) && scp $$NODE_NAME:$$KUBECONFIG \$$KUBECONFIG && KIND_PORT=\$$(sed -nE 's/.+server:.+:([0-9]+)/\1/p' \$$KUBECONFIG) && ssh -fNL \$$KIND_PORT:127.0.0.1:\$$KIND_PORT $$NODE_NAME" \
+		&& echo -e 'kubectl ...\n'
+
+KIND_SERVICES = api api-db api-redis auth-server broker controllerhandler drush-alias keycloak keycloak-db ssh
+
+.PHONY: kind/preload
+kind/preload: kind/cluster $(addprefix build/,$(KIND_SERVICES))
+	for image in $(KIND_SERVICES); do \
+		KIND_CLUSTER_NAME="$(CI_BUILD_TAG)" ./local-dev/kind load docker-image $(CI_BUILD_TAG)/$$image; \
```
Schnitzel: we decided in the k3d testing infra against the `load` feature to push images inside the kubernetes cluster, mostly because:
1. The loading takes a lot of time and also does not profit of any docker layer caching
2. if you have a tiny change on the image you need to load the whole image again, which takes a lot of time again
3. If you stop the kind cluster and start it again you need to push all images again. 

This is why we create a local-registry and push the images in there:
https://github.com/amazeeio/lagoon/blob/fix-ci/Makefile#L329-L336

and then use them inside k3d (search for `172.17.0.1:5000`)

Do you think we could use the same system also for kind? would make a bit a nicer DX

Makefile
```diff
@@ -925,3 +924,100 @@ ui-development: build/api build/api-db build/local-api-data-watcher-pusher build
...
+	chmod a+x local-dev/kind
+endif
+
+local-dev/jq:
+ifeq ($(GOJQ_VERSION), $(shell jq -v 2>/dev/null | sed -nE 's/gojq ([0-9.]+).*/v\1/p'))
+	$(info linking local jq version $(KIND_VERSION))
+	ln -s $(shell command -v jq) ./local-dev/jq
+else
+	$(info downloading gojq version $(GOJQ_VERSION) for $(ARCH))
+	curl -sSL https://github.com/itchyny/gojq/releases/download/$(GOJQ_VERSION)/gojq_$(GOJQ_VERSION)_$(ARCH)_amd64.tar.gz | tar -xzC local-dev --strip-components=1 gojq_$(GOJQ_VERSION)_$(ARCH)_amd64/gojq
```
Schnitzel, line 957: somehow there is no tar.gz version for the osx version of gojq, see https://github.com/itchyny/gojq/releases/tag/v0.11.2

Makefile
```diff
@@ -925,3 +924,100 @@ ui-development: build/api build/api-db build/local-api-data-watcher-pusher build
...
+		&& echo '    endpoint = ["http://registry.172.25.0.2.nip.io:32080"]'                                         >> $$KINDCONFIG \
+		&& echo 'nodes:'                                                                                             >> $$KINDCONFIG \
+		&& echo '- role: control-plane'                                                                              >> $$KINDCONFIG \
+		&& echo '  image: $(KIND_IMAGE)'                                                                             >> $$KINDCONFIG \
+		&& echo '  extraMounts:'                                                                                     >> $$KINDCONFIG \
+		&& echo '  - containerPath: /var/lib/kubelet/config.json'                                                    >> $$KINDCONFIG \
+		&& echo '    hostPath: $(HOME)/.docker/config.json'                                                          >> $$KINDCONFIG \
+		&& KIND_CLUSTER_NAME="$(CI_BUILD_TAG)" ./local-dev/kind create cluster --config=$$KINDCONFIG \
+		&& echo -e 'Interact with the cluster during the test run like so:\n' \
```
Schnitzel: this is meant to connect from a local host to the running jenkins in order to interact during testing? As the generated command doesn't work locally, maybe we can make this clearer?

Schnitzel (changes requested):
- we still need to test Lagoon 2 against OpenShift/MiniShift, I think we could still run lagoon inside the kind cluster and just start a minishift cluster on the side and run the openshift tests there?
- we're switching from `./local-dev/helm/helm` to `./local-dev/helm` which is ok, but if you checkout the `main` code and try to run it (with an existing `./local-dev/helm/helm` file it fails, which is a bit strange DeveloperExperience. Mabye we can add some nice code that realizes that `./local-dev/helm/helm` exists and removes it?
- what's the plan around k3d? Should this kind testing completely replace the k3d testing infrastructure?

smlx:
Hi @Schnitzel thanks for the review!

>     * we still need to test Lagoon 2 against OpenShift/MiniShift, I think we could still run lagoon inside the kind cluster and just start a minishift cluster on the side and run the openshift tests there?

This is news to me! Could you elaborate on the requirements here?
* Do you mean that the `lagoon-core` and `lagoon-remote` helm charts need to install into OpenShift? Or only `lagoon-remote`? Or will the "remote" be installed without the helm charts?
* OpenShift 3.x, or OpenShift 4.x?

In any case I think this might be outside the scope of this PR, which is really just about getting CI running on `main` in k8s. Could we create an issue for OpenShift support in CI for Lagoon 2 and address this separately, since this PR is already quite large?

>     * we're switching from `./local-dev/helm/helm` to `./local-dev/helm` which is ok, but if you checkout the `main` code and try to run it (with an existing `./local-dev/helm/helm` file it fails, which is a bit strange DeveloperExperience. Mabye we can add some nice code that realizes that `./local-dev/helm/helm` exists and removes it?

I'm not sure if this is worth adding logic to the `Makefile` for. Could we solve this by documenting the use of `git clean -dfx` when checking out a new major Lagoon release?

>     * what's the plan around k3d? Should this kind testing completely replace the k3d testing infrastructure?

I don't know. The reason for using `kind` is that the charts CI already uses `kind` and I am re-using that testing infrastructure.

If `k3d` is _only_ used to test Lagoon in CI then I guess it could possibly be removed. However I don't know if it also has some local development use-case that `kind` cannot replace?

In any case, I think removal of `k3d` is also outside the scope of this PR and would be best addressed in a separate issue.

tobybellwood:
I agree that the scope of this PR is to get the current Lagoon main branch tested on the latest lagoon-charts release.

For phase 2 - getting the lagoon-core inside KinD to deploy to a local Openshift 3 cluster running lagoon-remote would cover all our existing test cases.

I would leave the current k3d cluster in place to test against a locally-run docker-compose stack (analogous to lagoon-core) for API/UI development initially.  Technically the local docker-compose cluster could probably also connect to the lagoon-remote running in KinD, but it may be more logical to separate them initially whilst we figure it out.  We'll also need to work out what else is needed for API/UI development - how do we run Elasticsearch/Kibana and Harbor?

Makefile
smlx: This is an interesting point that I hadn't considered. I'll see if I can rework this a bit.

Makefile
smlx: oh, that's annoying! I'll fix it up.

Makefile
smlx: Sure, I'll amend the instructions.

Makefile
```diff
@@ -925,3 +924,100 @@ ui-development: build/api build/api-db build/local-api-data-watcher-pusher build
...
+		&& echo '    endpoint = ["http://registry.192.168.48.2.nip.io:32080"]'                                       >> $$KINDCONFIG \
+		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.configs."registry.192.168.160.2.nip.io:32443".tls]' >> $$KINDCONFIG \
+		&& echo '    insecure_skip_verify = true'                                                                    >> $$KINDCONFIG \
+		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."registry.192.168.160.2.nip.io:32080"]'     >> $$KINDCONFIG \
+		&& echo '    endpoint = ["http://registry.192.168.160.2.nip.io:32080"]'                                      >> $$KINDCONFIG \
+		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.configs."registry.172.25.0.2.nip.io:32443".tls]'    >> $$KINDCONFIG \
+		&& echo '    insecure_skip_verify = true'                                                                    >> $$KINDCONFIG \
+		&& echo '  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."registry.172.25.0.2.nip.io:32080"]'        >> $$KINDCONFIG \
+		&& echo '    endpoint = ["http://registry.172.25.0.2.nip.io:32080"]'                                         >> $$KINDCONFIG \
```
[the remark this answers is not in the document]
smlx: These entries are _only_ for allowing containerd to pull images from the registry insecurely. e.g. after building a drupal image, pushing it to the local registry, and defining a deployment. The different private IPs are the ones that docker provides on our CI nodes. It's not the cleanest. :disappointed: 

The mirror is a good idea! I didn't add it initially since I was only testing in CI where we are logged in.

smlx:
Sorry about the issues running this locally. I didn't make it clear that this was only intended to the the minimum required to get CI going, which includes assuming that it is being run from Jenkins and that `testlagoon/<service>:$CI_BUILD_TAG` is the result of the current CI run. In fact `kind/test` locally won't currently test the code in the current working directory unless it's part of the Jenkins pipeline.

I'll work on improving the local development experience a bit.

Makefile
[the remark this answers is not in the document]
Schnitzel: >  The different private IPs are the ones that docker provides on our CI nodes. It's not the cleanest.

gotcha, maybe add a comment or so there, so this is more clearer why they are there?

Schnitzel:
> In any case I think this might be outside the scope of this PR, which is really just about getting CI running on main in k8s. Could we create an issue for OpenShift support in CI for Lagoon 2 and address this separately, since this PR is already quite large?

sure, created https://github.com/amazeeio/lagoon/issues/2382

> I'm not sure if this is worth adding logic to the Makefile for. Could we solve this by documenting the use of git clean -dfx when checking out a new major Lagoon release?

yea makes sense

> k3d vs kind

I guess its okay if we leave k3d in here for now and accept that we have k3d and kind at the same time.
But I think in the end we should really only have one of them, I'm okay if its kind if we say that k3d is not actually a real real replication of what we have running in production, even though that k3d starts much faster than kind :) But I think it's important that we have the same local development experience as also is used in the CI

Makefile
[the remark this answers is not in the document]
smlx: I figured out how to calculate this instead of hard-coding IPs.

Makefile
smlx: Okay, so this now uses the installed harbor registry to push/pull images. It does seem to be faster for iterative local development in my testing. Could you try it?

Makefile
smlx: Got this working and also created https://github.com/itchyny/gojq/issues/49

Makefile
smlx: I added some more instructions here for local development.

smlx:
>  But I think in the end we should really only have one of them, I'm okay if its kind if we say that k3d is not actually a real real replication of what we have running in production, even though that k3d starts much faster than kind :) But I think it's important that we have the same local development experience as also is used in the CI

The reason I went with kind in charts testing is that the helm CI tooling is built around it. I don't know much about k3d and helm chart testing.

smlx:
Ready for review again @tobybellwood @Schnitzel

smlx:
Just noting here: because we now `docker push` into the registry running on `kind`, the IP range needs to be configured as insecure in the local docker daemon. On linux, containers are in the `172.16.0.0/12` range, but on macOS it might be `192.168.0.0/16`? Not sure where that information should go..

tobybellwood (approved):
nice - works in Jenkins and easy to run locally!

PR #2374, opened 2020-12-03
Only monitor first route/ingress

### LINKED ISSUE #2373: ICM annotation improvements
Improve the annotation addition for the IngressMonitorController so that it only enables monitoring for the first (the LAGOON_ROUTE) route of Standard Availability projects and enables the ICM on all routes of High Availability Projects.

Schnitzel:
instead of enabling monitoring for all custom routes/ingress only enable for the first one
also fixes missing monitoring for active/standby routes

# Closing issues
closes #2373


twardnw:
This only takes care of making sure just the first route is monitored, do we want to address the monitoring of all routes in HA projects in a separate ticket?

Schnitzel:
> This only takes care of making sure just the first route is monitored, do we want to address the monitoring of all routes in HA projects in a separate ticket?

ah yes sorry should have mentioned that. Currently the build pod has no idea about HA vs SA, so we first need to get that into the build pod and then can do specific things based on that :)

twardnw:
Availability is passed in
https://github.com/amazeeio/lagoon/blob/main/node-packages/commons/src/api.ts#L994
https://github.com/amazeeio/lagoon/blob/main/node-packages/commons/src/tasks.ts#L316
We already use it for assigning the correct alertContact

Schnitzel:
@twardnw  correct, but it's not inside the oc-build-deploy-dind container yet where we need it, so I felt let's do this one here easily and get it out. And then add more later. Do you think it's a huge issue if we don't monitor every route of an HA site?

twardnw (approved):

PR #2385, opened 2020-12-08
Use the lastTransitionTime for failed kubernetes build jobs

### LINKED ISSUE #2046: Deployment duration does not stop incrementing for failed kubernetes deployments
**Describe the bug**
Deployment duration does not stop incrementing for failed kubernetes deployments.

**Screenshots**
image
Wait a minute and refresh
image

shreddedbacon:
# Checklist

# Closing issues
closes #2046

### DIFF
--- a/services/kubernetesbuilddeploymonitor/src/index.ts
+++ b/services/kubernetesbuilddeploymonitor/src/index.ts
@@ -181,17 +181,18 @@ ${podLog}`;
       }
     })(jobInfo.status.active ? 'active' : jobInfo.status.conditions[0].type.toLowerCase());
 
+    let completedTime = jobInfo.status.completionTime
     if (status == 'failed') {
       // failed jobs in kubernetes don't have a completionTime in them
       // the status conditions will have a lastTransitionTime in them that will contain the state the job went into
       // since jobs only have 1 attempt, there shouldn't be any other conditions
-      jobInfo.status.completionTime = jobInfo.status.conditions[0].lastTransitionTime
+      completedTime = jobInfo.status.conditions[0].lastTransitionTime
     }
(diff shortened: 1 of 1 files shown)


services/kubernetesbuilddeploymonitor/src/index.ts
```diff
@@ -181,6 +181,13 @@ ${podLog}`;
       }
     })(jobInfo.status.active ? 'active' : jobInfo.status.conditions[0].type.toLowerCase());
 
+    if (status == 'failed') {
+      // failed jobs in kubernetes don't have a completionTime in them
+      // the status conditions will have a lastTransitionTime in them that will contain the state the job went into
+      // since jobs only have 1 attempt, there shouldn't be any other conditions
+      jobInfo.status.completionTime = jobInfo.status.conditions[0].lastTransitionTime
```
rocketeerbkw: This mutates the `jobInfo` which is considered bad practice in JS these days. It might cause some poor, unsuspecting, future developer (karma says it's you 😉) to check this field at some point later on in this file and assume something that isn't true.

It'd be better to create a new variable that's set with the `jobInfo.status.completionTime` or `jobInfo.status.conditions[0].lastTransitionTime` to use in the `updateDeployment`. Similar to `status` right above (but doesn't need a fancy anonymous function).

rocketeerbkw (changes requested):

services/kubernetesbuilddeploymonitor/src/index.ts
shreddedbacon: Sure, easy fix

rocketeerbkw (approved):

PR #2408, opened 2020-12-18
Set an error so active-standby task retries

shreddedbacon:
# Checklist

Minor fix to the active/standby task image so that it actually tries to get the resource until it has a status condition.

Will need to update the tests to also ensure that after we run the active/standby, in addition to checking the routes actually moved, we need to check that the API actually updates with the production/standbyProduction environments changing.

# Closing issues


shreddedbacon:
I'll update the tests next week, just wanted to get a draft PR in to outline the issue.

PR #2403, opened 2020-12-15
Add initial support for mongodb-dbaas

shreddedbacon:
# Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for changelog and subsystem label(s) applied

Explain the **details** for making this change. What existing problem does the pull request solve?

# Closing issues

Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

PR #2416, opened 2020-12-22
Add Mongodb DBAAS operator (PR2403 & PR2411)

tobybellwood:
Covers PRs:
#2403 and #2411


tobybellwood:
(need to make sure it actually tests the node-MongoDB test :facepalm:) - this will require a MongoDB provider in lagoon-charts.

smlx:
with https://github.com/uselagoon/lagoon-charts/pull/201 this is now unblocked

PR #2417, opened 2020-12-22
Use the latest build-deploy controller and update active-standby task to use patch

shreddedbacon:
# Checklist

This brings the lagoon build-deploy controller up to be the latest version with fixes for using the `lagoon-deployer` token to create task pods, and also updates the active-standby task to use a patch operation instead of update.


tobybellwood:
@shreddedbacon - we need this against `main` too, don't we?

shreddedbacon:
@tobybellwood yes, probably. ~But main also needs a bunch of other stuff still...~ Edit: Maybe it is all there actually

PR #2436, opened 2020-12-31
Auto cron schedules and configurable backup retention

cdchris12:
# Checklist

This PR adds support for better backup pruning via the k8up operator's new smart scheduling functionality. Users can now also specify the specific timeframes they would like their production backups to be retained for in their `.lagoon.yml` files as well.

# Closing issues


Schnitzel:
discussed in standup:
- add tags system `production` and `development` to backup, check and prune schedules
- rename yaml key to `productionBackupSchedule`

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,16 +941,71 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+    if containsElement "daily" "${LAGOON_BACKUP_RETENTION_YAML_KEYS[@]}"; then
+      PRODUCTION_DAILY_BACKUP_RETENTION=$(cat .lagoon.yml | shyaml keys production-backup-retention.daily -q)
+    else
+      PRODUCTION_DAILY_BACKUP_RETENTION=""
+    fi
+  else
+    PRODUCTION_MONTHLY_BACKUP_RETENTION=""
+    PRODUCTION_WEEKLY_BACKUP_RETENTION=""
+    PRODUCTION_DAILY_BACKUP_RETENTION=""
+  fi
```
Schnitzel: shyaml already has capability to handle non existing keys, see https://github.com/amazeeio/lagoon/blob/36cc7bda3a5b09fc63d19e2841cda8ebcd523ca6/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L652-L658

so  you can just do:

```
    PRODUCTION_MONTHLY_BACKUP_RETENTION=$(cat .lagoon.yml | shyaml get-value production-backup-retention.monthly "")
    PRODUCTION_WEEKLY_BACKUP_RETENTION=$(cat .lagoon.yml | shyaml get-value production-backup-retention.weekly "")
    PRODUCTION_DAILY_BACKUP_RETENTION=$(cat .lagoon.yml | shyaml get-value production-backup-retention.daily "")
```

also there is `LAGOON_YAML_BACKUP_RETENTION_KEYS` created, but never used?

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,16 +941,71 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+
+  # Pull in environment type (development/production)
+  LAGOON_ENVIRONMENT_TYPE=${ENVIRONMENT_TYPE}
```
Schnitzel: why define a new variable if there is already one? 
let's just use `ENVIRONMENT_TYPE` directly

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,16 +941,71 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+  TEMPLATE_PARAMETERS+=(-p LAGOON_ENVIRONMENT_TYPE="${ENVIRONMENT_TYPE}")
+
+  # Set template parameters for retention values (prefer .lagoon.yml values over supplied defaults after ensuring they are valid integers)
+  if [ ! -z $PRODUCTION_MONTHLY_BACKUP_RETENTION ] && [ "$PRODUCTION_MONTHLY_BACKUP_RETENTION" -eq "$PRODUCTION_MONTHLY_BACKUP_RETENTION" ] && [ $LAGOON_ENVIRONMENT_TYPE = 'production'] 2>/dev/null; then
```
Schnitzel: why do we do this here?
```
[ "$PRODUCTION_MONTHLY_BACKUP_RETENTION" -eq "$PRODUCTION_MONTHLY_BACKUP_RETENTION" ]
```
we compare the same to the same?

also the `2>/dev/null` probably is not needed? as you don't run any commands that have output?

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,16 +941,71 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+    TEMPLATE_PARAMETERS+=(-p MONTHLY_BACKUP_RETENTION="${PRODUCTION_MONTHLY_BACKUP_RETENTION}")
```
Schnitzel: `TEMPLATE_PARAMETERS` is a system of `oc` in k8s deployments we use helm, therefore we don't need to define any `TEMPLATE_PARAMETERS`

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,16 +941,71 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+    _MONTHLY_BACKUP_RETENTION=${PRODUCTION_MONTHLY_BACKUP_RETENTION}
+  else
+    TEMPLATE_PARAMETERS+=(-p MONTHLY_BACKUP_RETENTION="${LAGOON_MONTHLY_BACKUP_RETENTION}")
```
Schnitzel: `LAGOON_MONTHLY_BACKUP_RETENTION` is never defined, can't we just use `MONTHLY_BACKUP_DEFAULT_RETENTION` ?

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,16 +941,71 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+  fi
+
+  # Pull in environment type (development/production)
+  LAGOON_ENVIRONMENT_TYPE=${ENVIRONMENT_TYPE}
+  TEMPLATE_PARAMETERS+=(-p LAGOON_ENVIRONMENT_TYPE="${ENVIRONMENT_TYPE}")
+
+  # Set template parameters for retention values (prefer .lagoon.yml values over supplied defaults after ensuring they are valid integers)
+  if [ ! -z $PRODUCTION_MONTHLY_BACKUP_RETENTION ] && [ "$PRODUCTION_MONTHLY_BACKUP_RETENTION" -eq "$PRODUCTION_MONTHLY_BACKUP_RETENTION" ] && [ $LAGOON_ENVIRONMENT_TYPE = 'production'] 2>/dev/null; then
+    TEMPLATE_PARAMETERS+=(-p MONTHLY_BACKUP_RETENTION="${PRODUCTION_MONTHLY_BACKUP_RETENTION}")
+    _MONTHLY_BACKUP_RETENTION=${PRODUCTION_MONTHLY_BACKUP_RETENTION}
```
Schnitzel: let's just use `MONTHLY_BACKUP_RETENTION` (without the underline), the underline suggests that something is special with this variable, but I don't really see anything?

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,16 +941,71 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+
+  # Let the controller decide when to run backups
+  BACKUP_SCHEDULE="@daily-random"
   TEMPLATE_PARAMETERS+=(-p BACKUP_SCHEDULE="${BACKUP_SCHEDULE}")
-  # TODO: -p == --set in helm
-  # Run Checks on Sunday at 0300-0600
-  CHECK_SCHEDULE=$( /kubectl-build-deploy/scripts/convert-crontab.sh "${NAMESPACE}" "M H(3-6) * * 0")
+
+  # Let the controller decide when to run checks
+  CHECK_SCHEDULE="@weekly-random"
```
Schnitzel, line 963: can you document the difference between `@weekly-random` and `@weekly` ?

images/kubectl-build-deploy-dind/helmcharts/k8up-schedule/templates/schedule.yaml
```diff
@@ -19,6 +19,7 @@ spec:
     schedule: '{{ .Values.check.schedule }}'
   prune:
     retention:
-      keepDaily: 7
-      keepWeekly: 6
+      keepDaily: '{{ .Values.prune.retention.keepDaily }}'
```
Schnitzel, line 22: for each value used in helm templates we should also define a default in the helmchart `values.yaml`, can you do that for these new variables?

services/kubernetesbuilddeploy/src/index.ts
```diff
@@ -19,6 +19,9 @@ const lagoonGitSafeBranch = process.env.LAGOON_GIT_SAFE_BRANCH || "master"
 const lagoonVersion = process.env.LAGOON_VERSION
 const overwriteKubectlBuildDeployDindImage = process.env.OVERWRITE_KUBECTL_BUILD_DEPLOY_DIND_IMAGE
 const registry = process.env.REGISTRY || "registry.lagoon.svc:5000"
+const monthlyBackupRetention = process.env.MONTHLY_BACKUP_DEFAULT_RETENTION || "1"
+const weeklyBackupRetention = process.env.WEEKLY_BACKUP_DEFAULT_RETENTION || "4"
```
Schnitzel: the current default is `6`, why are we changing this to 4?

services/kubernetesbuilddeploy/src/index.ts
```diff
@@ -19,6 +19,9 @@ const lagoonGitSafeBranch = process.env.LAGOON_GIT_SAFE_BRANCH || "master"
 const lagoonVersion = process.env.LAGOON_VERSION
 const overwriteKubectlBuildDeployDindImage = process.env.OVERWRITE_KUBECTL_BUILD_DEPLOY_DIND_IMAGE
 const registry = process.env.REGISTRY || "registry.lagoon.svc:5000"
+const monthlyBackupRetention = process.env.MONTHLY_BACKUP_DEFAULT_RETENTION || "1"
```
Schnitzel, line 22: currently we don't define any monthly backup retention, is that intentional changed to 1?

Schnitzel (changes requested):

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: I had originally built this because the call to `shyaml` was returning an error code if the specified value didn't exist, but I can no longer reproduce that error. I'll update the code accordingly.

About `LAGOON_YAML_BACKUP_RETENTION_KEYS`; this was the variable I assigned the array of keys in the `production-backup-retention` yaml structure to. I was using it on lines 949, 955, and 961 to set variables based on the existence of these keys. With the requested re-write here, I should be able to eliminate this altogether though.

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: Makes sense, will do.

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: Comparing the value to itself via the `-eq` operator is how we ensure this value is an integer, as the `-eq` operator will attempt to parse a string into an int and fail if such conversion is not possible. The output redirection just squashes the output from those operations if the conversion fails.

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: I thought the same thing, but I saw other functionality in this file which still retained the `TEMPLATE_PARAMETERS` system, so I figured it wouldn't hurt to include those bits here as well. Based on your comment, though, I'll go ahead and rip those out :)

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: Good catch; that's what we should be using here. Will update.

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
Schnitzel: ah, ok, but let's document this here then that the `-eq` is used to make sure it's an integer, as currently it looks like a bug :)

and please don't redirect errors into `/dev/null` unless you have a specific handling for it with a custom error message

services/kubernetesbuilddeploy/src/index.ts
cdchris12: Since we're keeping 6 weekly backups already, this monthly backup should line up with those weekly backups. I can change this to `0` easily enough, though , if you'd prefer?

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -841,16 +841,46 @@ if oc --insecure-skip-tls-verify -n ${OPENSHIFT_PROJECT} get schedules.backup.ap
...
-  BACKUP_SCHEDULE=$( /oc-build-deploy/scripts/convert-crontab.sh "${OPENSHIFT_PROJECT}" "M H(22-2) * * *")
+  # Pull in .lagoon.yml variables
+  PRODUCTION_MONTHLY_BACKUP_RETENTION=$(cat .lagoon.yml | shyaml get-value production-backup-retention.monthly "")
+  PRODUCTION_WEEKLY_BACKUP_RETENTION=$(cat .lagoon.yml | shyaml get-value production-backup-retention.weekly "")
+  PRODUCTION_DAILY_BACKUP_RETENTION=$(cat .lagoon.yml | shyaml get-value production-backup-retention.daily "")
+
+  # Pull in Lagoon variables
+  LAGOON_MONTHLY_BACKUP_RETENTION=${MONTHLY_BACKUP_DEFAULT_RETENTION}
+  LAGOON_WEEKLY_BACKUP_RETENTION=${WEEKLY_BACKUP_DEFAULT_RETENTION}
+  LAGOON_DAILY_BACKUP_RETENTION=${DAILY_BACKUP_DEFAULT_RETENTION}
```
Schnitzel: these are unused now.

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -841,16 +841,46 @@ if oc --insecure-skip-tls-verify -n ${OPENSHIFT_PROJECT} get schedules.backup.ap
...
+
+  # Pull in environment type (development/production)
+  TEMPLATE_PARAMETERS+=(-p ENVIRONMENT_TYPE="${ENVIRONMENT_TYPE}")
+
+  # Set template parameters for retention values (prefer .lagoon.yml values over supplied defaults after ensuring they are valid integers via "-eq" comparison)
+  if [ ! -z $PRODUCTION_MONTHLY_BACKUP_RETENTION ] && [ "$PRODUCTION_MONTHLY_BACKUP_RETENTION" -eq "$PRODUCTION_MONTHLY_BACKUP_RETENTION" ] && [ $ENVIRONMENT_TYPE = 'production']; then
+    TEMPLATE_PARAMETERS+=(-p MONTHLY_BACKUP_RETENTION="${PRODUCTION_MONTHLY_BACKUP_RETENTION}")
+  else
+    TEMPLATE_PARAMETERS+=(-p MONTHLY_BACKUP_RETENTION="${MONTHLY_BACKUP_RETENTION}")
```
Schnitzel: `MONTHLY_BACKUP_RETENTION` is not defined?

Schnitzel: this probably should be `MONTHLY_BACKUP_DEFAULT_RETENTION` ?

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -841,16 +841,46 @@ if oc --insecure-skip-tls-verify -n ${OPENSHIFT_PROJECT} get schedules.backup.ap
...
-  # Run Checks on Sunday at 0300-0600
-  CHECK_SCHEDULE=$( /oc-build-deploy/scripts/convert-crontab.sh "${OPENSHIFT_PROJECT}" "M H(3-6) * * 0")
+  # Let the controller deduplicate prunes (will run weekly at a random time throughout the week)
+  CHECK_SCHEDULE="@weekly-random"
   TEMPLATE_PARAMETERS+=(-p CHECK_SCHEDULE="${CHECK_SCHEDULE}")
 
-  # Run Prune on Saturday at 0300-0600
-  PRUNE_SCHEDULE=$( /oc-build-deploy/scripts/convert-crontab.sh "${OPENSHIFT_PROJECT}" "M H(3-6) * * 6")
+  # Let the controller deduplicate prunes (will run at 0000 on Sunday morning)
+  PRUNE_SCHEDULE="@weekly"
```
Schnitzel: does this mean all environments will prune at the same time? let's not do that :)

images/oc-build-deploy-dind/openshift-templates/backup-schedule.yml
```diff
@@ -32,17 +32,29 @@ parameters:
...
+    required: true
+  - name: WEEKLY_BACKUP_RETENTION
+    description: How many weekly backups should be retained after pruning
+    required: true
+  - name: DAILY_BACKUP_RETENTION
+    description: How many daily backups should be retained after pruning
+    required: true
+  - name: LAGOON_ENVIRONMENT_TYPE
+    description: What type of environment this is (production/development)
+    required: true
```
Schnitzel: I don't see us inject this as parameters anywhere?

services/kubernetesbuilddeploy/src/index.ts
Schnitzel: mhh okay, let's leave it at `1` then.

.lagoon.yml
```diff
@@ -39,6 +39,8 @@ tasks:
         command: es-curl PUT _cluster/settings -d '{"persistent":{"cluster.routing.allocation.enable":null}}'
         service: logs-db
 
+production-backup-retention:
+  monthly: 2
```
Schnitzel: do we need to define this here? we're not going to use lagoon to deploy lagoon in the future, so I think this can go?

docs/using_lagoon/lagoon_yml.md
```diff
@@ -41,6 +41,11 @@ tasks:
 routes:
   insecure: Redirect
 
+production-backup-retention:
+  monthly: 1
+  weekly: 4
+  daily: 7
+
```
Schnitzel: the longer I look at this I would suggest to change it to this:
```
backup-retention:
  production:
    monthly: 1
    weekly: 4
    daily: 7
```
feels much better if in the future we also want to allow to change the backup-retention for `development` projects? (we don't have to define this right now)

Schnitzel (changes requested):

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: Good eye! Will be removed

images/oc-build-deploy-dind/openshift-templates/backup-schedule.yml
cdchris12: Another good catch. Will fix

.lagoon.yml
cdchris12: I added this both for testing purposes as well as to provide a bit of context on how to use this functionality. It shouldn't hurt our testing setup to keep it in. :)

images/oc-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: Updated to `@weekly-random`, which will run this weekly at a random time

docs/using_lagoon/lagoon_yml.md
cdchris12: Updated the .lagoon.yml file definition and builddeploy code as requested.

.lagoon.yml
Schnitzel: we changed this now to be `backup-retention.production`

> added this both for testing purposes
mhh testing should happen before the production deployment of Lagoon and this is only is used when we deploy lagoon to production. If we realize then that this is broken, then it's too late ;)

> how to use this functionality.
Lagoon documentation happens in the lagoon docs and not in `.lagoon.yml` of Lagoon itself.
As we have proper documentation in the docs already, let's remove this here

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -933,25 +933,48 @@ if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/Schedule" ]]; then
...
+    WEEKLY_BACKUP_RETENTION=${WEEKLY_BACKUP_DEFAULT_RETENTION}
+  fi
+  if [ ! -z $PRODUCTION_DAILY_BACKUP_RETENTION ] && [ "$PRODUCTION_DAILY_BACKUP_RETENTION" -eq "$PRODUCTION_DAILY_BACKUP_RETENTION" ] && [ $ENVIRONMENT_TYPE = 'production']; then
+    DAILY_BACKUP_RETENTION=${PRODUCTION_DAILY_BACKUP_RETENTION}
+  else
+    DAILY_BACKUP_RETENTION=${DAILY_BACKUP_DEFAULT_RETENTION}
+  fi
+
+  # Let the controller decide when to run backups (will run daily at a random time throughout the day)
+  BACKUP_SCHEDULE="@daily-random"
```
Schnitzel: we don't want to change the backup schedule system. Having `@daily-random` means that backups could happen during the day which could cause production websites to go down.
Let's keep the Lagoon schedule system that automatically does the backups during the night.

Schnitzel (changes requested):

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
cdchris12: Agreed. It'll be easy enough to add support for whatever framework the controller requires once that functionality is rolled out.

Schnitzel (approved):

PR #2216, opened 2020-09-23
Changes to support Openshift in the Lagoon Build and Deploy Controllers

shreddedbacon:
# Checklist
- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for changelog and subsystem label(s) applied

This contains the initial changes to Lagoon to support using the Lagoon Build and Deploy Controllers for Openshift deployments. There are additional changes that are in the controllers themselves to also support Openshift, see the builddeploy controllers repository for more.

This PR also adds a new activeSystemsX values `lagoon_osController<Job|Misc|BuildDeploy|Remove>` to allow for transitioning existing projects to using the controller as required.

# Closing issues
Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

PR #2437, opened 2021-01-04
Reenable webhooks, and webhook tests

tobybellwood:
This PR ties to a branch of lagoon-charts that has the webhook services re-enabled.

It disables the check-branch-sha.yaml as the pullrequest type doesn't have a LAGOON_GIT_BRANCH to check for.

It also adds in the webhooks components from #2216 to enable the removal and deletion of PR and branches from k8s.


tobybellwood:
PR has been updated to create an ansible var file that is populated in the `tests` image entrypoint with the values passed from the lagoon-test secret.  These extra-vars are then called by the ansible-playbook to replace  (amongst others) - the graphql URL, API URL, webhook URLs and git URLs.  Once the corresponding PR has been merged to uselagoon/lagoon-charts, the `CHARTS_TREEISH` branch in the makefile can be set back to main again.

PR #2440, opened 2021-01-04
fix dist directory created by root:

Schnitzel:
- with creating a volume inside the container the files are never mounted into the host
- .gitkeep makes sure the folder already exists

### DIFF
--- a/node-packages/commons/.gitignore
+++ b/node-packages/commons/.gitignore
@@ -1,2 +1,3 @@
 node_modules
-dist
+dist/*
+!dist/.gitkeep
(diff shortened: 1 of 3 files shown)


node-packages/commons/.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
-dist
```
smlx, line 2: You can also `git add -f dist/.gitkeep` to avoid the need for this change.

node-packages/commons/.gitignore
```diff
@@ -1,2 +1,3 @@
...
+dist/*
+!dist/.gitkeep
```
smlx, line 3: You can also `git add -f dist/.gitkeep` to avoid the need for this change.

PR #2442, opened 2021-01-05
Simplify makefile targets

smlx:
# Checklist

I noticied that one of the `Makefile` targets was missing a `.PHONY` rule. I attempted to fix this by merging that target into another. The alternative is to just add a `.PHONY` rule. @Schnitzel which do you prefer?

This is macOS specific, so I'll need someone on that platform to review.

# Closing issues

n/a


Schnitzel:
looks good to me :)

PR #2443, opened 2021-01-05
Add more error checking and resilience to bitbucket sync script

### LINKED ISSUE #2234: api bitbucket-sync fails if user is deleted or deactivated inside bitbucket
it looks like that it's possible that users are deleted or deactivated within bitbucket, which then causes our sync script to throw an error and fail completely, the error is:
```
        "(node:1090) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'toLowerCase' of undefined",
        "    at /app/services/api/dist/bitbucket-sync/repo-permissions.js:77:47",
        "    at Generator.next (<anonymous>)",
        "    at fulfilled (/app/services/api/dist/bitbucket-sync/repo-permissions.js:5:58)",
        "    at process._tickCallback (internal/process/next_tick.js:68:7)",
        "(node:1090) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)",
        "(node:1090) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.",
        "\u001b[2K\u001b[1GDone in 231.62s."
```

and seems to happen here:
https://github.com/amazeeio/lagoon/blob/a5f505fc42cc2eebf0cbcb450bdad9dbd2826ad6/services/api/src/bitbucket-sync/repo-permissions.ts#L86

it stops the whole sync script completely, so if the error happens in a project that is synced at the very beginning all other projects are not synced anymore.

I think we should improve the error handling of the script:
- if a user cannot be resolved or any other error happens for one user, continue with the other users
- if a problem for one project happens, continue with the other projects

rocketeerbkw:
# Checklist

There were some edge cases that cause breaks in the bitbucket sync script. This fixes those errors and adds more checks to not cause a break if something similar were to happen again.

# Closing issues

closes #2234

### DIFF
--- a/node-packages/commons/src/bitbucketApi.ts
+++ b/node-packages/commons/src/bitbucketApi.ts
@@ -1,6 +1,34 @@
 import axios from 'axios';
 import { find, pathOr, propEq, propOr } from 'ramda';
 
+interface ApiResourceLink {
+  href: string;
+  name: string;
+}
+
+// https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm8296923984
+interface ApiProject {
+  key: string;
+  id: number;
+  name: string;
+  description: string;
+  public: boolean
+  links: {
+    [key: string]: ApiResourceLink[]
+  };
+}
+
+// https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm8296923984
+interface ApiRepo {
+  slug: string;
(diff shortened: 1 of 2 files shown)


tobybellwood:
merged in #2494

PR #2457, opened 2021-01-15
Fastly controller support

### LINKED ISSUE #2335: inject Fastly annotations if Fastly operator is present on the cluster
We have the Fastly Operator now available on k8s clusters. When it is present, all client domains should route via the CDN, so we need to inject the necessary annotations to ingresses for this.

`fastly.amazee.io/watch`
`fastly.amazee.io/service-id`

We would need to store the service-id along with the cluster information in the API. Also these settings need to be able to be overridden at the project or environment level too, pulling in a customer provided api-secret-key from a BUILDTIME env var

shreddedbacon:
# Checklist

Add support for fastly controller annotations to lagoon builds

# Closing issues

closes #2335

### DIFF
--- a/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
+++ b/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
@@ -687,17 +687,70 @@ if [ -n "$(cat .lagoon.yml | shyaml keys fastly.api-secrets.$FASTLY_API_SECRETS_
       echo -e "A fastly api secret was defined in the .lagoon.yml file, but no platform tls configuration id could be found in the .lagoon.yml\n\nPlease check if the platform tls configuration id has been set correctly."
       exit 1
     fi
-    helm template ${FASTLY_API_SECRET_NAME} \
-      kubectl-build-deploy/helmcharts/fastly-api-secret \
-      --set fastly.apiToken="${FASTLY_API_TOKEN}" \
-      --set fastly.platformTLSConfiguration="${FASTLY_API_PLATFORMTLS_CONFIGURATION}" \
-      -f /kubectl-build-deploy/values.yaml "${HELM_ARGUMENTS[@]}" > $YAML_FOLDER/00-${FASTLY_API_SECRET_NAME}.yaml
-      ## this api secret needs to exist before the ingress is created, so try prioritise it by putting it numerically ahead of any ingresses
-    FASTLY_API_SECRETS+=(${FASTLY_API_SECRET_NAME})
+
+    # run the script to create the secrets
+    . /kubectl-build-deploy/scripts/exec-fastly-api-secrets.sh
+
     let FASTLY_API_SECRETS_COUNTER=FASTLY_API_SECRETS_COUNTER+1
   done
 fi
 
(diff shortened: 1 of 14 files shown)


Schnitzel:
@shreddedbacon 
nice work, looks good so far, can you create also some documentation entries, as I'm a bit confused what exactly we would expect from the customers to do here  :) 

also as we want fastly to be fully transparent to the customers I wonder what we could do on maybe a lagoon/cluster/project level that fastly support can automaticaly be enabled?

shreddedbacon:
> @shreddedbacon
> nice work, looks good so far, can you create also some documentation entries, as I'm a bit confused what exactly we would expect from the customers to do here :)

Yeah, it is still a WIP :) I'll be working on the documentation and tests today.

> also as we want fastly to be fully transparent to the customers I wonder what we could do on maybe a lagoon/cluster/project level that fastly support can automaticaly be enabled?

I'm trying to figure out how best to do this, I think it will have to be something injected by the controller some how to be able to be cluster specific. But it could also be good to include the fastly service ID/platformtls ID in the `Openshift|Kubernetes` resource like we do with `consoleUrl` etc.. and keep all the logic in Lagoon core with the ability to override using variables per project/environment like other things.

Schnitzel:
> I think it will have to be something injected by the controller some how to be able to be cluster specific. But it could also be good to include the fastly service ID/platformtls ID in the Openshift|Kubernetes resource like we do with consoleUrl etc.. and keep all the logic in Lagoon core with the ability to override using variables per project/environment like other things.

I think we would also need an overwrite per Lagoon `Group`. As we have customers that have multiple projects in the same Fastly Service, but we don't want to share the Fastly Service across customers (we do it currently for the no-cache service, but in the future we want to have fastly configs which do cache and therefore we need individual service so that the customers can purge their fastly caches).
This could be solved via https://github.com/amazeeio/lagoon/pull/2457 which would allow a group to add env variables overwrites, which then could be used to overwrite the default fastly service though.

shreddedbacon:
Referencing this PR on the controller https://github.com/amazeeio/lagoon-kbd/pull/32 which will use `LAGOON_FASTLY_NOCACHE_SERVICE_ID` to inject into all builds started by the controller.

Adding support for this variable into the build script at the moment

shreddedbacon:
I still need to write the tests for this to make sure that the annotations get added correctly for the ingress objects. As there are so many ways to define overrides, we should definitely test these.

I've added some documentation in code around the different override solutions, but really this should not be attempted by anyone without talking to someone that understands Fastly first. This is mainly due to how Fastly itself works (not the controller), as changing a domains service ID is not really supported without first removing domains from the existing service.

But this does now also add support for consuming `LAGOON_FASTLY_NOCACHE_SERVICE_ID` that the lagoon build deploy controller will inject once pr https://github.com/amazeeio/lagoon-kbd/pull/32 is merged into the controllers and released.

shreddedbacon:
Tests have failed on Openshift, but this PR only adds fastly support to kubectl-builddeploy.

Flaky Openshift tests?

Schnitzel:
@shreddedbacon 
merged (as I wantt his in 1.13.2 that needs to go out today).

Can you create a documentation either in docs.lagoon.sh or in our internal confluence on how to use this?

PR #2459, opened 2021-01-15
Allow historical storage data to be recorded - For use during migrations

### LINKED ISSUE #2451: Allow historical storage data to be recorded for environments
Currently the `addOrUpdateEnvironmentStorage` mutation only accepts the storage data to record, and it hardcodes the date to be the time of submission. When moving projects between clusters, we need a way to move historical storage data also.

Two possible solutions:
1. Update the existing mutation to accept a date (it can be optional for backwards compatibility)
2. Add a new mutation specifically for this use case

The advantage to (1) is simplicity. The advantage to (2) is the API client wouldn't need to load all data, then loop through it all and call a bunch of API calls.

justinlevi:
# Checklist

This PR adds the ability to allow for historical storage data to be recorded. This is useful during migrations. Adds an additional `updated` field on the mutation. 

```
mutation {
  addOrUpdateEnvironmentStorage(input:{
    environment:20013, 
    persistentStorageClaim:"BananaSmoothiesRule", 
    bytesUsed:1234,
    updated:"2020-12-10"
  }){
    id
    environment{
      name
    }
    persistentStorageClaim
    bytesUsed
    updated
  }
  ```

# Closing issues

Put `closes #2451`

### DIFF
--- a/services/api/src/resources/environment/resolvers.ts
+++ b/services/api/src/resources/environment/resolvers.ts
@@ -3,6 +3,7 @@ import { sendToLagoonLogs } from '@lagoon/commons/dist/logs';
 import { createRemoveTask } from '@lagoon/commons/dist/tasks';
 import { ResolverFn } from '../';
 import { isPatchEmpty, prepare, query, whereAnd } from '../../util/db';
+import convertDateToMYSQLDateTimeFormat from '../../util/convertDateToMYSQLDateTimeFormat'
 import { Helpers } from './helpers';
 import { Sql } from './sql';
 import { Sql as projectSql } from '../project/sql';
@@ -344,10 +345,15 @@ export const addOrUpdateEnvironment: ResolverFn = async (
 
 export const addOrUpdateEnvironmentStorage: ResolverFn = async (
   root,
-  { input },
+  { input: unformattedInput },
   { sqlClient, hasPermission },
 ) => {
   await hasPermission('environment', 'storage');
+
+  const input = {
+    ...unformattedInput,
+    updated: unformattedInput.updated ? unformattedInput.updated: convertDateToMYSQLDateTimeFormat(new Date().toISOString())
+  };
 
   const prep = prepare(
     sqlClient,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: services/api/src/resources/environment/resolvers.ts [hunk 1]
```diff
@@ -344,21 +344,30 @@ export const addOrUpdateEnvironment: ResolverFn = async (
...
       CALL CreateOrUpdateEnvironmentStorage(
         :environment,
         :persistent_storage_claim,
-        :bytes_used
+        :bytes_used,
+        :updated
       );
     `,
   );
+  console.log("🚀 ~ file: resolvers.ts ~ line 368 ~ prep", prep(input))
```
Schnitzel: let' remove that :)

services/api/src/resources/environment/resolvers.ts
```diff
@@ -344,21 +344,30 @@ export const addOrUpdateEnvironment: ResolverFn = async (
...
 export const addOrUpdateEnvironmentStorage: ResolverFn = async (
   root,
-  { input },
+  { input: unformattedInput },
   { sqlClient, hasPermission },
 ) => {
   await hasPermission('environment', 'storage');
 
+  const d = new Date();
+  const now = d.toISOString().split('T')[0]+' ' + d.toTimeString().split(' ')[0];
```
Schnitzel: this seems very hacky, is that the correct way to generate a specific formatted string?

Schnitzel (changes requested):

services/api/src/resources/environment/resolvers.ts
justinlevi: fixed

services/api/src/resources/environment/resolvers.ts
justinlevi: Yeah, you're probably right. I've updated and reused a formatter function we use elsewhere.

Schnitzel (approved):

rocketeerbkw (approved):

PR #2481, opened 2021-01-25
Implement rootless workloads

smlx:
# Checklist

- [x] Documentation has been written/updated. See uselagoon/lagoon#2541.

# Description

This PR implements rootless workloads on Lagoon, disabled by default, and controlled via feature flags as documented in uselagoon/lagoon#2541.

For Openshift 3.11 compatibility we run as a non-root uid, root gid, and non-root `fsGroup`. This also means that no changes are required in [lagoon-images](https://github.com/uselagoon/lagoon-images).

The changes made include:
* add a non-root `securityContext` to all pods.
* add an `initContainer` to the shared storage nginx pod. This will update the ownership of the shared storage volume to the correct permissions going forward. See [this comment below](https://github.com/amazeeio/lagoon/pull/2481#issuecomment-774070068) for details.

This depends on https://github.com/uselagoon/lagoon-charts/pull/235. Once that lands, we can remove https://github.com/amazeeio/lagoon/pull/2481/commits/4a4fa5d06f50c5e953a9938ac63a8efabb29b9dd from this PR.

# Closing issues

Partially addresses https://github.com/amazeeio/lagoon/issues/2059.


smlx:
Not sure what milestone you want to stick this into @tobybellwood?

tobybellwood:
I think it's the last component of the 2.0 release?

Schnitzel:
discussed in the AIO security meeting:
- implementing this will have a big impact on all running workload today, therefore we don't think it will make it into 2.1 as we have to do a lot of testing
- looking at openshift they leave the user running as group root, we should probably do the same, as in some cases the uid cannot be changed in the PVC during mounting. Therefore we don't want to add another group and leave it in root group.

smlx:
Looking at this a bit more, I think we might be able to keep the group set to non-root, and add root as a supplementary group on openshift. I'll have to do some more research and testing.

smlx:
## Migrating permissions of RWX volumes from some common storage volume provisioners

Given the need to migrate existing workloads, I tried to understand how our current RWX storage provisioners matched permissions between the pod and the storage volume to enable access.

Here's what you get when mounting a PVC for a RWX volume with the new non-root pod securityContext from this PR:

```
fsGroup: 10001
runAsGroup: 0
runAsUser: 10000
```

### nfs-provisioner

`kind` running locally with `nfs-provisioner` (now deprecated), default configuration, RWX volumes are given 777 permissions, with `setgid`.

```
$  kubectl exec -it nonroot-7f977db68d-hlx2j -- sh -c 'id && stat -c "%A %u %g" /storage'
uid=10000 gid=0(root) groups=10001
drwxrwsrwx 0 0
```

### AWS efs-provisioner

AWS with `efs-provisioner` (also now deprecated!), and our current configuration, RWX volumes are assigned a unique group ID.
This is added as a supplementary group ID to the pod, and increments for a unique value on each provisioned PV. See [here](https://github.com/kubernetes-retired/external-storage/tree/master/aws/efs#parameters) for details.

```
$ kubectl exec -it nonroot-5f5ff9fd8c-cql88 -- sh -c 'id && stat -c "%A %u %g" /storage'
uid=10000 gid=0(root) groups=10001,40003
drwxrws--x 0 40003
```

### Azure file

AKS using the azure-file storageclass, RWX volumes are given 777 permissions and also the group ID given by `fsGroup`.

```
$ kubectl exec -it nonroot-5f5ff9fd8c-fgg5g -- sh -c 'id && stat -c "%A %u %g" /storage'
uid=10000 gid=0(root) groups=10001
drwxrwxrwx 0 10001
```

### GKE nfs-client-provisioner

GKE using the nfs-client-provisioner, RWX volumes are simply given 777 permissions.

```
$ kubectl exec -it nonroot-5fc7554c49-kks8r -- sh -c 'id && stat -c "%A %u %g" /storage'
uid=10000 gid=0(root) groups=10001
drwxrwxrwx 0 0
```

## Discussion

Amazingly, _every_ RWX provisioner I checked uses a different mechanism to give the pod access to the storage.

Since existing workloads are running as root we have the problem that e.g. `rsync` has created files inside the storage volumes with permissions such as:

```
$ stat -c "%A %u %g(%G)" example0.jpg
-rw-rw-r-- 0 82(www-data)
$ stat -c "%A %u %g(%G)" example1.jpg
-rw-rw---- 0 82(www-data)
```

This is a problem, since `example0.jpg` will be read-only and `example1.jpg` not accessible at all to the new non-root pod securityContext.

So the solution I've come up with is to recursively set group ownership to match the mountpoint, and to give group access permissions. This is basically the same thing the `fix-permissions` script does.

This will:

1. Match the permissions set by all the above StorageClass drivers on the root of the volume.
2. Give the new non-root pod securityContext `g+rw` access to all files, and `g+x` on directories.

There is one circumstance I can see where this logic will break, and that is where the volume has been `chgrp`'d to something other than the StorageClass default. This is pretty unlikely though IMO and we can deal with it on a case-by-case basis.

The other problem with this approach is that the initContainer will run on every pod start, which is a lot of overhead for a process which only needs to run once. To mitigate this I'm creating a sentinel file inside the storage volume, which is checked before doing anything in the initContainer. This should mean that the permission change is only done approximately once.

```
set -e
SENTINEL="/storage/.lagoon-rootless-migration-complete"
if ! [ -f "$SENTINEL" ]; then
  find /storage -mindepth 1 -exec chgrp $(stat -c "%g" /storage) {} +
  find /storage -mindepth 1 -exec chmod g+rw {} +
  find /storage -mindepth 1 -type d -exec chmod g+x {} +
fi
touch "$SENTINEL"
```

I've manually tested this on EKS and GKE by creating files with broken permissions and letting the initContainer fix them.

The Azure file provisioner enforces permissions by ignoring all chown/chmod requests, so the initContainer isn't strictly required:

```
# mkdir -p /storage/blah; echo blah > /storage/blah/blah; chown -hR 0:82 /storage/blah; chmod -R o= /storage/blah; ls -lahR /storage
/storage:
total 4K     
drwxrwxrwx    2 root     10001          0 Feb  5 10:54 .
drwxr-xr-x    1 root     root        4.0K Feb  5 13:52 ..
drwxrwxrwx    2 root     10001          0 Feb  5 13:52 blah

/storage/blah:
total 1K     
drwxrwxrwx    2 root     10001          0 Feb  5 13:52 .
drwxrwxrwx    2 root     10001          0 Feb  5 10:54 ..
-rwxrwxrwx    1 root     10001          5 Feb  5 13:52 blah
```

## Future work

Once all workloads are running with the new pod securityContext, the initContainer can be removed.

smlx:
After some discussion, this needs a feature flag via environment variable for initial rollout.

smlx:
~Here's how the rootless rollout might progress for existing workloads:~

~1. Merge this PR for rootless workload support, disabled by default.
2. Gradually migrate workloads using the `LAGOON_ROOTLESS_WORKLOAD` build-time env var.
3. Once all workloads are migrated, set `rootless: true` by default, and remove the `LAGOON_ROOTLESS_WORKLOAD` logic.
4. Remove `LAGOON_ROOTLESS_WORKLOAD` env var from all projects.~

edit: that's all wrong. I think we can automate this.

smlx:
New plan for rollout: do some automation that clones running workloads, runs them rootless, and performs diagnostics to identify any problems.

smlx:
As this feature can be controlled at the cluster, project, and environment level, I'm setting this as ready for review.

Automated testing for projects/environments to assess if they can run rootless with no changes will be left for a future PR.

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -25,6 +25,52 @@ function contains() {
     [[ $1 =~ (^|[[:space:]])$2($|[[:space:]]) ]] && return 0 || return 1
 }
 
+# featureFlag searches for feature flag variables in the following locations
+# and order:
+#
+# 1. The cluster-force feature flag, prefixed with LAGOON_FEATURE_FLAG_FORCE_,
+#    in the build environment. This is set via a flag on the build-deploy
```
Schnitzel: this sounded like the `lagoon build environments` I would suggest:
```
as a build pod environment variable
```

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -25,6 +25,52 @@ function contains() {
...
+#    controller. This overrides the other variables and allows policy
+#    enforcement at the cluster level.
+#
+# 2. The regular feature flag, prefixed with LAGOON_FEATURE_FLAG_, in the
+#    Lagoon environment env-vars. This allows policy control at the environment
```
Schnitzel: let's just use `global` scope and document it here

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -557,6 +603,12 @@ yq write -i -- /kubectl-build-deploy/values.yaml 'routesAutogenerateShortSuffix'
 for i in $ROUTES_AUTOGENERATE_PREFIXES; do yq write -i -- /kubectl-build-deploy/values.yaml 'routesAutogeneratePrefixes[+]' $i; done
 yq write -i -- /kubectl-build-deploy/values.yaml 'kubernetes' $KUBERNETES
 yq write -i -- /kubectl-build-deploy/values.yaml 'lagoonVersion' $LAGOON_VERSION
+# check for ROOTLESS_WORKLOAD feature flag, disabled by default
+if [ "$(featureFlag ROOTLESS_WORKLOAD)" = enabled ]; then
+	yq write -i -- /kubectl-build-deploy/values.yaml 'rootless' true
+else
+	yq write -i -- /kubectl-build-deploy/values.yaml 'rootless' false
```
Schnitzel: outcome of discussion with @smlx 
in order to make this whole change more kubernetes native we change it to:

- remove the `rootless` variable from helm values
- if rootless is disabled set 
```
podSecurityContext:
  fsGroup: 0
```
in the helm values to basically overwrite the default non-root values.

Schnitzel:
just for some reference for the future, here is the codechange since k8s 1.12 which changed the behavior to NOT change the persistent storage if it's an RWX mount:
https://github.com/kubernetes/kubernetes/pull/67280/files

images/kubectl-build-deploy-dind/helmcharts/nginx-php-persistent/templates/deployment.yaml
```diff
@@ -32,8 +32,33 @@ spec:
           emptyDir: {}
       priorityClassName: {{ include "nginx-php-persistent.lagoonPriority" . }}
       enableServiceLinks: false
+      {{- if .Values.rootless }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      # This initContainer sets permissions on the RWX storage compatible with
+      # the pod securityContext.
+      initContainers:
```
Schnitzel, line 41: put this in an `if` which only runs if `.Values.podSecurityContext.fsGroup != 0`

Schnitzel (changes requested):

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
smlx: updated the comment

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
smlx: updated the comment and logic

images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
smlx: I made this slightly different to the comment: now we don't make _any_ changes to `podSecurityContext`, and instead only set it to the rootless variables when rootless is enabled.

images/kubectl-build-deploy-dind/helmcharts/nginx-php-persistent/templates/deployment.yaml
smlx: added equivalent logic and an explanatory comment

smlx:
@Schnitzel @tobybellwood I cleaned this up a lot and simplified it to reduce the changes and hopefully make the logic much clearer. Let me know what you think.

PR #2486, opened 2021-01-27
K8p schedule quote fix

Schnitzel:
the v1.13.4 release brought a new way to inject the prune specs into the k8up schedule objects, unfortunately this is created with quotes and therefore it's interpreted as strings, while the k8up operator needs integers.
this PR fixes this


cdchris12 (approved):
LG2M

PR #2490, opened 2021-02-03
UI feature: Additional filters added to problems tab

### LINKED ISSUE #2489: UI: Provide extra filters to problems tab on environments
There have been many requests to improve the current filtering of problem results inside the 'Problem' tab on an environment.  

Some additional filtering options such as by severity, source and service would be ideal so we can better display available results.

image

timclifford:
# Checklist

- [x] Additional select filters for problems tab so we can improve results visibility

image

# Closing issues
closes #2489

### DIFF
--- a/services/ui/src/components/Filters/helpers.js
+++ b/services/ui/src/components/Filters/helpers.js
@@ -19,12 +19,6 @@ export const getSourceOptions = gql`
   }
 `;
 
-export const getServiceOptions = gql`
-  query getProblemServices($environment: Int) {
-    services: problemServices(environment: $environment)
-  }
-`;
-
 const getSeverityEnumQuery = gql`
   query severityEnum {
     __type(name: "ProblemSeverityRating") {
--- a/services/ui/src/components/Filters/helpers.js
+++ b/services/ui/src/components/Filters/helpers.js
@@ -19,6 +19,12 @@ export const getSourceOptions = gql`
   }
 `;
 
+export const getServiceOptions = gql`
+  query getProblemServices($environment: Int) {
+    services: problemServices(environment: $environment)
+  }
+`;
+
 const getSeverityEnumQuery = gql`
   query severityEnum {
     __type(name: "ProblemSeverityRating") {
--- a/services/ui/src/components/Problems/index.js
+++ b/services/ui/src/components/Problems/index.js
@@ -1,229 +1,370 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
+import { useQuery } from "@apollo/react-hooks";
+import getSeverityEnumQuery, {getProjectOptions, getSourceOptions, getServiceOptions} from 'components/Filters/helpers';
 import { bp, color, fontSize } from 'lib/variables';
 import useSortableProblemsData from './sortedItems';
 import { getFromNowTime } from "components/Dates";
 import Problem from "components/Problem";
+import SelectFilter from 'components/Filters';
 
-const Problems = ({ problems }) => {
-    const { sortedItems, requestSort, getClassNamesFor } = useSortableProblemsData(problems);
+const getOptionsFromProblems = (problems, key) => {
+  let uniqueOptions= problems && new Set(problems.map(p => {
+    return p[key];
+  }));
+  return [...uniqueOptions];
+};
 
-    const [problemTerm, setProblemTerm] = useState('');
-    const [hasFilter, setHasFilter] = React.useState(false);
+const Problems = ({problems, environment}) => {
+  const { sortedItems, requestSort, getClassNamesFor } = useSortableProblemsData(problems);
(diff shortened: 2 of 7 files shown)


services/ui/src/components/Problems/index.js
```diff
@@ -1,229 +1,368 @@
...
+        if (item[key] !== null) {
+          return servicesSelected.indexOf(item['service'].toString()) > -1;
+        };
+    });
+  }
 
-    const handleProblemFilterChange = (event) => {
-        setHasFilter(false);
+  // Filter by select fields and text input field.
+  const filterResults = (item) => {
```
bomoko: The "filtered" logic here is a little confusing.

On line 97 you assign "filtered" true - and then you _check_ filtered on 99, 103, and 107 - but then you _also_ set filtered there. AND used filtered as a return value.

I think that the confusion mostly comes from the naming. The name of this function is "filterResults" which initially suggested to me that this should take some results, and filter them.
But after reading through the code, I realized that this isn't actually a filter, but it's a predicate function that tells us whether or not an _item should or should not be filtered_. This slightly confusing naming is repeated in the "sourceFilter", "severityFilter" and "serviceFilter" names.

These should be changed to something more accurate, I think - something that suggests that it's going to return a boolean value. (an isX() or shouldX() function name might make sense).

services/ui/src/components/Problems/index.js
```diff
@@ -1,229 +1,368 @@
...
+      filtered = sourceFilter(item);
+    }
+
+    if (servicesSelected.length > 0 && filtered) {
+      filtered = serviceFilter(item);
+    }
+
+    // Filter text input
+    const lowercasedFilter = problemTerm.toLowerCase();
+    if (filtered && (problemTerm != null || problemTerm !== '')) {
```
bomoko: if you moved this into its own function (like the other three) then you could, I think, reduce this entire function to something like ...

shouldItemBeFiltered = (i) => isFilteredBySeverity(i) || isFilteredByService(i) || isFilteredbySource(i) || isFilteredByTextFilter(i)

I'm not suggesting that you do this kind of thing .. but I think we could get this a lot simpler

bomoko (changes requested):

services/ui/src/components/Filters/helpers.js
```diff
@@ -19,6 +19,12 @@ export const getSourceOptions = gql`
   }
 `;
 
+export const getServiceOptions = gql`
+  query getProblemServices($environment: Int) {
+    services: problemServices(environment: $environment)
```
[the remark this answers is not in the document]
timclifford: Yep it can

services/ui/src/components/Problems/index.js
timclifford: Tidied up the naming of these now

services/ui/src/components/Problems/index.js
timclifford: moved into its own function and clean the whole function up

services/ui/src/components/Problems/index.js
```diff
@@ -1,229 +1,368 @@
...
-        return requestSort(key);
+  const isFilteredByTextFilter = (item) => {
+    return Object.keys(item).some(key => {
+      if (item[key] !== null) {
+        return item[key].toString().toLowerCase().includes(problemTerm.toLowerCase());
+      }
+    });
+  }
+
+  const shouldItemBeShown = (item) => {
```
bomoko, line 112: Hi @timclifford 

As we discussed earlier - if you move the various checks you're doing here in the "if" statements' conditionals, we can simplify the logic of this function to something like (with the functions renamed)

const shouldItemBeShown = (item) => matchesSeveritySelector(item) && matchesServiceSelector(item) && matchesSourceSelector(item) && matchesTextFilter(item)

bomoko (changes requested):

services/ui/src/components/Problems/index.js
timclifford: Thanks again for looking at this - as discussed, renamed functions to something a little clearer and improved the conditional logic too!

bomoko (approved):

bomoko (approved):
Thanks for the changes, @timclifford - lgtm

tobybellwood:
@bomoko and @timclifford - if this needs to make it to Lagoon in the next few weeks, this PR will need to be rebased on the `master` branch instead of `main` before we release 1.14 (as 1.14 is a Lagoon 1-only release)

timclifford:
> @bomoko and @timclifford - if this needs to make it to Lagoon in the next few weeks, this PR will need to be rebased on the `master` branch instead of `main` before we release 1.14 (as 1.14 is a Lagoon 1-only release)

Hi @tobybellwood As discussed, I re-branched off of master with the same changes as above into a new PR #2526

PR #2502, opened 2021-02-11
Only create DBaaS consumers if they don't already exist

### LINKED ISSUE #2499: Migrating from ASB to DBaaS issues
When performing a migration from the ansible service broker in openshift, to the mariadb consumer resource for the dbaas-operator, there are some issues encountered that could cause a lot of issues.

Manually creating the mariadbconsumer resource works fine. But at the first deployment lagoon in openshift does an `oc process`, then `oc apply` of a template, and this overwrites what was manually created, and the dbaas attempts to provision a new database (and is successful) so the original database remains stale, and the new one is provisioned. So 2 databases/users now exist for that environment

Subsequent deployments result in an unchanged `mariadbconsumer` though, so I'm not sure what is happening here.

Here is where we do the oc process https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L591-L604 then we oc apply it here https://github.com/amazeeio/lagoon/blob/master/images/oc-build-deploy-dind/build-deploy-docker-compose.sh#L895 with a bunch of other resources.

I'm not sure why this remains `unchanged` after the initial creation, but i think we should only `oc process` this resource if it doesn't exist.

shreddedbacon:
# Checklist

When migrating an ansible service broker provisioned database to a dbaas-operator consumer resource, when lagoon does a re-deploy of the environment, it re-creates a blank consumer overwriting what was created as part of the migration.

This PR aims to solve the issue by checking if the consumer has already been provisioned (or manually defined) with a consumer database name existing in the resource spec. If there is no existing database name, or the resource doesn't exist

# Closing issues

closes #2499


Schnitzel:
I think this is not needed, and rather can be fixed with creating the resources during the migration via `kubectl apply`, see #2499

Schnitzel:
discussed with @shreddedbacon and we agreed that it's easier right now to add the check if the dbaas object already exists, instead of debugging this, but only on openshift.

shreddedbacon:
@Schnitzel I've updated the PR to remove the kubernetes checks as discussed.

PR #2504, opened 2021-02-11
Limit the size of the CN field in certificate requests

### LINKED ISSUE #2310: Ingress can contain invalid TLS specification
# Describe the bug

If one of the auto-generated domains is longer than 63 characters you can get an error from `cert-manager` and the certificate is not generated.

## To Reproduce

1. Create a project / environment combination with names long enough to hit the 63 character limit.
2. Deploy into Kubernetes.
3. ACME certificates aren't generated, and this event is seen:

```
21m         Warning   Failed              certificate/nginx-tls                         The certificate request has failed to complete and will be retried: Failed to wait for order resource "nginx-tls-j9v2s-618972317" to become ready: order is in "errored" state: Failed to finalize Order: 400 urn:ietf:params:acme:error:badCSR: Error finalizing order :: CSR doesn't contain a SAN short enough to fit in CN
```

## Expected behavior

I expected to get a valid certificate on my auto-generated route.

## Screenshots

n/a

## Additional context

I've tested this workaround and it does work https://github.com/jetstack/cert-manager/issues/1462#issuecomment-472416825

So probably the solution is to truncate the auto-generated domain if it is too long and prefix it to the list of TLS domains in the ingress.

smlx:
# Checklist

This PR checks the auto-generated domain length and, if it is larger than the maximum 63 characters, inserts a shortened version into the ingress's `tls.hosts` list. This first domain is used in the `CN` field of the certificate request, and when it is larger than 63 characters the certificate request is invalid and `cert-manager` cannot generate certificates for the ingress.

IMPORTANT NOTE: this PR _also_ makes a change to the lagoonbuild-controller deploy spec. Previously we were replacing the `${environment}` and `${project}` variables in the `routerPattern` before sending it to the controller. However this replacement logic is also present in the controller itself:

https://github.com/amazeeio/lagoon-kbd/blob/149a57af9b7e5258194214e89d7f27f90d59a3e6/controllers/lagoonbuild_controller.go#L311-L327

Since this PR uses the raw `routerPattern`, it removes the logic which does the replacement in the Lagoon API. However this may break any other build-deploy systems - I've only tested against the lagoonbuild-deploy controller.

~This PR also depends on https://github.com/amazeeio/lagoon-kbd/pull/37~ merged.

# Closing issues

Closes: #2310


seanhamlin (approved):
I like this a lot. If I understand this correctly, the hashed `CN` is only used when the normal pattern is > 63 characters. This all seems OK to me.

tobybellwood:
@smlx - this is targeted at `main` - do we want this in the 1.14 release - if so, it'll need to be modified for `master`?

Schnitzel:
moved to the Lagoon2 milestone as this is against `main`

PR #2515, opened 2021-02-18
Set StorageClass properly for node-persistent pvc

tobybellwood:
This correctly allocates the correct storageclass in the correct accessmode for node-persistent deployment pvcs (as per nginx-php-persistent) for k8s deployments


thom8:
LGTM 👍

thom8 (approved):

PR #2529, opened 2021-02-23
Fix SSH check for scaling CLI deployments in kubernetes

shreddedbacon:
# Checklist

As part of working on idling in Kubernetes, I uncovered that there is an issue with how we check if a CLI pod is scaled in Kubernetes.

Previously we would use this check, but it appears that Kubernetes doesn't set `.status.replicas` at all when a deployment is scaled to 0
```
if [[ $($OC get ${DEPLOYMENT} -o go-template --template='{{.status.replicas}}') == "0" ]]; then
```
The returned value is actually `<no value>` as it doesn't exist.

This PR changes the check to the following, which assumes a 0 value if `.status.replicas` returns null by jq
```
if [[ $($OC get ${DEPLOYMENT} -o json | jq -r '.status.replicas // 0') == "0" ]]; then
```


smlx (approved):
LGTM

PR #2542, opened 2021-02-25
Add container to the list of rollout options

KorvinSzanto:
This PR adds `container` to the list of `pre-rollout` and `post-rollout` options in the .lagoon.yml documentation. The container parameter is used [here](https://github.com/amazeeio/lagoon/blob/main/images/kubectl-build-deploy-dind/scripts/exec-tasks-run.sh#L22-L26) to run the given command on the proper container.

This PR also indents the other second-level bullet items in the hopes that it will render properly on readthedocs because [they don't with the current indentation](https://lagoon.readthedocs.io/en/latest/using_lagoon/lagoon_yml/#post-rollout-tasks-post_rolloutirun) even though readthedocs totally supports second-level bullets.

I figured this may not require discussion given that it's a documentation page, if that's not the case I'll close and follow the process outlined by the PR template.

### DIFF
--- a/docs/using_lagoon/lagoon_yml.md
+++ b/docs/using_lagoon/lagoon_yml.md
@@ -107,13 +107,15 @@ Here you can specify tasks which need to run against your project, _after_:
 Common uses for post-rollout tasks include running `drush updb`, `drush cim`, or clearing  various caches.
 
 * `name`
-  * The name is an arbitrary label for making it easier to identify each task in the logs.
+    * The name is an arbitrary label for making it easier to identify each task in the logs.
 * `command`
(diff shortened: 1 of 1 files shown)


tobybellwood:
Just a quick note that our documentation is being constantly updated at our https://docs.lagoon.sh site (driven from our `main` branch) - we're trying to grapple with moving it off readthedocs - and this readthedocs_legacy branch is frozen in time!

This page lives at https://docs.lagoon.sh/lagoon/using-lagoon-the-basics/lagoon-yml#post-rollout-tasks-post_rollout-i-run now (which is https://github.com/amazeeio/lagoon/blob/main/docs/using-lagoon-the-basics/lagoon-yml.md)

KorvinSzanto:
Ah okay, I'll see if they can reapply this to main here in a bit

PR #2553, opened 2021-03-03
Removes verbose details from error logging in Problems

### LINKED ISSUE #2357: Problems DB error logs very verbose
# Describe the bug

We are seeing some errors from elasticsearch when indexing `lagoon-logs`:
```
error_class=Fluent::Plugin::ElasticsearchErrorHandler::ElasticsearchError error="400 - Rejected by Elasticsearch [error type]: illegal_argument_exception [reason]: 'Limit of total fields [1000] in index [lagoon-logs--_-all_environments-_-2020.12] has been exceeded'"
```

The error message concerned appears to be generated by the problems DB system when it encounters an error inserting duplicate records:
```
"event"=>"problems:drutiny:resultset:problem_insert_error"
```

The log message itself appears to contain a full result set and is >91KB in size.

A contributing factor is that `lagoon-logs` is expected to contain a `project` field containing the name of the project concerned. Because this field is missing, errors for all projects are ending up in a single index.

## To Reproduce

1. Run Lagoon at scale in production :smiley: 
2. Occasionally see these error messages in logs-concentrator.

## Expected behavior

There might be a way to pare down the log message a bit to avoid sending so many logs and hitting the elasticsearch limit.

In addition, maybe the `project` field of the log message should be populated with the name of the project concerned? That way the logs will end up in the `lagoon-logs-<project>-*` index and the customer/tenant will be able to see the logs in Kibana. However I'm not sure if this behaviour is desirable.. should customers be able to view these error logs?

## Screenshots

n/a

## Additional context

n/a

bomoko:
As described in #2357, the logging for the Problems DB (particularly the Drutiny scan) is far too verbose. Essentially we were sending along the entire request into the logs (which can be json objects with hundreds of fields or nested objects).

This PR simply removes the metadata that delivers the entire webhook through to the logs.

# Checklist

- [*] Affected Issues have been mentioned in the Closing issues section

# Closing issues

closes #2357

### DIFF
--- a/services/webhooks2tasks/src/handlers/problems/processDrutinyResults.ts
+++ b/services/webhooks2tasks/src/handlers/problems/processDrutinyResults.ts
@@ -141,7 +141,7 @@ export async function processDrutinyResultset(
                     '',
                     uuid,
                     `${webhooktype}:${event}:problem_insert_error`,
-                    { data: body },
+                    {},
                     `Error inserting problem id ${element.id} for ${lagoonProjectId}:${environmentDetails.id} -- ${error.message}`
                   )
                 );
@@ -153,7 +153,7 @@ export async function processDrutinyResultset(
       '',
       uuid,
       `${webhooktype}:${event}:unhandled`,
-      { data: body },
+      {},
       `Could not process incoming Drutiny scan results, reason: ${error}`
     );
   }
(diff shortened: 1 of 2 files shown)


tobybellwood:
Problems logging is being reworked

PR #2557, opened 2021-03-05
Add option to mount local directories into lagoon-core pods for local development

smlx:
# Checklist

This PR is a PoC for local development using `kind` that tries to approximate the experience using the existing `docker-compose` workflow. It mounts your local `./services/api/src` into the `api` pod running in the local cluster so that code is hot-reloadable.

It currently relies on [a branch](https://github.com/uselagoon/lagoon-charts/tree/insecure-host-mounts) in the `lagoon-charts` repository.

Test this by running `make -j8 -O kind/up INSECURE_HOST_MOUNT_API_SRC=true`. Then when the `api` pod appears you can `kubectl exec` into it and see that changes made to your local `./services/api/src` are reflected inside the pod.

# Closing issues

n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -986,9 +986,9 @@ helm/repos: local-dev/helm
 	./local-dev/helm repo add bitnami https://charts.bitnami.com/bitnami
 	./local-dev/helm repo update
 
+# stand up a kind cluster configured appropriately for lagoon testing
 .PHONY: kind/cluster
 kind/cluster: local-dev/kind
-	# stand up a kind cluster configured appropriately for lagoon testing
 	./local-dev/kind get clusters | grep -q "$(CI_BUILD_TAG)" && exit; \
 		docker network create kind || true \
 		&& export KUBECONFIG=$$(mktemp) \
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -1005,6 +1005,11 @@ kind/cluster: local-dev/kind
 		&& echo '  extraMounts:'                                                                      >> $$KINDCONFIG \
 		&& echo '  - containerPath: /var/lib/kubelet/config.json'                                     >> $$KINDCONFIG \
 		&& echo '    hostPath: $(HOME)/.docker/config.json'                                           >> $$KINDCONFIG \
+		&& if [ '$(INSECURE_HOST_MOUNT_API_SRC)' = true ]; then \
+		echo '  - containerPath: /lagoon/services/api/src'                                            >> $$KINDCONFIG \
+		&& echo '    hostPath: ./services/api/src'                                                    >> $$KINDCONFIG \
+		&& echo '    readOnly: true'                                                                  >> $$KINDCONFIG \
```
rocketeerbkw: does this make the files read only or just specifying a one-way sync? part of the dev process inside the api is compiling typescript which needs write access to `/services/api/src/dist` and `/node-packages`

smlx: oh, okay. yes, this currently makes everything inside `services/api/src` read-only from inside the pod. does the process doing the recompilation run inside the pod?

smlx: I thought this might be useful to make it read-only from inside the pod but of course can leave it as read-write if that makes more sense.

PR #2559, opened 2021-03-05
fix S3_FILES_HOST variable name 

Schnitzel:
was called `S3_HOST` before, but the lagoon-charts use S3_FILES_HOST: 
https://github.com/uselagoon/lagoon-charts/blob/3d2ad1019a9dc9b7f1afda392ba32583d0405cde/charts/lagoon-core/templates/api.deployment.yaml#L112


Schnitzel:
superseeded by https://github.com/uselagoon/lagoon/pull/2742

Issue #2322, opened 2020-11-12
Lagoon Environments with slashes in their name aren't handled correctly when deploying to Kubernetes


smlx:
# Describe the bug

https://github.com/amazeeio/lagoon/blob/master/services/kubernetesbuilddeploy/src/index.ts#L41-L64

In this block of code we first `ocsafety` the environment name (at the top) and _then_ check to see if it matches the project's `productionEnvironment` (at the bottom). If any characters in the environment name are replaced in `ocsafety` then the environment will not be recognised as a production environment, and in fact it will be reset to `development` here: https://github.com/amazeeio/lagoon/blob/master/services/kubernetesbuilddeploy/src/index.ts#L389

## To Reproduce

1. Create an environment with a `/` in the name and `environmentType: PRODUCTION` in a project on a kubernetes deployment target.
2. Set the `productionEnvironment` of the project to this new environment.
3. Deploy.
4. See that the deployed `lagoon-env` has `LAGOON_ENVIRONMENT_TYPE: development`, and the `environmentType` has been set to `development` in the Lagoon API.

## Expected behavior

openshiftbuilddeploy has the correct behaviour because it compares `productionEnvironment` to the unsanitised environment name: https://github.com/amazeeio/lagoon/blob/master/services/openshiftbuilddeploy/src/index.ts#L48-L68

## Screenshots

n/a

## Additional context

Current workaround is to set `productionEnvironment` to the sanitised version of the environment name. e.g. replace all the slashes with hyphens.

PR #2565, opened 2021-03-11
Add Environment and Project IDs to controller builds

### LINKED ISSUE #2568: API updates not happening when environments with special characters complete.
# Describe the bug

There is currently an issue where if an environment that has special characters in it, eg `/`, are handled by the controller. 

Currently Lagoon will make safe (dns compliant) environment names before sending to the controller. This is fine, but any time an update or entry is returned to Lagoon to be handled by the controllerhandler, the API check for the environment fails as lagoon is expecting to see the environment name before it was made safe.

## To Reproduce

1. Deploy an environment like `feature/my-cool-feature`
2. On completion check the API/UI for the existence of any `ROUTES`
3. See there are no routes

## Expected behavior

Routes and/or other updates should be in the API

## Additional context

Checking the controllerhandler logs shows that the updates failed because of the environment not being found, further investigation shows that this is because of the environment name being made dns-compliant before being sent to the controller.

shreddedbacon:
# Checklist

Somewhat related to #2322, Lagoon sends a DNS compliant version of the environment name to the build deploy controller. When a build is completed, it responds to Lagoon with this modified environment name. In the event that the original environment name contained slashes or other special characters, Lagoon would not actually update the API with the information required.

Lagoon will prefer to use the environment ID if it is present, but will fall back to using the environment name if no ID is found.

This is a partner to https://github.com/amazeeio/lagoon-kbd/pull/41 , which adds the support for these IDs into the build deploy controller.

# Closes

closed #2568

### DIFF
--- a/node-packages/commons/src/api.ts
+++ b/node-packages/commons/src/api.ts
@@ -866,6 +866,36 @@ export async function getEnvironmentByName(
   return result;
 }
 
+
+export async function getEnvironmentById(
+  id: number
+): Promise<any> {
+  const result = await graphqlapi.query(`
+    {
+      environmentById(id: ${id}") {
+        id,
+        name,
+        route,
+        routes,
+        deployType,
+        environmentType,
+        openshiftProjectName,
+        updated,
+        created,
+        deleted,
+      }
+    }
+  `);
+
+  if (!result || !result.environmentById) {
+    throw new EnvironmentNotFound(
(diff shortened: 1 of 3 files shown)


shreddedbacon:
Just FYI, there is no issue releasing this before the controller is updated as it will fall back to the original behaviour.

Once the controller gets updated, then it will use the ID values properly.

PR #2569, opened 2021-03-12
Keycloak crashing on mass api queries

timclifford:
# Checklist

# Closing issues

### DIFF
--- a/services/webhooks2tasks/src/handlers/problems/harborScanningCompleted.ts
+++ b/services/webhooks2tasks/src/handlers/problems/harborScanningCompleted.ts
@@ -59,6 +59,11 @@ const DEFAULT_REPO_DETAILS_MATCHER = {
     let vulnerabilities = [];
     vulnerabilities = await getVulnerabilitiesFromHarbor(harborScanId);
 
+                                          // Test data
+                                          // vulnerabilities.push(...vulnerabilities);
+                                          // vulnerabilities.push(...vulnerabilities);
+                                          // vulnerabilities.push(...vulnerabilities);
+
     let { id: lagoonProjectId } = await getProjectByName(lagoonProjectName);
 
     const result = await getOpenShiftInfoForProject(lagoonProjectName);
@@ -152,6 +157,12 @@ const validateAndTransformIncomingWebhookdata = async (rawData) => {
     lagoonServiceName,
    } = matchRepositoryAgainstPatterns(repository.repo_full_name, harborScanPatternMatchers.allProblemHarborScanMatchers);
 
+
+
+  lagoonProjectName = "credentialstest-project1-openshift";
+
+
+
   return {
     resources,
     repository,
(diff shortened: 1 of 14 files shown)


tobybellwood:
for reference, what are the default values for max_allowed_packet, max_connections and wait_timeout currently?

services/webhooks2tasks/src/handlers/problems/harborScanningCompleted.ts
```diff
@@ -76,29 +76,38 @@ const DEFAULT_REPO_DETAILS_MATCHER = {
...
-      messageBody
-    );
-
-    const buffer = new Buffer(JSON.stringify(webhookData));
-
-    await channelWrapperWebhooks.publish(`lagoon-webhooks`, '', buffer, {
-      persistent: true,
-    });
-
+    if (batchedArrVulns) {
```
bomoko, line 86: Hey @timclifford - I don't know about this logic here, I think it introduces an error.
The problem is that when we get a new set of vulnerabilities - we actually clear out the current set of vulnerabilities for that environment (https://github.com/amazeeio/lagoon/blob/9e64343280ac99e0af5f5c4d7b144a2aff149781/services/webhooks2tasks/src/handlers/problems/processHarborVulnerabilityList.ts#L37)

If I'm reading this correctly, we're splitting a set of vulns into multiple sets - but doesn't this mean that everytime we process a new chunked set of vulnerabilities, the last set is deleted _even if that set is part of the current vunerability scan_? (i.e. if there are three chunks, only the problems from the 3rd chunk will show up?)

timclifford:
> for reference, what are the default values for max_allowed_packet, max_connections and wait_timeout currently?

api-db and keycloak db
| max_connections | 400   |
| max_allowed_packet | 67108864 |
| wait_timeout  | 28800 |

services/webhooks2tasks/src/handlers/problems/harborScanningCompleted.ts
timclifford: @bomoko ah yes - I thought I had this chunking post deletion of problems from source, but you are right this would be an issue. I was thinking though that rather than change this slightly, the best solution would be to be replace this with the change to the api we discussed. Which would be to create a bulkProblems mutation in the api that we would use here (instead of iterating through each problem and adding that single problem via the addProblem query). Then we could also control this issue within the resolver of the api itself (which seems like the better place for it). Ideally, the issue we have lies in the fact that keycloak is timing out/has too many open connections due to being overloaded by the api so makes sense to actually try tackle that there.

timclifford: so i suggest perhaps we hold this - and revise once we have considered how the api should handle large volumes of problems like this.

tobybellwood:
We're no longer bulk adding problems to the API, this is no longer an issue (with this exact thing :stuck_out_tongue_winking_eye: )

PR #2572, opened 2021-03-15
Add support in kubectl-build-deploy-dind for running rootless

smlx:
# Checklist

This PR is part of https://github.com/amazeeio/lagoon/pull/2523 that I have split out separately since I realised that it was possible to add support to kubectl-build-deploy-dind for running as either `root` or rootless.

# Closing issues

Closes #2523


tobybellwood:
Happy to merge this, I've tested locally with and without rootless builds, and it's all fine!

I think we can close the other PRs too, unless we want to run our tests as rootless builds?

smlx:
I think we should just make rootless builds the default if they work because when Pod Security Standards are eventually integrated into Lagoon it would be neat to apply a "no pods running as root" policy on all Lagoon workload namespaces.

PR #2574, opened 2021-03-16
Merge service and the first component of routerPattern into a single label on autogenerated routes

### LINKED ISSUE #2573: Prefixing services as subdomains breaks Oauth2 authentication to environments
# Describe the bug

Situation: A web developer uses Oauth2 for authentication to their PR environments.

Oauth2 generally is quite strict about limiting redirect URIs to be absolute. The reasons behind this are described in the [Oauth2 threat model RFC](https://tools.ietf.org/html/rfc6819#section-5.2.3.5).

However some Oauth2 providers such as auth0 do allow limited [wildcards for subdomains](https://auth0.com/docs/applications/wildcards-for-subdomains) as a convenience. These rules are similar to those for wildcard DNS records. In particular, only the leftmost label in the domain may be a wildcard.

On Openshift this works because for autogenerated routes we prefixed the service, environment, and project to the cluster domain as a single label joined by hyphens. e.g. `app-myproject-pr-1234.lagoon.mycluster.example.com`.

On k8s however, we separate these components into separate labels e.g. `app.pr-1234.myproject.lagoon.mycluster.example.com`

This means that there is no way to capture auto-generated domains using a wildcard given the rules linked above.

## To Reproduce

1. Deploy a PR environment using Oauth2 via auth0.
2. Authentication doesn't work.
3. There's no way to get it working with the current autogenerated router pattern system.

## Expected behavior

I would expect existing Lagoon workflows to continue working on kubernetes the same way they work on openshift.

## Screenshots

n/a

## Additional context

n/a

smlx:
# Checklist

Instead of ingresses with domains like `${service}.${routerPattern}`, we can do `${service}-${routerPattern}`. This enables some use-cases in kubernetes that used to work on openshift by allowing a single domain wildcard match all auto-generated domains of a project.

This has the potential to break things since the autogenerated route is changing. So it's a draft pending review.

# Closing issues

Closes: #2573


shreddedbacon:
LGTM

Only issue I can see is that this will create all new autogenerated ingress on any environment deployed after the change, leaving the existing ones in place.

Would there be a requirement to remove the old ones at some stage?

smlx:
Good question. I don't know.

In fact this behaviour of not removing existing ingresses might be good for backwards compatibility since a deploy won't break auto-generated domains of existing environments.

tobybellwood:
Would the prefix domains also need to be accounted for here?  i presume the chance of OAuth'ing into an en|de version of a domain is low, but probably needs to be thought through?
`{{- printf "%s.%s-%s" .prefix .root.Release.Name .root.Values.routesAutogenerateSuffix | trimSuffix "-" -}}`

smlx:
> Would the prefix domains also need to be accounted for here?

I imagine that if you are explicitly requesting prefix domains then you don't want to apply OAuth2 wildcards, but maybe @twardnw is more familiar with the prefix domain use-case?

smlx:
> Only issue I can see is that this will create all new autogenerated ingress on any environment deployed after the change, leaving the existing ones in place.

Is this correct? The hostname is the only thing changing in this PR - the name of the ingress object will be the same I believe?

In that case it will overwrite existing autogenerated routes which may cause other problems...

shreddedbacon:
> > Only issue I can see is that this will create all new autogenerated ingress on any environment deployed after the change, leaving the existing ones in place.
> 
> Is this correct? The hostname is the only thing changing in this PR - the name of the ingress object will be the same I believe?
> 
> In that case it will overwrite existing autogenerated routes which may cause other problems...

You are right, that may cause some issues too then?

shreddedbacon:
Would it be worth having this configurable then? For those that are impacted by it they could change the behaviour in the .lagoon.yml as they see fit?

smlx:
yep, probably worth being configurable I guess. will have a look into the right way to configure this.

tobybellwood:
This has been implemented in https://github.com/uselagoon/lagoon/pull/2953. Closing :smile_cat:

PR #2583, opened 2021-03-24
Scale idled environments up if SSH connections are made

shreddedbacon:
# Checklist

With the introduction of idling to Kubernetes clusters, sometimes if an environment is idled, the user may experience an issue trying to run commands in the requested service if other services are idled.

This adds a step to the SSH connection request that will scale any idled services up. It does this before the requested service is loaded and aims to help reduce this issue.


services/ssh/home/rsh.sh
```diff
@@ -120,6 +124,36 @@ fi
...
+        # for unidling an entire environment and waiting for the number of `readyReplicas`
+        # to be 1 for each deployment, could add considerable delays for the ssh connection to establish.
+        # WAIT_TO_UNIDLE_SERVICES will default to false so that it just scales the deployments
+        # and won't wait for them to be ready, but if set to true, it will wait for `readyReplicas` to be 1
+        if [[ "$WAIT_TO_UNIDLE_SERVICES" =~ [Tt][Rr][Uu][Ee] ]]; then
+          while [[ ! $($OC get ${DEP} -o go-template --template='{{.status.readyReplicas}}') == "1" ]]
+          do
+            sleep 1
+          done
+        fi
```
Schnitzel: I would suggest two changes:
1. can we split the actual unidling and the waiting into two for loops? In this case we first trigger scaling of all deployments and then maybe wait for them to be available, which would probably reduce the time that we need to wait.
2. if a deployment had 2 replicas, the `--template='{{.status.readyReplicas}}') == "1"` will fail, I think we should just check that is has 1 or more replicas.

Schnitzel (changes requested):

services/ssh/home/rsh.sh
shreddedbacon: On 1) waiting for all services in the environment to become available could add considerable delays to establishing the ssh connection though. Which is why I did it the way I did, where it unidles them but won't wait for them to become ready (unless the `WAIT_TO_UNIDLE_SERVICES` var is set to true, which defaults to false)

I'll fix the 1+ replica check now.

PR #2592, opened 2021-04-01
Updating backupcommands for solr, postgres-single, and mongodb-single

cdchris12:
This PR specifically excludes the `lost+found` folder from being included in any backup attempts, as this folder is not readable by the user these containers are running as.

# Checklist

- [] Documentation has been written/updated


Schnitzel:
@cdchris12 
can you cherry-pick these fixes into a PR against `master`? we would like to merge this into 1.14.4

smlx:
GNU `tar` is picky about the order of `--exclude=*` and the `tar` source.

We need to change this to `--exclude="lost+found" .` instead of `. --exclude="lost+found"`.

Otherwise we get errors like:
```
tar: The following options were used after any non-optional arguments in archive create or update mode.  These options are positional and affect only arguments that follow them.  Please, rearrange them properly.
tar: --exclude ‘lost+found’ has no effect
tar: Exiting with failure status due to previous errors
```

cdchris12:
> GNU `tar` is picky about the order of `--exclude=*` and the `tar` source.
> 
> We need to change this to `--exclude="lost+found" .` instead of `. --exclude="lost+found"`.

Agreed, but we aren't using GNU tar here, we're using busybox's tar, which does accept this format.

smlx:
Not in `solr`, we have switched to upstream base image which is Debian (GNU coreutils).

tobybellwood:
Noting that only some solr is debian based now, there will be both it and solr for some time

Schnitzel:
@cdchris12 
can you explain what the `|| [ $? -eq 1 ]` exactly does? I understand that it checks if the return code of `tar` is not `0` then it checks if its `1` and if it is `1` it sees it as success.
But what does tar return code `1` exactly mean? I'm worried that this could swallow errors and we have failing backups without knowing that they are failing.

Schnitzel:
also please fix the conflicts :)

cdchris12:
> @cdchris12
> can you explain what the `|| [ $? -eq 1 ]` exactly does? I understand that it checks if the return code of `tar` is not `0` then it checks if its `1` and if it is `1` it sees it as success.
> But what does tar return code `1` exactly mean? I'm worried that this could swallow errors and we have failing backups without knowing that they are failing.

Your analysis of this code is correct; it checks for the `tar` command to return `1`; if that happens, it returns `0` so that `wrestic` will see this as a success instead of a failure. This could have unintended consequences by trapping other errors, but `wrestic` will see a return code of `1` as a failure (rightfully so) and mark the backup as failed.  I'm not sure this is the _right_ answer, but it's the answer that gets our backups working in the short term. :)

Schnitzel:
@cdchris12 
okay, in which circumstances does tar return with rc `1` then? I assume you are trying to only catch the rc `1`, because an `|| true` could have caught all possible errors?

cdchris12:
> @cdchris12
> okay, in which circumstances does tar return with rc `1` then? I assume you are trying to only catch the rc `1`, because an `|| true` could have caught all possible errors?

That's right, `|| true` would catch every error. I didn't actually know this until just now reading the docs, but `tar` only returns `1` in the specific instance where files are changed during the backup operation. With that in mind, my code is actually the perfect workaround for that issue. :)

https://man7.org/linux/man-pages/man1/tar.1.html

smlx:
Unfortunately the "exit code 1 for missing files" thing is GNU tar specific :disappointed: 

For example it works on a Debian based image:

```
$ docker run --rm -it debian sh -c 'tar -xf /tmp/fake; echo $?'
tar: /tmp/fake: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
2
```

But not on Alpine/Busybox, which exits with 1 for other fatal errors that we definitely need to know about:

```
$ docker run --rm -it alpine sh -c 'tar -xf /tmp/fake; echo $?'
tar: can't open '/tmp/fake': No such file or directory
1
```

Schnitzel:
discussed with @cdchris12 in a call, we think the best solution right now is to go with the `|| [ $? -eq 1 ]` solution, plus getting installing gnu tar in the `mongodb`, `elasticsearch` and `solr` base images (https://github.com/uselagoon/lagoon-images/issues/169)
Until the base images are updated the backup will swallow any errors from tar, but we know that they are very uncommon and mostly happen because of the changed files (which is what `|| [ $? -eq 1 ]` will ignore).
We also looked at the `--ignore-failed-read` paramater for tar, but this one would cause busybox tar to completely fail and therefore is not a good solution for now.

Schnitzel (approved):

PR #2611, opened 2021-04-15
adding service to deleteProblem mutation

timclifford:
Currently the deleteProblem mutation will delete _all_ problems from an environment given a particular identifier.

The problem with this is that the same identifier may affect multiple services. 

This PR will allow more precise targeting when deleting problems.

# Checklist


tobybellwood:
@timclifford is this still something we need?

bomoko:
@tobybellwood - rerolled and tested successfully.

I basically just needed to incorporate a couple of items that were changed in the interim. But this is definitely needed and useful.

bomoko:
@rocketeerbkw I've tagged you as reviewer just 'cause I did the reroll, so it didn't seem right for me to approve too :)

rocketeerbkw (dismissed):
> But this is definitely needed and useful.

That's not obvious, perhaps this was discussed elsewhere.

The PR description is still the default template, with no changes. Can we update that so that I know why it's "needed and useful" and what I should be reviewing the functionality of? With all three "to-do" items unchecked, it's also not clear that's is ready for review (besides the ping).

I know this is an old issue, but that's all the more reason to ask for more info :smile:

services/api/src/resources/problem/sql.ts
```diff
@@ -128,14 +128,19 @@ export const Sql = {
...
-      .del()
-      .toString(),
+      });
+
+    if (service != undefined) {
+      q.where('lagoon_service', service);
+    }
+
+    return q.del().toString()
+  },
```
tobybellwood: Minor formatting nit - :see_no_evil: at the rest of the file(s)

```suggestion
      });
    if (service != undefined) {
      q.where('lagoon_service', service);
    }
    return q.del().toString();
  },
```

tobybellwood (approved):
This is a sensible addition. No major changes to queries or access, just allows targeted deletion of problems. Tested with/without service specified ok.

Of note here, is that deleting a problem with a service that has no matches (or non-existent) will return "success" - as does deleting a problem with an identified that has no matches (or non-existent). Until we have more fine-grained errors returned via graphql, this is acceptable, and is consistent with current behaviour.

PR #2646, opened 2021-05-11
Enables setting ids directly in facts and problems

### LINKED ISSUE #2630: `addFact`/`addFacts`/`addProblem` submitted with `id` parameter should warn user
# Describe the bug

The `addFact`, `addFacts` and `addProblem` mutations have `id` as part of the [`AddFactInput`](https://github.com/amazeeio/lagoon/blob/main/services/api/src/typeDefs.js#L208) but the [resolvers](https://github.com/amazeeio/lagoon/blob/main/services/api/src/resources/fact/resolvers.ts#L62) doesn't use it so the id is always autogenerated by the database.

## To Reproduce

1. Add a test fact (in a fresh database)
    ```
    mutation {
      addFact(input: {
        id: 100
        description: "desc"
        environment: 1
        name: "test-fact"
        value: "val"
        source: "src"
      }) {
        id
        name
        description
        source
        value
      }
    }
    ```
2. See that the id is not what was entered
    ```
    {
      "data": {
        "addFact": {
          "id": 1,
          "name": "test-fact",
          "description": "desc",
          "source": "src",
          "value": "val"
        }
      }
    }
    ```

## Expected behavior

The mutation saves all data correctly by the user, including the `id`.

bomoko:
Currently Facts and Problems don't set the id manually and ignore the incoming id. This PR fixes that.

# Checklist

# Closing issues

closes #2630


bomoko:
I have second thoughts about this - being able to set the ID, even though it's specified by the API, feels like we're just asking for trouble (what happens, for instance, if someone sets their `id` to the max integer value?)

How best to approach this @tobybellwood and @rocketeerbkw?

tobybellwood:
CLosing this PR, as allowing users to specify their ids is undesirable. Handling any b/c issue separately

PR #2653, opened 2021-05-16
Split each test into it's own setup/teardown, reduce api-data-watcher-pusher, update tests

tobybellwood:
Currently, we rely on the api-data-watcher-pusher to set up all the projects, environments, variables, users and keys that all the tests need to operate.  This is overly cumbersome (if you just need a single test stream to pass) - and can lead to a divergence between writing a test and configuring the API payload.

This PR:
* creates templatable graphQL queries for common API functions
* For each top-level test, the project set-up, environment creation, and variable setting is done as part of the test, as is the project removal
* It splits the features-kubernetes tests up slightly, reducing overall time
* Moves the "promote" test into the API section
* Creates a git project&repo per Drupal version, to avoid the project being removed across tests
* Makes the drush tests multi-drupal versions compatible
* Updates the entire test stack to Python 3.8 and Ansible 2.11

*TO-DO* in this space (not necessarily this PR)
* more split-out of features-kubernetes to take it to 10-15 minutes
* introduce Drupal9 tests
* minimise duplication of Drupal tests, and test DBaaS more
* code tidyup
* create next iteration of `high-cotton` that is a real project, and persistent


tobybellwood:
tagging @bomoko as it'll interfere with some of the tests you've written!

PR #2659, opened 2021-05-19
Adds async calls to parallelize Bitbucket groups syncs

bomoko:
This PR is to adjust the Bitbucket sync script to run certain operations in parallel, as well as some further optimizations of the reads and writes from the various apis.

# Checklist

### DIFF
--- a/services/api/src/bitbucket-sync/repo-permissions.ts
+++ b/services/api/src/bitbucket-sync/repo-permissions.ts
@@ -32,6 +32,7 @@ const isNotEmpty = R.complement(R.isEmpty);
 // Returns true if user was added or already exists.
 // Returns false if adding user failed and no user exists.
 const addUser = async (email: string): Promise<boolean> => {
+  // try {
   try {
     await api.addUser(email);
   } catch (err) {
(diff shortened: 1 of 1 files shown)


services/api/src/bitbucket-sync/repo-permissions.ts
```diff
@@ -61,108 +61,160 @@ const getBitbucketRepo = async (gitUrl: string, projectName: string) => {
...
+              ) {
+                const userAddedOrExists = await addUser(email);
+                if (!userAddedOrExists) {
+                  // Errors for this case are logged in addUser
+                  continue;
+                }
+
+                existingUsers.push(email);
+              }
+              if (!userExistsInGroupAlready) {
```
bomoko: Hmm, I just realized this might need to go/change - it might be too aggressively ignoring users. I need to check if their permissions have changed.

bomoko (commented):
Check permissions before ignoring write

services/api/src/bitbucket-sync/repo-permissions.ts
bomoko: Removed this in case, still way faster.

PR #2664, opened 2021-05-19
Changes permission check logic for AddFacts

### LINKED ISSUE #2663: AddFacts putting strain on Keycloak
Currently the AddFacts call is overcalling keycloak for environment permissions checks - it calls once per FACT. This should be changed to once per ENVIRONMENT.

bomoko:
Currently, the `AddFacts` resolver is doing a keycloak call for every fact pushed through to it. This is potentially overwhelming the API and keycloak.

This PR simplifies this to only check keycloak once per environment per call. It also serializes writes to the DB for facts.

# Checklist

# Closing issues

closes #2663

### DIFF
--- a/services/api/src/resources/fact/resolvers.ts
+++ b/services/api/src/resources/fact/resolvers.ts
@@ -82,19 +82,24 @@ export const addFacts = async (
   { sqlClient, hasPermission }
 ) => {
 
-  // We first check that the user has access to all of the environments, so this is an atomic operation.
-  await facts.map(async (fact) => {
-    const { environment } = fact;
-    const env = await environmentHelpers(sqlClient).getEnvironmentById(environment);
+  const environments = facts.reduce((environmentList, fact) => {
+    let { environment } = fact;
+    if(!environmentList.includes(environment)) {
+      environmentList.push(environment)
+    }
+    return environmentList
+  }, []);
 
+  for(let i = 0; i < environments.length; i++) {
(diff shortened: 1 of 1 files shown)


services/api/src/resources/fact/resolvers.ts
```diff
@@ -82,19 +82,24 @@ export const addFacts = async (
...
       project: env.project,
     });
-  });
-
-  return await facts.map(async (fact) => {
-    const { environment, name, value, source, description } = fact;
+  }
 
+  const returnFacts = []
+  for(let i = 0; i < facts.length; i++) {
```
rocketeerbkw: These don't conform to our code style standards. We don't use ASI and use spaces after control words.

I already know that on `master` we're pretty bad on this issue, so this is just a "for next time" comment. With VS Code and the "Prettier - Code formatter" plugin you can run "Format Document" on any file to correct style issues.

rocketeerbkw (approved):

services/api/src/resources/fact/resolvers.ts
bomoko: Hey @rocketeerbkw - no worries, let me fix this :)
I won't reformat the whole file since that's offsides when it comes to making actual changes - but I'll fix these up.

As for the ASI ... ahem, that was less a conscious choice than it is a consequence of writing golang recently. Sorry :)

PR #2665, opened 2021-05-20
Allow Drush Cache Clear task for Drupal 9

tobybellwood:
The Drush cache clear task is hardcoded to check for Drupal 7 or 8 - this PR just adds D9 to trigger the `drush cr` task.

### DIFF
--- a/services/api/src/resources/task/resolvers.ts
+++ b/services/api/src/resources/task/resolvers.ts
@@ -345,10 +345,10 @@ export const taskDrushCacheClear: ResolverFn = async (
 
   const command =
     'drupal_version=$(drush status drupal-version --format=list) && \
-  if [ ${drupal_version%.*.*} == "8" ] || [ ${drupal_version%.*.*} == "9" ]; then \
-    drush cr; \
-  elif [ ${drupal_version%.*} == "7" ]; then \
+  if [ ${drupal_version%.*} == "7" ]; then \
     drush cc all; \
+  elif [ ${drupal_version%.*.*} -ge "8" ] ; then \
+    drush cr; \
   else \
     echo "could not clear cache for found Drupal Version ${drupal_version}"; \
     exit 1; \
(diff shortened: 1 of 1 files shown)


klonos (approved):
LGTM 👍🏼

services/api/src/resources/task/resolvers.ts
```diff
@@ -345,7 +345,7 @@ export const taskDrushCacheClear: ResolverFn = async (
 
   const command =
     'drupal_version=$(drush status drupal-version --format=list) && \
-  if [ ${drupal_version%.*.*} == "8" ]; then \
+  if [ ${drupal_version%.*.*} == "8" ] || [ ${drupal_version%.*.*} == "9" ]; then \
```
seanhamlin: Is it possible in bash to do a `>= 8`?

```suggestion
  if [ "${drupal_version%.*.*}" -ge "8" ]; then \
```

tobybellwood: yes and no - because the first parser matches D7 too - it pulls 7.59 (b/c no semver) and then can't do an integer comparison :cry: and throws an error (although it does then clear)
```
www-data@64846eb8a028:/app$ drupal_version=$(drush status drupal-version --format=list) && \
>   if [ ${drupal_version%.*.*} -ge "8" ]; then \
>     drush cr; \
>   elif [ ${drupal_version%.*} == "7" ]; then \
>     drush cc all; \
>   else \
>     echo "could not clear cache for found Drupal Version ${drupal_version}"; \
>   fi
bash: [: 7.59: integer expression expected
'all' cache was cleared.  
```

tobybellwood: (and whilst it does return 0 eventually, I'm not sure how the current task system might handle the bash error)

tobybellwood: bugged me just enough to switch the check logic.  At least it's now Drupal 10 safe :relieved:

seanhamlin: +1 thanks Toby.

PR #2671, opened 2021-05-22
webhooks2tasks harbor v2 Webhook integration fix

### LINKED ISSUE #2670: Issue with Harbor upstream image scan webhooks
Presently there is an upstream issue affecting some installations of Harbor - the problem is described here https://github.com/goharbor/harbor/issues/14585

For our purposes we're receiving misleading data from the Trivy scan in the Harbor webhook payload, telling us that a scan is still running when it is, in fact, complete. We don't rely on this payload _except_ to check the scan status. But we know that Harbor will not send this particular webhook without the scan both being complete and successful (incomplete scans are a different webhook event)

bomoko:
Here we remove the check discussed in #2670 - namely the part of the Harbor webhook payload that describes the current state of the scan (we check for a "success" in the "scan_overview" section (which is returning inaccurate information).

# Checklist

# Closing issues

closes #2670


bomoko:
moving to lagoon-one

PR #2681, opened 2021-05-26
Allow skipping Trivy scanning when building Lagoon

cdchris12:
# Checklist

This PR allows for the use of the `SKIP_SCAN` variable when building Lagoon via the Makefile, which will allow users to skip the requirement to have every image scanned by Trivy, allowing these images to be built significantly faster. 

This PR also includes a change to the docker build command string which suppresses the "Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them" text that is generated with every call to `docker build` currently.


tobybellwood (approved):
LGTM if it helps!

PR #2685, opened 2021-05-27
Backporting Harbor fixes to master

cdchris12:
# Checklist

This PR fixes the issue with the HarborSync script not working after the migration to Harbor v2.1.x

### DIFF
--- a/services/api/src/resources/project/harborSetup.ts
+++ b/services/api/src/resources/project/harborSetup.ts
@@ -20,6 +20,9 @@ const lagoonWebhookAddress = R.compose(
   R.propOr('', 'LAGOON_ROUTES'),
 )(process.env) as string;
 
+const apiVersion = R.propOr('v2.0', 'HARBOR_API_VERSION', process.env);
+// Use an empty string for backwards compatibility with Harbor version 1.x.x
+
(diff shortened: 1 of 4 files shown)


services/api/src/resources/project/harborSetup.ts
```diff
@@ -226,8 +226,8 @@ async function resetHarborWebhook(sqlClient: MariaClient, harborClient, lagoonPr
           }
         ],
         event_types: [
```
tobybellwood: my one question here is whether we need to wrap this in a HARBOR_API_VERSION variable check to make it compatible with Harbor v1?

cdchris12: Good catch. I'll write that in

tobybellwood: giphy-downsized

tobybellwood:
The test failure here is real
building locally with `HARBOR_API_VERSION=` set to an empty string in docker-compose is somehow triggering it to mangle the API URL in harborclient
```
The push refers to repository [172.16.0.1:8084/ci-api-control-k8s/pr-1/node]
Get http://172.16.0.1:8084/v2/: dial tcp 172.16.0.1:8084: connect: connection refused
```

PR #2686, opened 2021-05-27
feat: Add `EnvironmentsByKubernetes` api query

### LINKED ISSUE #2684: Add API query to list all projects on a cluster
- New `projectsByKubernetes` query
- Filter by `id`, `createdAfter`
- Sort by `name`, `created`

rocketeerbkw:
# Checklist

# Closing issues

closes #2684


rocketeerbkw:
This query can be executed by admins and regular users. For regular users we have to loop over every project and do a keycloak check to ensure they can view the openshift data (by default, only project maintainers can view it).

This will be really slow if anyone has a lot of projects and put more strain on keycloak, so I'm wondering if we should even allow this for non-admins. The upside is we're not using this in the UI or the CLI, but do we wanna let those horses out of the barn right now?

rocketeerbkw:
I only added `projectsByKubernetes` and didn't add `projectsByOpenshift` because we want to get rid of openshift references eventually anyway.

tobybellwood:
This makes sense, but a pivot to EnvironmentsByKubernetes would account for prod/dev split DeployTargets.

@CGoodwin90 to refactor

services/api/src/typeDefs.js
```diff
@@ -1083,6 +1089,10 @@ const typeDefs = gql`
     """
     projectByName(name: String!): Project
     """
+    Returns all Environment Objects for a specified Kubernetes
+    """
+    environmentsByKubernetes(kubernetes: KubernetesInput!, order: EnvOrderType, createdAfter: String): [Environment]
```
shreddedbacon: The `allEnvironments` query has a filter for the `type` too. I think it could be useful to add this filter to this resolver too. Then you could do all production environments on cluster X

services/api/src/resources/environment/resolvers.ts
```diff
@@ -308,6 +309,68 @@ export const getEnvironmentByKubernetesNamespaceName: ResolverFn = async (
...
+    { kubernetes, order, createdAfter },
+    { sqlClientPool, hasPermission, models, keycloakGrant }
+  ) => {
+    const openshift = await openshiftHelpers(
+      sqlClientPool
+    ).getOpenshiftByOpenshiftInput(kubernetes);
+
+    let userProjectIds: number[];
+    try {
+      await hasPermission('project', 'viewAll');
```
shreddedbacon: You only need to do the `openshift:viewAll` permission check here, but that means the query will only work for legacy tokens until #3437 lands. But this is the same behaviour that `allEnvironments` query has with `environment:viewAll` which is also resolved by the same PR.

shreddedbacon (changes requested):

services/api/src/resources/environment/resolvers.ts
```diff
@@ -308,6 +309,71 @@ export const getEnvironmentByKubernetesNamespaceName: ResolverFn = async (
...
+  try {
+    await hasPermission('openshift', 'viewAll');
+  } catch (err) {
+    if (!keycloakGrant) {
+      logger.warn('No grant available for getEnvironmentsByKubernetes');
+      return [];
+    }
+
+    // Only return projects the user can view
+    userProjectIds = await models.UserModel.getAllProjectsIdsForUser({
```
shreddedbacon: For non-admin users using this query, it fails because the way this works was changed in the keycloak optimisations PR.

You'll need to update this to use the following
```
const userProjectRoles = await models.UserModel.getAllProjectsIdsForUser({
  id: keycloakGrant.access_token.content.sub,
}, keycloakUsersGroups);
userProjectIds = getUserProjectIdsFromRoleProjectIds(userProjectRoles);
```

You'll also need to update the resolver function to include `keycloakUsersGroups` and also import 
```
import { getUserProjectIdsFromRoleProjectIds } from '../../util/auth';
```

shreddedbacon (changes requested):

services/api/src/resources/openshift/helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+      const notEmpty = R.complement(R.anyPass([R.isNil, R.isEmpty]));
+      const hasId = R.both(R.has('id'), R.propSatisfies(notEmpty, 'id'));
+      const hasName = R.both(R.has('name'), R.propSatisfies(notEmpty, 'name'));
+
+      const openshiftFromId = asyncPipe(R.prop('id'), async id => {
+        const rows = await query(sqlClientPool, Sql.selectOpenshift(id));
+        const openshift = R.prop(0, rows);
+
+        if (!openshift) {
+          throw new Error('Unauthorized');
```
shreddedbacon: Throw the same `invalid input for kubernetes` error here

shreddedbacon (changes requested):
One last thing, but otherwise all good!

shreddedbacon (approved):

PR #2693, opened 2021-06-04
Fact api changes for fact search

### LINKED ISSUE #2483: Types for Facts
It would be useful to introduce typing for Facts. 

This would help to determine (at least) how they should be displayed, but looking further down the line, introducing types to facts will greatly improve our ability to search and compare.

For example, we may wish to have a "URL" type for facts that puts a link in the facts listing page, enabling a user to directly visit a page associated with an environment.

timclifford:
# Checklist

_Main changes to api_
-  adds new fact search query that returns projects/environments via getProjectsByFactSearch and  getEnvironmentsByFactSearch resolvers. These are available as `projectsByMetadata` and `environmentsByFactSearch` 
- adds 'fact_reference' field which will allow for better fact categorisation
- adds concept of 'key' facts so we can give priority to particular facts
- updates 'knex" to `v0.95` 

# Closing issues

closes #2483


services/api/src/resources/fact/resolvers.ts
```diff
@@ -139,3 +228,169 @@ export const deleteFactsFromSource: ResolverFn = async (
...
+
+  const rows = await query(
+    sqlClientPool,
+    Sql.selectFactReferenceByDatabaseId(insertId)
+  );
+
+  return R.prop(0, rows);
+};
+
+export const deleteFactReferenceById: ResolverFn = async (
```
bomoko: I'm wondering if, from a usability perspective, it might not make sense to change this slightly?

Instead of deleting by fact reference id - maybe removing a fact should just be a function of the environment id and the fact's name? There really should be no real reason for the user to know the fact reference's id, right?

bomoko:
I think it would be good to get @rocketeerbkw's eyes on on this too. We're adding a bit to the API and making permission changes.

PR #2711, opened 2021-06-23
Experimental support for external secret storage

### LINKED ISSUE #469: Assess secret storage systems for Lagoon
Currently it's not possible to store secret environment variables across an entire OpenShift project. Each new environment that is created for a new PR or branch needs to have ENV vars set manually in the deployment. Could we move to Vault to manage this instead of env?

Based on some investigations in our project, the following steps outline what may be needed to achieve a production level integration of Vault.

# Install Consul.
	** https://github.com/kubernetes/charts/tree/master/stable/consul
# Install Vault.
	** https://github.com/kubernetes/charts/tree/master/incubator/vault
# Install the Goldfish Vault Web UI.
	** https://github.com/kubernetes/charts/tree/master/incubator/goldfish
	** https://github.com/Caiyeon/goldfish
# Configure the Vault Kubernetes Auth Backend.
	** https://www.vaultproject.io/docs/auth/kubernetes.html
	** This will allow Kubernetes service accounts to authenticate to Vault.
# For each application that is to use Vault stored secrets.
	** Create a `ConfigMap` with a file defining the secrets to be consumed.
	** Change the Dockerfile ENTRYPOINT for the workload so that `vaultenv` is used to fetch the secrets from Vault and store them in environment variables.
	** https://github.com/channable/vaultenv
	** The section “Exposing Secrets to Applications” from this article goes into more detail about using `vaultenv` - https://www.elastic.co/blog/kubernetes-vault-integration-devops-team

smlx:
This PR implements experimental support for external secret storage providers.

# Closing issues

Possibly closes: #469

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -954,7 +954,7 @@ KIND_VERSION = v0.11.1
 GOJQ_VERSION = v0.12.3
 CHART_TESTING_VERSION = v3.4.0
 KIND_IMAGE = kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6
-TESTS = [api,features-kubernetes,features-kubernetes-2,features-api-variables,active-standby-kubernetes,nginx,drupal-php73,drupal-php74,drupal-postgres,python,gitlab,github,bitbucket,node-mongodb,elasticsearch]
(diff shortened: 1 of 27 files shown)


tobybellwood:
We don't need to keep this open. We can always come back to it later (albeit in a different repo!)

PR #2713, opened 2021-06-25
Simplify retest logic

smlx:
This change speeds up retest logic and makes it less likely to fail.
It also removes the hard-coded chart-testing version.

# Closing issues

n/a


tobybellwood (approved):
Yup - now the API-data-watcher-pusher cycle is faster, this works well

PR #2727, opened 2021-06-29
Allow the make jobserver to co-ordinate with sub-makes

smlx:
# Checklist

Sub-make processes [automatically co-ordinate](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html) with the parent `make` jobserver to maintain the top-level `-jN` command. So remove the hard-coded value here.

# Closing issues

n/a


tobybellwood:
ah -I think this is a leftover from me trying to super-parallelise docker pushes...

PR #2731, opened 2021-06-29
Remove some sleeps from test suite

smlx:
# Checklist

This change makes local test iterations much faster. However I don't know if there is another reason for these sleeps to be there @tobybellwood?

# Closing issues

n/a


tobybellwood:
ah - random-wait was the successor to the 30s sleep - so that can go.  My experience was that the fastest tests off the mark (nginx, node, python) would fail if they triggered immediately - and the 30s massively reduced that risk.

The random wait is to avoid the tests (when running multiple) crashing into each other as much and spread them out a little between 30s-120s.  Of course, the errors will still happen (because of the other keycloak issue)

smlx:
OK I re-added the random waits in the test as discussed.

PR #2745, opened 2021-07-01
Updating k8up template backupcommand and file-extension annotations

cdchris12:
# Checklist

This PR resolves some issues discovered with our current implementation of k8up backup commands and the resulting file extensions.

### DIFF
--- a/images/oc-build-deploy-dind/openshift-templates/mongo/deployment.yml
+++ b/images/oc-build-deploy-dind/openshift-templates/mongo/deployment.yml
@@ -71,7 +71,7 @@ objects:
           branch: ${SAFE_BRANCH}
           project: ${SAFE_PROJECT}
         annotations:
-          appuio.ch/backupCommand: /bin/sh -c "/bin/busybox tar -cf - -C /data/db ."
+          appuio.ch/backupCommand: /bin/sh -c "tar -cf - -C /data/db ."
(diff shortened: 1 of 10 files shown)


images/oc-build-deploy-dind/openshift-templates/mongo/deployment.yml
```diff
@@ -71,7 +71,7 @@ objects:
           branch: ${SAFE_BRANCH}
           project: ${SAFE_PROJECT}
         annotations:
-          appuio.ch/backupCommand: /bin/sh -c "/bin/busybox tar -cf - -C /data/db ."
+          appuio.ch/backupCommand: /bin/sh -c "tar -cf - -C /data/db ."
```
Schnitzel: @cdchris12 
are these changes expected? seems to me like it's another issue?

Schnitzel (changes requested):

images/oc-build-deploy-dind/openshift-templates/mongo/deployment.yml
cdchris12: We had a discussion a while ago about the use of `tar` vs `/bin/busybox tar` and came to the conclusion that either one would work for us. It's more compatible to let sh use whatever `tar` binary is installed on the system, as busybox isn't installed in every container

Schnitzel (approved):

PR #2763, opened 2021-07-14
Performance tweaks for Deployments UI page

rocketeerbkw:
# Checklist

This PR makes some changes to the API and UI to improve performance for environments have a lot of deployments (e.g. > 700). The API optimized SQL queries, table indexes, resolver execution and added an optional `limit` parameter. The UI API queries have been optimized to not fetch duplicate data and reduce the amount of work the API has to do and to only load 25 deployments (set via environment variable `LAGOON_UI_DEPLOYMENTS_LIMIT`).

Locally these changes combined make a page with 746 deployments go from not loading at all to loading in 10 seconds with a cold cache and ~1 second with a hot cache.

Additionally the API now has a "graceful shutdown" method which properly closes the mariadb connection pool. This helps reduce errors when developing locally with frequent API restarts and should avoid potential issues in production in cases where the API pod is shut down.

# Closing issues

n/a


bomoko (approved):

bomoko:
LGTM - tested by generating a good few thousand deployments locally. Limit filtering working like a charm.

PR #2768, opened 2021-07-16
adds p-N class opendistro groups

### LINKED ISSUE #2764: Kibana login redirect on too many groups
Kibana ends up in an endless login redirect if a user has too many groups

bomoko:
This is a WIP to add p-N class role names in order to shorten what's passed to the lagoon-opensecurity-distro client.

This is a potential way of addressing #2764 without going down the hashing route (which, for short hashes, turns out to be a little complex).

Here is the unwound code for Keycloak - all keycloak roles should keep their current names - the only thing that changes is what we send along.

```
var ArrayList = Java.type("java.util.ArrayList");
var groupsAndRoles = new ArrayList();
var forEach = Array.prototype.forEach;

// add all groups the user is part of
forEach.call(user.getGroups().toArray(), function(group) {
  // remove the group role suffixes
  //lets check if the group has a parent if this is a child
  if(group.getFirstAttribute("type") == "role-subgroup") {
    var parent = group.getParent();
    if(parent.getFirstAttribute("type") == "project-default-group") {
        var projectIds = parent.getFirstAttribute("lagoon-projects");
        if(projectIds !== null) {
            forEach.call(projectIds.split(","), function(g) {
              groupsAndRoles.add("p" + g);  
            });
            return;
        }
    }
  }
 
  var groupName = group.getName().replace(/-owner|-maintainer|-developer|-reporter|-guest/gi,"");
  groupsAndRoles.add(groupName);
  return;
});

// add all roles the user is part of
forEach.call(user.getRoleMappings().toArray(), function(role) {
   var roleName = role.getName();
   groupsAndRoles.add(roleName);
});

exports = groupsAndRoles;
```

# Checklist

# Closing issues

closes #2764


bomoko:
Hey @Schnitzel - so rather than try have the `syncGroup` code determine whether it should be creating a short project name for the role, I've made it possible to pass a tenant name (although I think I might change the function's name `syncGroupWithExistingTenant`).
I _think_ this addresses your worry.

I've also added in the checks for the default project groups into the keycloak mapper.

bomoko:
I've just realized that one other thing we could do is actually simply add a "short-name" attribute to groups (now that I know that we have access to group attributes) and do away with the dynamic calculation of the short name completely.

services/keycloak/start.sh
```diff
@@ -90,7 +90,7 @@ function configure_opendistro_security_client {
...
... is a child\n  if(group.getFirstAttribute(\"type\") == \"role-subgroup\") {\n    var parent = group.getParent();\n    if(parent.getFirstAttribute(\"type\") == \"project-default-group\") {\n        var projectIds = parent.getFirstAttribute(\"lagoon-projects\");\n        if(projectIds !== null) {\n            forEach.call(projectIds.split(\",\"), function(g) {\n              groupsAndRoles.add(\"p\" + g);  \n            });\n            return;\n        }\n    }\n  }\n \n  var groupName = group.getName().replace(\/-owner|-maintainer|-developer|-reporter|-guest\/gi,\"\");\n  groupsAndRoles.add(groupName);\n  return;\n});\n\n\/\/ add all roles the user is part of\nforEach.call(user.getRoleMappings().toArray(), function(role) {\n   var roleName = role.getName();\n   groupsAndRoles.add(roleName);\n});\n\nexports = groupsAndRoles;","id.token.claim":"true","access.token.claim":"true","userinfo.token.claim":"true","multivalued":"true","claim.name":"groups","jsonType.label":"String"},"name":"groups","protocolMapper":"oidc-script-based-protocol-mapper"}' | /opt/jboss/keycloak/bin/kcadm.sh create -r ${KEYCLOAK_REALM:-master} clients/$CLIENT_ID/protocol-mappers/models --config $CONFIG_PATH -f -
```
Schnitzel, line 93: @bomoko 
the only issue here is that this will be only executed if the `lagoon-opendistro-security` does not exist, see
https://github.com/uselagoon/lagoon/pull/2768/files#diff-8a14c757e38a4f4906e6dd229db8406de31fec05d5b8c5e1c6b0514e5a8c1f19R83

so in an existing keycloak installation this will not be updated. 
I'm ok in keeping this in an update documentation, but maybe if you have time to figure out how to update this automatically on a keycloak start it would be epic :)

bomoko: Ah, good call @Schnitzel - I'll try work that out asap.

PR #2793, opened 2021-07-30
Adding new "backupEnvironment" API endpoint

cdchris12:
# Checklist

# Closing issues

### DIFF
--- a/node-packages/commons/src/api.ts
+++ b/node-packages/commons/src/api.ts
@@ -967,25 +967,31 @@ export async function getEnvironmentByOpenshiftProjectName(
   return result;
 }
 
-export async function getBackupEnvironment(
-  openshiftProjectName: string,
-  baas_bucket_name: string
-): Promise<any[]> {
+export async function getBackupEnvironment( backupEnvironmentInput: {
+  openshiftProjectName?: string;
+  baas_bucket_name?: string;
+}): Promise<any[]> {
   const result = await graphqlapi.query(`
     {
-      backupEnvironment(openshiftProjectName: "${openshiftProjectName}", baas_bucket_name: "${baas_bucket_name}") {
-        id,
-        name,
-        project {
-          name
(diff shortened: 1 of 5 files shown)


node-packages/commons/src/api.ts
```diff
@@ -967,6 +967,31 @@ export async function getEnvironmentByOpenshiftProjectName(
   return result;
 }
 
+export async function getBackupEnvironment(
+  openshiftProjectName: string,
+  baas_bucket_name: string
+): Promise<any[]> {
+  const result = await graphqlapi.query(`
+    {
+      backupEnvironment(openshiftProjectName: "${openshiftProjectName}", baas_bucket_name: "${baas_bucket_name}") {
```
rocketeerbkw: String interpolation here is pretty dirty (ya, I know you just copy/pasted, sorry). The library supports passing in parameters as a part of the query itself, so that the api can do validation on the server side. It would also prevent some string double escaping issues (acts more like prepared statements). A better example to copy would be [`allProjectsInGroup`](https://github.com/uselagoon/lagoon/blob/main/node-packages/commons/src/api.ts#L638-L653). You would create an outer query that has typed parameters, then pass the paramters as the second argument to `graphqlapi.query`.

node-packages/commons/src/api.ts
```diff
@@ -967,6 +967,31 @@ export async function getEnvironmentByOpenshiftProjectName(
...
+        id,
+        name,
+        project {
+          name
+        }
+      }
+    }
+  `);
+
+  if (!result || !result.environmentByOpenshiftProjectName) {
```
rocketeerbkw: The `result` is a data structure that matches the fields you queried, so here `result.environmentByOpenshiftProjectName` will always be undefined because the query you are running is now called `backupEnvironment`.

rocketeerbkw (changes requested):

PR #2800, opened 2021-08-09
parsed user logs and transport to rabbitMQ and logstash

### LINKED ISSUE #2807: API request logs should include audit information
In order to be able to better understand where every Lagoon request comes from and support future auditing efforts, we should add the following data to each request log wherever possible:

- IP the request originated from
- ~Lagoon user which made the request (or a value which can be uniquely mapped to a single user)~ Already completed in #2690

timclifford:
# Checklist

# Closing issues

closes #2807 closes #2800


Schnitzel:
@timclifford 
mhh is there a specific reason that we use `rabbit-chatter` as a connection system to rabbitmq? the api already has rabbitmq connections open via https://github.com/uselagoon/lagoon/blob/de9a20c02a165c3976ca24cce586e676c3d83330/node-packages/commons/src/logs.ts
I would like to prevent to not use too many rabbitmq connections, especially at it seems that `rabbit-chatter` closes the connection (see https://github.com/TBear79/rabbit-chatter)

timclifford:
@Schnitzel ahh good point - makes sense. Updated the custom transporter to use the [existing mq connection for lagoon-logs](https://github.com/uselagoon/lagoon/pull/2800/commits/e8e7f2396109f5b5606f0ca492acec095992122f#diff-41857eb52e5629d7454e16df16319dbbce69d27d4ea6c4a8c3d5dabfd0441f51R36-R43)

services/api/src/resources/backup/resolvers.ts
```diff
@@ -151,6 +153,8 @@ export const deleteBackup: ResolverFn = async (
   pubSub.publish(EVENTS.BACKUP.DELETED, R.prop(0, rows));
 
   userActivityLogger.user_action(`User deleted backup '${backupId}'`, {
+    project: '',
```
rocketeerbkw, line 156: For the majority of our resolvers, we have the capability of knowing what the project is, even if we don't already load that data in each resolver. For `deleteBackup`, we already load the project id in `perms`, so we can also know the name, for example.

I think we loose a lot of the utility of `sendToLagoonLogs` if we don't add the project when we can. If we added project for everything it would be a lot easier (pretty trivial actually?) to show a log in the UI of all related actions.

This def requires a lot more work, so we could tackle in a separate issue?

services/api/package.json
```diff
@@ -53,12 +54,14 @@
     "morgan": "^1.9.0",
     "newrelic": "^6.9.0",
     "node-cache": "^4.2.1",
+    "rabbit-chatter": "^2.0.0",
```
rocketeerbkw, line 57: I didn't see this being imported anywhere in this PR is it needed still?

rocketeerbkw (commented):
I don't see any un-mergeable issues, if some improvements can be addressed now that's great otherwise it can be a follow up task.

PR #2802, opened 2021-08-10
Adds redis caching to bb sync script

bomoko:
This PR potentially takes some of the pressure of the API when doing a bit bucket sync

It pulls the full user list down from BitBucket for every project, creates a hash from the email and permission list currently on that project on the BB side, and compare that to a cached hash (in redis) - if there is a difference, we continue with the sync as usual, if there has been no change to the hash in the last 24 hours, we know we can skip processing the current project.

# Checklist


Schnitzel:
@bomoko can you fix the merge conflict? thanks!

PR #2806, opened 2021-08-17
Add support for project routerPatterns

### LINKED ISSUE #2791: implement a project-level router pattern
It would be super handy for some clients that have a managed CDN to be able to set up a project level router pattern. This would ideally live as a project level setting in the API.

shreddedbacon:
# Checklist

# Closing issues

closes #2791


shreddedbacon:
Still need to add environment variable override support into `x-build-deploy-dind` images. But this may not be required, just gives an additional place to be more specific at targeting where it is applied.

PR #2819, opened 2021-08-26
add drupal-keepup as optional test routine for local development

tobybellwood:
Build a local Lagoon with a running Drupal project like so

`make kind/test TESTS=[drupal-keepup]`

It should also be safe to use with other CI routines as it completes successfully - it just doesn't get torn down after testing completes.


tobybellwood:
Closing -will consider reworking as a variable that can be set for any test

PR #2820, opened 2021-08-27
Custom Backup and Restore Location Support

### LINKED ISSUE #2680: Allow project-level BaaS bucket & credentials
Currently there is only the option to use the same s3-compatible bucket for all projects on a cluster, this should be configurable so that a project may bring it's own backup storage

cdchris12:
# Checklist

This PR introduces the logic necessary to support custom, user-defined backup and restore locations for Lagoon's Backup As A Service.

# Closing issues

Closes #2680


docs/using-lagoon-advanced/backups.md
```diff
@@ -25,3 +25,8 @@ Backups stored in Restic will be tracked within Lagoon, and can be recovered via
 
 Backups of development environments are attempted nightly and are strictly a best effort service.
 
+## Custom Backup and/or Restore Locations
+
+Lagoon supports custom backup and restore locations via the use of the "[Custom Backup Settings](https://github.com/uselagoon/lagoon/blob/main/docs/using-lagoon-advanced/environment-variables.md#custom-backup-settings)" and/or "[Custom Restore Settings](https://github.com/uselagoon/lagoon/blob/main/docs/using-lagoon-advanced/environment-variables.md#custom-restore-settings)" variables stored in the Lagoon API for each project.
+
+***Note that using these variables may affect any SLAs in effect for Lagoon managed backups***
```
rocketeerbkw: Does Lagoon the open source project have SLAs? Not sure where we are with Lagoon vs amazee.io docs, but maybe this note makes more sense in amazee.io? @tobybellwood or @AlannaBurke?

rocketeerbkw (changes requested):
Is it confusing that we would have two environment variables for the backup bucket name `LAGOON_BAAS_BUCKET_NAME` and `LAGOON_BAAS_CUSTOM_BACKUP_BUCKET`?

docs/using-lagoon-advanced/backups.md
cdchris12: This concern was exactly why I preferred to leave this more vague. The fact of the matter is that once any user of a Lagoon project chooses to own the backup and/or restore locations, we can no longer rely on Lagoon automatically creating and maintaining access to them. If you've a better to way to say that, please feel free to suggest an edit. :)

services/keycloak-db/Dockerfile
```diff
@@ -13,5 +13,5 @@ ENV MARIADB_DATABASE=keycloak \
 
 COPY my_query-cache.cnf /etc/mysql/conf.d/my_query-cache.cnf
 USER root
-RUN chown -R mysql /etc/mysql/conf.d/
+RUN fix-permissions /etc/mysql/conf.d/
```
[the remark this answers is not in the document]
cdchris12: I also thought that, but this was specifically keeping the keycloak-db container from starting if built locally. This is actually just a cherrypick from a branch that's already merged in upstream, so I think we can mark this as resolved.

services/api/src/resources/backup/resolvers.ts
```diff
@@ -13,27 +13,70 @@ import { Sql } from './sql';
...
+    })
+    if (lagoonBaasCustomRestoreEndpoint) {
+      lagoonBaasCustomRestoreEndpoint = lagoonBaasCustomRestoreEndpoint.value
+    }
+    let lagoonBaasCustomRestoreBucket = projectEnvVars.find(obj => {
+      return obj.name === "LAGOON_BAAS_CUSTOM_RESTORE_BUCKET"
+    })
+    if (lagoonBaasCustomRestoreBucket) {
+      lagoonBaasCustomRestoreBucket = lagoonBaasCustomRestoreBucket.value
+    }
```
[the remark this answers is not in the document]
cdchris12: We don't actually need those variables, as we can already infer them from the `s3parts` variable we're creating [here](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/backup/resolvers.ts#L23-L27). We do, however, need to make sure those variables are set, as all four variables are required to be set for this functionality to function correctly.

services/api/src/resources/backup/resolvers.ts
```diff
@@ -13,27 +13,70 @@ import { Sql } from './sql';
...
 import { Sql as environmentSql } from '../environment/sql';
 import { Helpers as environmentHelpers } from '../environment/helpers';
+import { Helpers as projectHelpers } from '../project/helpers';
+import { getEnvVarsByProjectId } from '../env-variables/resolvers';
 import { EVENTS } from './events';
+import { logger } from '../../loggers/logger';
 
 export const getRestoreLocation: ResolverFn = async (
-  { restoreLocation },
+  _root,
```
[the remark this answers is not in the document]
cdchris12: Updated all the args here to make them more readable and no longer implicitly private

services/api/src/resources/backup/resolvers.ts
```diff
@@ -13,27 +13,70 @@ import { Sql } from './sql';
...
   _args,
-  _context
+  _context,
```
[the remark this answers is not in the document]
cdchris12: Updated as requested

rocketeerbkw (approved):

docs/using-lagoon-advanced/backups.md
AlannaBurke: How about:
***Please note that these settings are for use with open-source Lagoon installations only. If you are an amazee.io customer, please see your amazee.io administrator before making these changes.***

rocketeerbkw: @cdchris12 Since it looks like you'll have to resolve a conflict anyway... we settled on the following language;

> Proceed with caution: Setting these variables will override backup/restore storage locations that may be configured at a cluster level. Any misconfiguration will cause backup/restore failures.

Would that still cover your initial concerns?

cdchris12: That works for me. I'll commit that now :)

PR #2822, opened 2021-08-30
correctly fix permissions for my_query-cache.cnf

Schnitzel:
this fixes an error when the keycloak-db is started as rootless (how it is normal during a lagoon-core install), causing the whole testing pipeline to fail.

```2021/08/30 15:47:45 [ ERROR ] Error while parsing '/etc/mysql/conf.d/my_query-cache.cnf': open /etc/mysql/conf.d/my_query-cache.cnf: permission denied```

### DIFF
--- a/services/keycloak-db/Dockerfile
+++ b/services/keycloak-db/Dockerfile
@@ -13,5 +13,5 @@ ENV MARIADB_DATABASE=keycloak \
 
 COPY my_query-cache.cnf /etc/mysql/conf.d/my_query-cache.cnf
 USER root
-RUN chown -R mysql /etc/mysql/conf.d/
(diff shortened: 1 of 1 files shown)


shreddedbacon:
I'm hitting this too when running tests, did something change in the latest `mariadb-10.4` images @tobybellwood ?

tobybellwood:
not as far as I can tell - I think the error was introduced with the query cache (https://github.com/uselagoon/lagoon/pull/2797/files)?  Does setting the fix-permissions work for you (or using this image `testlagoon/keycloak-db:pr-2822`?

shreddedbacon:
Yes, making the change in this PR fixes the keycloak-db starting issue in the tests

PR #2829, opened 2021-09-07
Support for multiple deployment targets per project

shreddedbacon:
# Checklist

This is a WIP for supporting multiple deployment targets per project.


shreddedbacon:
@tobybellwood can you think of anything that I may not have covered off in this, or anything else I should call out somewhere else in our docs with the change to environments now containing the openshift.

Or any else that might seem vital that I may have overlooked.

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -1477,6 +1546,9 @@ CALL add_enum_webhook_to_type_in_project_notification();
 CALL add_index_for_deployment_environment();
 CALL add_index_for_task_environment();
 CALL add_router_pattern_to_project();
+CALL add_openshift_to_environment();
+CALL add_openshift_project_pattern_to_environment();
+CALL add_deployments_disabled_to_project();
```
tobybellwood, line 1551: should we also add `migrate_project_openshift_to_environment` in here, now that it's rerun-safe?

shreddedbacon: I'm ok with that, if you think it looks ok.

PR #2831, opened 2021-09-08
Adding startingDeadlineSeconds to all cli native cronjob templates

cdchris12:
# Checklist

- [] Affected Issues have been mentioned in the Closing issues section
- [] Documentation has been written/updated

This PR adds the `startingDeadlineSeconds` value to both native cronjob templates for cli cron tasks. This prevents stale jobs from preventing the cronjob from running altogether.

# Closing issues

### DIFF
--- a/images/kubectl-build-deploy-dind/helmcharts/cli-persistent/templates/cronjob.yaml
+++ b/images/kubectl-build-deploy-dind/helmcharts/cli-persistent/templates/cronjob.yaml
@@ -13,6 +13,7 @@ spec:
   concurrencyPolicy: Forbid
   successfulJobsHistoryLimit: 0
   failedJobsHistoryLimit: 1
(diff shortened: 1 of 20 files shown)


tobybellwood:
As I understand it, this will only retry the cronjob for 240 seconds after it's scheduled time.  As cronjob runs start every ~10s, this will mean 24 failures, which should result in a "missed schedule" error (and not the exceeded retries error) - per https://stackoverflow.com/questions/51065538/what-does-kubernetes-cronjobs-startingdeadlineseconds-exactly-mean

As we have a forbid concurrency policy in place, this will cause cronjobs to try to schedule for this 4 minute period, but the retries will have finished and the job failed by the time the next schedule comes around (as long as it's over 5 minutes)

End users may be able to use the fail messages to trigger some sort of alert on long-running crons

PR #2614, opened 2021-04-19
WIP - Stripping out Billing Code

justinlevi:
# Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for changelog and subsystem label(s) applied

Explain the **details** for making this change. What existing problem does the pull request solve?

# Closing issues

Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

PR #2837, opened 2021-09-15
Removing Billing Code from API

### LINKED ISSUE #2687: Remove Billing Code from Lagoon
IO Billing code doesn't belong in lagoon as it is specific to us. A plugin architecture makes more sense.

justinlevi:
related:
* https://github.com/uselagoon/lagoon/pull/2614
* https://github.com/uselagoon/lagoon/pull/2688
* https://github.com/uselagoon/lagoon/pull/2794

# Checklist

This PR removes all billing related code from the API in favor of an external proprietary solution. 

# Closing issues

`closes #2687`


justinlevi:
@rocketeerbkw Could you take a look at this one?

node-packages/commons/src/tasks.ts
```diff
@@ -325,9 +321,6 @@ const getControllerBuildData = async function(deployData: any) {
   if (monitoringConfig != "invalid"){
     alertContactHA = monitoringConfig.uptimerobot.alertContactHA || ""
     alertContactSA = monitoringConfig.uptimerobot.alertContactSA || ""
-    if (monitoringConfig.uptimerobot.statusPageId) {
-      uptimeRobotStatusPageIds.push(monitoringConfig.uptimerobot.statusPageId)
-    }
```
rocketeerbkw: This uptime robot data exists on the `openshift` field, I only remember discussing it being OK to remove from the billing group. Is it OK to remove here too?

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -1468,7 +1449,6 @@ CALL add_fact_type_to_environment_fact();
 CALL add_fact_category_to_environment_fact();
 CALL add_fact_key_to_environment_fact();
 CALL add_metadata_to_project();
-CALL add_min_max_to_billing_modifier();
```
rocketeerbkw, line 1540: New installs won't get a `billing_modifier` table but are we also comfortable deleting existing ones? Could add a new migration to "drop if exists."

This could be a follow up ticket to leave data around "just in case."

services/api/src/models/group.ts
```diff
@@ -117,20 +99,13 @@ export const Group = (clients: {
 
   const transformKeycloakGroups = async (
     keycloakGroups: GroupRepresentation[]
-  ): Promise<Group[] | BillingGroup[]> => {
+  ): Promise<Group[]> => {
     // Map from keycloak object to group object
     const groups = keycloakGroups.map((keycloakGroup: GroupRepresentation):
-      | Group
-      | BillingGroup => ({
+      | Group => ({
```
rocketeerbkw: I think you can get rid of the OR `|` here too? `const groups = keycloakGroups.map((keycloakGroup: GroupRepresentation): Group => ({`

services/api/src/resolvers.js
```diff
@@ -314,19 +294,13 @@ const resolvers = {
   GroupInterface: {
     __resolveType(group) {
       switch (group.type) {
-        case 'billing':
-          return 'BillingGroup';
         default:
           return 'Group';
       }
```
rocketeerbkw: Would you mind just getting rid of the `switch` and `return 'Group'`?

services/api/src/typeDefs.js
```diff
@@ -414,26 +405,13 @@ const typeDefs = gql`
...
-    id: String
-    name: String
-    type: String
-    groups: [GroupInterface]
-    members: [GroupMembership]
-    projects: [Project]
-  }
-
-  type BillingGroup implements GroupInterface {
+  type Group {
```
rocketeerbkw: I think `Group` still needs to `implements GroupInterface` to avoid BC break.

services/keycloak/start.sh
```diff
@@ -1740,7 +1674,6 @@ function configure_keycloak {
     add_group_viewall
     add_deployment_cancel
     configure_task_cron
-    add_billing_modifier
```
rocketeerbkw, line 1743: Should we also add a function to delete the permissions/resources from keycloak? And write something to delete the billing groups that exist in keycloak?

This could be a follow up ticket to leave data around "just in case."

rocketeerbkw (changes requested):

rocketeerbkw (approved):
Rebased on `main`, resolved change requests, tested against a prod dataset.

PR #2841, opened 2021-09-21
Add a message in the UI that shows how many results are displayed

### LINKED ISSUE #2842: Indicate when backups/task list is truncated
In order to improve the performance of certain Lagoon UI screens, the lists were truncated. See https://github.com/uselagoon/lagoon/pull/2771

A nice UX change to improve the quality of life of developers (in lieu of actual pagination) would be to indicate that the list is truncated (when the number of backups/tasks exceed the limit).

There is a PR for this https://github.com/uselagoon/lagoon/pull/2841

shreddedbacon:
# Checklist

On the `tasks`, `backups`, and `deployments` pages, we limit the number of results that are displayed to users. We should let them know the results have been limited.

It would be better if this was a drop down selection box that could allow the user to select how many results are returned, rather than defining a hard value.

# Closes

Fixes #2842.

### DIFF
--- a/services/ui/src/pages/backups.js
+++ b/services/ui/src/pages/backups.js
@@ -108,6 +108,9 @@ export const PageBackups = ({ router }) => (
                 <Backups backups={environment.backups} />
               </div>
             </div>
+            <div className="message-wrapper">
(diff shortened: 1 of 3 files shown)


shreddedbacon:
I think #2843 is better

PR #2845, opened 2021-09-23
WIP: add support for service specific autogenerated route disabling

shreddedbacon:
…pose.yml

 

# Checklist

# Closing issues


tobybellwood:
On review of this, it appears that the routes are still being created for the false condition.

This is because there are two `kubectl apply` cycles in each build.

The first is targeted at the services and routes created in `/kubectl-build-deploy/lagoon/services-routes`, which rightly excludes the skipped routes (applied [here](https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L1365))
and the second is targeted at `/kubectl-build-deploy/lagoon/deploymentconfigs-pvcs-cronjobs-backups` (applied [here](https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L1684))

However, this second folder runs a straight `helm template` on the service, which creates the service and ingress anyway (one of the three calls in [this](https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/scripts/exec-kubectl-resources-with-images.sh#L19) file

My general feeling is that this second apply shouldn't be reapplying the ingress and services from earlier anyway.

tobybellwood:
ok - so I've fixed that up in my branch, and written tests - will discuss with @shreddedbacon Monday.

Basically, I've purged the service, ingress and DBaaS charts from the template after they've been applied to ensure that they don't get applied twice.

Also fixed up the issue from #2846 whilst I was at it

https://github.com/shreddedbacon/lagoon/compare/service-autoroutes...tobybellwood:service-autoroutes

shreddedbacon:
Wow, there were more changes than I expected required, we reference something way more than I imagined we did

shreddedbacon:
Closing because this is addressed in the `build-deploy-tool`

PR #2866, opened 2021-10-14
More verbose logging in ssh service

shreddedbacon:
# Checklist

Adds more verbose logging to the SSH service.
Also adds a UUID to incoming requests in logs and errors to help with tracking.


tobybellwood:
pending question - not for this PR though - should some of this log info be exposed to the user to keep them informed?

Schnitzel:
> pending question - not for this PR though - should some of this log info be exposed to the user to keep them informed?

I don't think we can, as it will break tooling like rsync which leverages an ssh connection and will be confused if you send text via the shell connection.

PR #2867, opened 2021-10-15
add X-Robots-Tag noindex, nofollow server-snippet to all ingresses

### LINKED ISSUE #2865: Replace lua scripting for x-robots header with annotation on ingress objects
We could get rid of https://github.com/uselagoon/lagoon-images/blob/main/images/nginx/helpers/100_x-robots-header-development.conf by bringing this logic into the build and adding this annotation to the ingress object

```
nginx.ingress.kubernetes.io/server-snippet: |-
       add_header X-Robots-Tag "noindex, nofollow";
```

tobybellwood:
This PR adds a default annotation to all ingresses to prevent robots from crawling `development` sites.

```
    {{- if eq .Values.environmentType "development"}}
    nginx.ingress.kubernetes.io/server-snippet: |-
       add_header X-Robots-Tag "noindex, nofollow";
    {{- end }}
```  

A couple of important notes:
- This only applies to development sites
- This effectively duplicates the ingress annotation already provided in the nginx images - which may eventually be deprecated.  (AFAIK, duplicate X-Robots-Tags are allowed, as long as they don't contradict)
- This extends the no-robots directive to all ingresses (not just nginx-backed ones)
- Documentation has been added to warn of setting additional `server-snippet` annotations in .lagoon.yml, as they will take precedence over the ingress-defined ones, which may inadvertently remove the protection.

We should probably flesh out the annotations tests a bit more to cover this (and other eventualities)

Closes #2865

### DIFF
--- a/docs/using-lagoon-the-basics/lagoon-yml.md
+++ b/docs/using-lagoon-the-basics/lagoon-yml.md
@@ -293,6 +293,10 @@ You can of course also redirect to any other URL not hosted on Lagoon, this will
 
 #### Trusted Reverse Proxies
 
+{% hint style="warning" %}
+Kubernetes will only process a single `nginx.ingress.kubernetes.io/server-snippet` annotation. Please ensure that if you use this annotation on a non-production environment route that you also include the `add_header X-Robots-Tag "noindex, nofollow";` annotation as part of your server-snippet. This is needed to stop robots from crawling development environments as the default server-snippet set to prevent this in development environments in the ingress templates will get overwritten with any `server-snippets` set in .lagoon.yml.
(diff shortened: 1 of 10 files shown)


twardnw:
This really also needs to address autogenerated routes on production environments too, a customer on Lagoon highlighted to me that content on autogenerated routes was being crawled/indexed because of the issue with the current lua script in the nginx images

tobybellwood:
ah - dang, that's an easy fix then!

I'll leave the development wrapper around the custom-ingress, and remove it for all the templated ones.

twardnw (approved):

Issue #1403, opened 2019-11-14
Migrated customer owners don't have access to project groups


rocketeerbkw:
When we migrated to RBAC, we created groups for each customer and project. We made the correct customer > group and project > group assignments but we didn't link the customer group and project group in any way.

This means that a user who was only a member of a `customer` has access to all the projects, but the cannot see or edit the project groups.

rocketeerbkw:
This is been open long enough that trying to fix it now would likely cause more problems than it solves. Anyone who had this issue was able to contact support and have their access restored already.

PR #2871, opened 2021-10-17
Cleanup pre-rbac data

### LINKED ISSUE #1442: Remove left over api-db tables from the RBAC migration
Now that it's been awhile, it's probably safe to remove the tables in `api-db` that are no longer being used after the RBAC migration:

- `customer`
- `customer_user`
- `user`
- `project_user`
- `user`

rocketeerbkw:
# Checklist

 * Remove api-db schemas
 * Remove migrations that would fail without the schemas
 * Deprecate api fields

# Closing issues

closes #1442


tobybellwood:
Noting that for https://github.com/uselagoon/lagoon/issues/1403 it may be required for users to need to be re-added to groups.

tobybellwood:
the error here is when it tries to get a token via ssh, keycloak bounces the request
```
lagoon lagoon-core-keycloak-bc99cf995-j69hf jenkinslagoonpr28713-control-plane keycloak 2021-11-30T19:12:24.307475524Z [33m19:12:24,307 WARN  [org.keycloak.events] (default task-1) type=TOKEN_EXCHANGE_ERROR, realmId=9c427b7b-e9dc-4e1b-a446-32372cf1b7f4, clientId=auth-server, userId=null, ipAddress=10.244.0.29, error=not_allowed, reason='requested_subject not found', auth_method=token_exchange, grant_type=urn:ietf:params:oauth:grant-type:token-exchange, requested_subject=null, client_auth_method=client-secret
```

PR #2872, opened 2021-10-17
Clean up any deploytarget configurations when removing projects

shreddedbacon:
This PR implements logic to ensure that any deployTargets are cleared up when a project is removed.  Less likely to be an issue in production, but in testing or local this resulted in projects incorrectly assuming that they had deploytargets configured.

### DIFF
--- a/services/api/src/resources/project/resolvers.ts
+++ b/services/api/src/resources/project/resolvers.ts
@@ -533,6 +533,15 @@ export const deleteProject: ResolverFn = async (
     );
   }
 
+  // clean up deploytarget configurations for this project
+  try {
+    await query(sqlClientPool, 'DELETE FROM deploy_target_config WHERE project = :pid', {
+      pid
+    });
+  } catch (err) {
+     // Not allowed to stop execution.
+  }
+
   // @TODO discuss if we want to delete projects in harbor or not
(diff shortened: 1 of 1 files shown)


rocketeerbkw:
At the time of merging, the resolver was using a stored procedure to delete the project and related data (`CALL DeleteProject(:name)`), so this should have been handled in there instead of the resolver.

The reasons are 1) Since `deploy_target_config` table has a `REFERENCES` to project, it theoretically needs to be deleted before the project 2) To follow convention.

Fortunately our foreign keys aren't being enforced at the moment!

shreddedbacon:
> At the time of merging, the resolver was using a stored procedure to delete the project and related data (`CALL DeleteProject(:name)`), so this should have been handled in there instead of the resolver.
> 
> The reasons are 1) Since `deploy_target_config` table has a `REFERENCES` to project, it theoretically needs to be deleted before the project 2) To follow convention.
> 
> Fortunately our foreign keys aren't being enforced at the moment!

We can move this into the right place easily enough. I see that the entire DeleteProject procedure doesn't even exist anymore since #2869 was merged, so won't be hard to put it where it needs to go.

tobybellwood:
sorry for crossed PRs there, nice pickup

PR #2881, opened 2021-10-20
Additional SSH service changes

shreddedbacon:
# Checklist

Try to reduce the number of API calls that are made to the targeted cluster.
Prefer Kubernetes targets first with the `lagoon.sh/service` labels, as Openshift 3.11 clusters are nearly non-existent.
Don't even try Openshift if a Kuberenetes is targeted.

### DIFF
--- a/services/ssh/Dockerfile
+++ b/services/ssh/Dockerfile
@@ -27,7 +27,8 @@ ENV TMPDIR=/tmp \
 ENV LAGOON=ssh \
     OC_VERSION=v3.11.0 \
     OC_HASH=0cbc58b \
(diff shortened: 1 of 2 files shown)


Schnitzel:
@shreddedbacon 
nice work! one thing: do we want to switch to kubectl with a newer version? I think this might also have some improvements on error handling, performance, etc.

shreddedbacon:
> @shreddedbacon nice work! one thing: do we want to switch to kubectl with a newer version? I think this might also have some improvements on error handling, performance, etc.

Yeah, I'm looking at that now

shreddedbacon:
I've added a specific branch for v2.0.0 of lagoon where the environment defined openshift reference doesn't exist [here](https://github.com/shreddedbacon/lagoon/tree/ssh-service-fixes-v2.0.0)

PR #2903, opened 2021-11-03
Schedule cronjobs on spot instances by default

smlx:
# Checklist

Use the standard `lagoon.sh/spot` label and taint to try to get cronjobs to execute on spot instances by default.

To be very conservative I thought to only target cronjobs initially. Potentially in future entire services in e.g. development environments could prefer spots. Setting the standard chart `.Values.affinity` and `.Values.tolerations` would allow this and would override the "fallback" spot scheduling for cronjobs in this PR.

This is a draft to gather feedback on the approach. I've only done one template so far, but the idea would be to apply the same change to all standard Lagoon cronjob templates. I also haven't tested that this actually works in a cluster yet :smile: 

# Closing issues

n/a


tobybellwood:
Now superseded and handled in uselagoon/build-deploy-tool

PR #2906, opened 2021-11-03
Feature/support drush gt 8

### LINKED ISSUE #1786: sql-sync fails with Drush 9

bomoko:
As described in #1787, the Lagoon UI sync tasks fail for newer versions of Drush because of the alias handling.

This builds on @mglaman's solution and includes the file sync task, which suffers from the same issue.

I've changed the logic slightly to only prefix the site alias when the `lagoon:alias` command is explicitly supported by Drush, since this is as good an indication as I can think of that the Drupal Integration modules is available.

This is only a short term fix - we will be deprecating these tasks and their corresponding API calls in the medium term.

# Checklist

# Closing issues

Closes #1786


tobybellwood (approved):
tested ok:
for projects with drupal_integrations, it works as expected (#2911 excepted)
for projects without drupal_integrations it relies on coded aliases

PR #2909, opened 2021-11-03
Environment check fixes

### LINKED ISSUE #2904: Deployment of branches not configured for deployment on project
# Describe the bug
Few customers reported that branches not configured for deployment (not present in the `branches` regex string) are deployed anyway

## Expected behavior

honor the `branches` string regex for deployments

### LINKED ISSUE #2905: Automated Creation of Restore Links Broken After v2.2.2
# Describe the bug

The Lagoon API fails to create a RabbitMQ message for a Restore Job, failing with the following message:

```
[error]: lagoon-logs: Send to lagoon-logs: Restore not initiated, reason: TypeError: Cannot read property 'envVariables' of undefined
```

## To Reproduce

1. Initiate a baas restore for any project via the Lagoon UI
2. See that the restore operation either fails immediately or runs forever
3. Find the referenced error message in the API pod logs

## Expected behavior

Lagoon should be able to automatically create restore links via the API, RabbitMQ queues, and KBD operator.

shreddedbacon:
# Checklist

Add an `includedDeleted` field to `environmentByName` query that if set to false, deleted environments will be excluded from the returned data. 
Since the default behaviour is to include deleted environments, adding this as an optional field that defaults to true.

Part of #2829 a new query was introduced which didn't return envvariables that were needed by the restore logic to check which baas bucket/endpoints etc are used. This adds that functionality back.

# Closing issues

closes #2904 
closes #2905

### DIFF
--- a/node-packages/commons/src/api.ts
+++ b/node-packages/commons/src/api.ts
@@ -855,11 +855,11 @@ export async function getActiveSystemForProject(
 export async function getEnvironmentByName(
   name: string,
   projectId: number,
-  excludeDeleted: boolean = false
+  includeDeleted: boolean = true
 ): Promise<any> {
   const result = await graphqlapi.query(`
     {
-      environmentByName(name: "${name}", project:${projectId}, excludeDeleted:${excludeDeleted}) {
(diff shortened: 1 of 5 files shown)


cdchris12:
LG2M

Issue #2848, opened 2021-09-27
Inconsistencies and double-ups in the kubectl apply process


tobybellwood:
following the build-deploy-docker-compose logic for a service deployment, I've realised where we can make some improvements

Given a project with a full slate of compose_services (ingress, service, dbaas, persistent storage etc..)

The service is pulled out and used to template a compose_service-specific yaml file `$YAML_FOLDER/${SERVICE_NAME}.yaml` here (https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L626)

If that service also has an ingress, then that is used to template the same yaml file `$YAML_FOLDER/${SERVICE_NAME}.yaml`, but overwrite it
here (https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L638)

At this point, that file will contain EITHER a service or, if one exists, an Ingress, but not both.

DBaaS services template the same file here (https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L647), but as a compose-service can't have ingress/service and DBaaS, this isn't exclusionary.

Routes are then created individually in the same folder by domain `$YAML_FOLDER/${ROUTE_DOMAIN}.yaml` and BaaS ($YAML_FOLDER/k8up-lagoon-backup-schedule.yaml)

At this point, kubectl apply is then performed on all the files created in this `$YAML_FOLDER/` folder here (https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L1365).

As a result of this kubectl apply, all the DBaaS, BaaS, Route resources are created, and whatever existed in the service&ingress file (note not both, because of the overwrite)

We create another folder at https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L1568 so that we can perform the cronjob, deployment steps, setting persistent volume info, image names etc

When the [exec-kubectl-resources-with-images](https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/scripts/exec-kubectl-resources-with-images.sh) script is called, it performs a greedy `helm template` and processes all resources against the values for `image`.  This creates another set of service and ingress resources in the new $YAML_FOLDER.

[...]

PR #2921, opened 2021-11-08
fix: populate the values.yaml for fastly ingress support

shreddedbacon:
# Checklist

As explained in #2848 ingresses get double applied. Which for adding fastly annotations to autogenerated routes means that the changes are added, but the second apply wipes the changes out.

This adds the `fastly.x` values to the main values.yaml file that is consumed again later on by the secondary apply.


tobybellwood (approved):
LGTM - can confirm that autogen routes now get fastly if needed
```
    lagoon.sh/autogenerated: "true"
...
    fastly.amazee.io/service-id: newlagoon
    fastly.amazee.io/watch: "true"
```

PR #2939, opened 2021-11-22
Passing no-tablespaces flag into drush sql-dump for task

timclifford:
# Checklist

# Closing issues

### DIFF
--- a/services/api/src/resources/task/resolvers.ts
+++ b/services/api/src/resources/task/resolvers.ts
@@ -419,8 +419,9 @@ export const taskDrushSqlDump: ResolverFn = async (
     project: envPerm.project
   });
 
-  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && drush sql-dump --result-file=$file --gzip && \
-TOKEN="$(ssh -p $TASK_SSH_PORT -t lagoon@$TASK_SSH_HOST token)" && curl -sS "$TASK_API_HOST"/graphql \
+  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && DRUSH_MAJOR_VERSION=$(drush status | grep 'Drupal version' | sed 's/[^:]*: *//; s/,.*//' | grep -oE '^\s*[0-9]+') && \
+   if [[ $DRUSH_VERSION -gt 9* ]]; then  drush sql-dump --extra-dump=--no-tablespaces --result-file=$file --gzip; else drush sql-dump --extra=--no-tablespaces --result-file=$file --gzip; fi && \
+  TOKEN="$(ssh -p $TASK_SSH_PORT -t lagoon@$TASK_SSH_HOST token)" && curl -sS "$TASK_API_HOST"/graphql \
 -H "Authorization: Bearer $TOKEN" \
 -F operations='{ "query": "mutation ($task: Int!, $files: [Upload!]!) { uploadFilesForTask(input:{task:$task, files:$files}) { id files { filename } } }", "variables": { "task": '"$TASK_DATA_ID"', "files": [null] } }' \
 -F map='{ "0": ["variables.files.0"] }' \
--- a/services/api/src/resources/task/resolvers.ts
+++ b/services/api/src/resources/task/resolvers.ts
@@ -419,9 +419,9 @@ export const taskDrushSqlDump: ResolverFn = async (
     project: envPerm.project
   });
 
-  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && DRUSH_MAJOR_VERSION=$(drush status | grep 'Drupal version' | sed 's/[^:]*: *//; s/,.*//' | grep -oE '^\s*[0-9]+') && \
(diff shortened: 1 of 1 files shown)


services/api/src/resources/task/resolvers.ts
```diff
@@ -419,7 +419,8 @@ export const taskDrushSqlDump: ResolverFn = async (
     project: envPerm.project
   });
 
-  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && drush sql-dump --result-file=$file --gzip && \
+  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && DRUSH_MAJOR_VERSION=$(drush status --field=drupal-version | grep -oE '^\s*[0-9]+') && \
+if [[ $DRUSH_MAJOR_VERSION -gt 9 ]]; then drush sql-dump --extra-dump=--no-tablespaces --result-file=$file --gzip; else drush sql-dump --extra=--no-tablespaces --result-file=$file --gzip; fi && \
```
tobybellwood: Should this not be `ge` to be 9 or greater instead of greater than 9?

```suggestion
if [[ $DRUSH_MAJOR_VERSION -ge 9 ]]; then drush sql-dump --extra-dump=--no-tablespaces --result-file=$file --gzip; else drush sql-dump --extra=--no-tablespaces --result-file=$file --gzip; fi && \
```

services/api/src/resources/task/resolvers.ts
```diff
@@ -419,7 +419,8 @@ export const taskDrushSqlDump: ResolverFn = async (
     project: envPerm.project
   });
 
-  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && drush sql-dump --result-file=$file --gzip && \
+  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && DRUSH_MAJOR_VERSION=$(drush status --field=drupal-version | grep -oE '^\s*[0-9]+') && \
```
tobybellwood: capture drush version, not drupal version :wink: 
```suggestion
  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && DRUSH_MAJOR_VERSION=$(drush status --field=drush-version | grep -oE '^\s*[0-9]+') && \
```

services/api/src/resources/task/resolvers.ts
timclifford: Ah yeah - that's right, it seems drush 9 introduced `--extra-dump`

drush 10 = https://www.drush.org/latest/commands/sql_dump/
drush 9 = https://drushcommands.com/drush-9x/sql/sql:dump/
drush 8 = https://drushcommands.com/drush-8x/sql/sql-dump/

services/api/src/resources/task/resolvers.ts
timclifford: 😬

tobybellwood:
ok (another comment incoming about the drush version command) but broadly this works on Drush9/10

in Drush 8, the issues in https://github.com/drush-ops/drush/issues/4489 arise, namely that whilst the dump works, it doesn't truncate the cache or watchdog tables, resulting in a potentially enormous backup.

Will continue digging

services/api/src/resources/task/resolvers.ts
```diff
@@ -419,7 +419,8 @@ export const taskDrushSqlDump: ResolverFn = async (
     project: envPerm.project
   });
 
-  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && drush sql-dump --result-file=$file --gzip && \
+  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && DRUSH_MAJOR_VERSION=$(drush status --field=drush-version | grep -oE '^s*[0-9]+') && \
```
tobybellwood: drush 8 has no --field command - this one works with 8/9/10 to the best of my testing|
```suggestion
  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).sql" && DRUSH_MAJOR_VERSION=$(drush status --fields=drush-version | awk '{ print $4 }' | grep -oE '^s*[0-9]+') && \
```

tobybellwood:
so, to replicate the truncating of a "common" structure-tables-list
`$command_specific['sql-dump'] = array('structure-tables-list' => 'cache,cache_*,history,sessions,watchdog,feeds_log');`

```
drush sql-dump --extra=--no-tablespaces | grep -Ev 'INSERT INTO `(cache|cache_|history|sessions|watchdog|feeds_log)' | gzip > $file.gz
```

or similar via awk (fewer matches for brevity)

```
drush sql-dump --extra=--no-tablespaces | awk '!/INSERT INTO `cache_ && !/INSERT INTO `watchdog' | gzip > $file.gz
```

The question is - do we want to define which tables aren't dumped for customers....

tobybellwood:
ok - going to press ahead with this - large dumps are better than broken dumps.

Issue #2177, opened 2020-08-26
Run SSH pod in each lagoon-remote


twardnw:
Currently, all SSH connections run via the SSH pod in the same location that all other Lagoon services run. While this simplifies the auto-generation of Drush aliases, the experience of customers using SSH from locations distant from the main cluster is impacted by this.

Each cluster should have it's own SSH pod running, and the auto-generated drush aliases should prefer this endpoint for any SSH interaction.

tobybellwood:
There's discussion over in #1820 about this

nicksantamaria:
Adding my vote for a solution along these lines. Our team is often frustrated by how long it takes to establish a ssh connection, and the latency involved with the Australia -> Switzerland -> Australia round trip.

tobybellwood:
Been implemented, awaiting wider rollout

Issue #1533, opened 2020-01-07
Keycloak group_attribute table value field limited to 255 characters


shreddedbacon:
**Describe the bug**
Keycloak group_attribute table has a limit of 255 characters, which is used for the `lagoon-projects` attribute. When adding projects to a group that has already full value, it will stop accepting the new value when a new project is added causing a permissions failure for any users that try to access that project that are in the group

```
describe GROUP_ATTRIBUTE;
+----------+--------------+------+-----+-----------------------------+-------+
| Field    | Type         | Null | Key | Default                     | Extra |
+----------+--------------+------+-----+-----------------------------+-------+
| ID       | varchar(36)  | NO   | PRI | sybase-needs-something-here |       |
| NAME     | varchar(255) | NO   |     | NULL                        |       |
| VALUE    | varchar(255) | YES  |     | NULL                        |       |
| GROUP_ID | varchar(36)  | NO   | MUL | NULL                        |       |
+----------+--------------+------+-----+-----------------------------+-------+
```

The error in keycloak is the following when trying to add project `9678` for example (258chars total value)
```
Caused by: java.sql.SQLException: Data too long for column 'VALUE' at row 1
--
  | Query is: insert into GROUP_ATTRIBUTE (GROUP_ID, NAME, VALUE, ID) values (?, ?, ?, ?), parameters ['21ab7da7-4dc7-4745-92ef-a9faf663b8a4','lagoon-projects','244,269,276,323,388,490,539,540,558,591,592,635,654,667,668,675,1213,1215,1217,1219,1220,1222,1223,1224,1225,1226,1227,1228,1230,1231,1232,3717,3718,4023,4024,4053,4054,4055,4056,4057,4059,4060,4061,4063,4065,4066,4214,4215,4216,4217,4218,4219,4220,6641,9678'
```

smlx:
https://github.com/keycloak/keycloak/discussions/8593 has some more information, but doesn't look like Keycloak upstream is planning on increasing this limit.

shreddedbacon:
The way to fix this is `ALTER TABLE GROUP_ATTRIBUTE MODIFY value TEXT;`

We should add this to the keycloak start script probably.

PR #2941, opened 2021-11-23
Add support for ssh-portal / service-api

### LINKED ISSUE #1820: Rewrite the SSH service image to decouple from api-db
**Describe the bug**
SSH service image uses Ubuntu as a base. For consistency, easier maintenance etc. it might be a good idea to switch to alpine.

**To Reproduce**
n/a

**Expected behavior**
SSH image no longer relies on Ubuntu.

**Screenshots**
n/a

**Additional context**
https://github.com/amazeeio/lagoon/issues/1818 is related

### LINKED ISSUE #1875: SSH multiplexing is incompatible with Lagoon SSH
We should just note that SSH multiplexing (`ControlMaster` in `ssh_config`) is not compatible with Lagoon SSH access. Users must disable multiplexing to get SSH working.

### LINKED ISSUE #1876: Lagoon SSH Host Key changes
**Describe the bug**
SSH HostKey warning when connecting to Lagoon via SSH.

**To Reproduce**
1. Connect to Lagoon via SSH.
2. Wait a week.
3. Connect to Lagoon via SSH.

**Expected behavior**
The HostKey should not change.

**Screenshots**
Screenshot from 2020-05-13 17-56-50

**Additional context**
n/a

### LINKED ISSUE #2181: Feature request: new API endpoint to check for user SSH permissions on a namespace
It would be helpful to have a new API endpoint to check if a user has permission to SSH to an environment. The input to the endpoint would be:

* user SSH key
* openshiftProjectName / namespace

The way I imagined the logic might be able to work in the API:

* check if the caller has permission to check SSH key permissions (via a "ssh-check" role or similar)
* look up the user by ssh key
* get the user token from keycloak
* check with keycloak if the user has permission to ssh to that environment
* return true/false. Or maybe the environment object / null. As long as there's a way to interpret the response as a boolean.

I believe the concept of permission-check APIs is new in the current Lagoon API, but there's at least one other use-case that has come up: https://github.com/amazeeio/lagoon/issues/1642

smlx:
# Checklist

This PR makes several changes in support of the `ssh-portal` service which is being added to the `lagoon-remote` chart.
The motivations for adding the `ssh-portal` service include:

* Improve security posture by no longer requiring administrative access to the "remote" cluster k8s API.
* Improve latency and performance for SSH connections.

See the individual commits for details, but the main changes are:

* Add a `service-api` client to keycloak.
* Add a new structured annotation to keycloak groups containing group -> project ID mappings.
* Add tests for the `ssh-portal`, based on the existing `remote-shell` tests.

The associated lagoon-charts PR is here: https://github.com/uselagoon/lagoon-charts/pull/382
The ssh-portal / service-api repository is here: https://github.com/uselagoon/lagoon-ssh-portal

# Closing issues

Closes: #1820
Closes: #1875
Closes: #1876 (for the ssh-portal host)
Closes: #2181
Partially addresses: #2177

# Performance improvements

The user SSH authorization check is approximately 90% faster using `ssh-portal` instead of the current `ssh` service, as well as having much lower latency for interactive shells due to not having to bounce through `lagoon-core`.
Below is a very rough "benchmark" in a Lagoon running locally in `kind`.

Current SSH service:
```
$ time ssh -i ./local-dev/cli_id_ed25519 -p 31320 ci-features-api-variables-control-k8s-lagoon-api-variables@172.18.0.2 date
Tue Feb 15 05:56:58 UTC 2022

real	0m1.388s
user	0m0.027s
sys	0m0.008s
```

ssh-portal:
```
$ time ssh -i ./local-dev/cli_id_ed25519 -p 30332 ci-features-api-variables-control-k8s-lagoon-api-variables@172.18.0.2 date
Tue Feb 15 05:56:53 UTC 2022

real	0m0.143s
user	0m0.017s
sys	0m0.005s
```

# Rollout considerations

For an existing installation of `lagoon-core`, the Keycloak groups will not have the annotations required to allow `ssh-portal` to work. Therefore there is a requirement to add functionality to the Lagoon API to update all the Keycloak group annotations. This is not included in this PR.


smlx:
I'd really appreciate review of the Keycloak `start.sh` changes in case I've somehow broken something.

services/api/src/models/group.ts
```diff
@@ -614,7 +614,8 @@ export const Group = (clients: {
         {
           attributes: {
             ...group.attributes,
-            'lagoon-projects': [newGroupProjects]
+            'lagoon-projects': [newGroupProjects],
+            'group-lagoon-project-ids': [`{${JSON.stringify(group.name)}:[${newGroupProjects}]}`]
```
rocketeerbkw, line 618: This will also suffer from https://github.com/uselagoon/lagoon/issues/1533. Just FYI, nothing that should be changed.

tests/tests/features/remote-ssh-portal.yaml
```diff
@@ -0,0 +1,75 @@
...
+  tasks:
+  - include: ../../tasks/git-init.yaml
+  - include: ../../tasks/git-add-commit-push.yaml
+
+- name: "{{ testname }} - api deployEnvironmentBranch on {{ project }}, which should deploy the first commit"
+  hosts: localhost
+  serial: 1
+  vars:
+    branch: "{{ branch }}"
+    project: "{{ project }}"
```
tobybellwood, line 15:
```suggestion
    branch: "{{ branch }}"
    project: "{{ project }}"
    namespace: "{{ project | regex_replace('_', '-') }}-{{ branch | regex_replace('/', '-') }}"
```

tobybellwood: In https://github.com/uselagoon/lagoon/pull/3039 we need to pass the {{ namespace }} to the deploy tasks, hence this change

services/keycloak/start.sh
```diff
@@ -108,17 +108,11 @@ function configure_api_client {
...
-    echo '{"type":"client","logic":"POSITIVE","decisionStrategy":"UNANIMOUS","name":"Client auth-server Policy","clients":["'$AUTH_SERVER_CLIENT_ID'"]}' | /opt/jboss/keycloak/bin/kcadm.sh create clients/$REALM_MANAGEMENT_CLIENT_ID/authz/resource-server/policy/client --config $CONFIG_PATH -r lagoon -f -
-    AUTH_SERVER_CLIENT_POLICY_ID=$(/opt/jboss/keycloak/bin/kcadm.sh get -r lagoon clients/$REALM_MANAGEMENT_CLIENT_ID/authz/resource-server/policy?name=Client+auth-server+Policy --config $CONFIG_PATH | python -c 'import sys, json; print json.load(sys.stdin)[0]["id"]')
-    # 4 Update user impersonate permission to add client policy (PUT)
-    IMPERSONATE_PERMISSION_ID=$(/opt/jboss/keycloak/bin/kcadm.sh get -r lagoon clients/$REALM_MANAGEMENT_CLIENT_ID/authz/resource-server/permission?name=admin-impersonating.permission.users --config $CONFIG_PATH | python -c 'import sys, json; print json.load(sys.stdin)[0]["id"]')
-    /opt/jboss/keycloak/bin/kcadm.sh update clients/$REALM_MANAGEMENT_CLIENT_ID/authz/resource-server/permission/scope/$IMPERSONATE_PERMISSION_ID --config $CONFIG_PATH -r ${KEYCLOAK_REALM:-master} -s 'policies=["'$AUTH_SERVER_CLIENT_POLICY_ID'"]'
```
rocketeerbkw: This file is designed to be run as an init/update style (each function run once, and in order), not as a well organized set "real" functions. While it makes sense to move similar code together (`configure_token_exchange`), it breaks the logic. Each function, once released, should be considered "final" and not changed (misspellings aside).

This change, for example, I think directly led to more issues related to [#3402](https://github.com/uselagoon/lagoon/pull/3042). The [api waits for the existence of the api-client](https://github.com/uselagoon/lagoon/blob/main/services/api/src/util/waitForKeycloak.ts#L16-L21) before advertising itself as ready, but by moving the token exchange config to the end, the api thinks it's ready to answer requests when it's not ready yet (eg, getting a token via ssh by querying the api then asking auth-server to exchange).

services/keycloak/start.sh
```diff
@@ -1673,6 +1667,56 @@ function remove_billing_modifier {
...
+    echo Creating realm roles mapper for service-api
+    echo '{"protocol":"openid-connect","config":{"multivalued":"true","id.token.claim":"false","access.token.claim":"true","userinfo.token.claim":"false","claim.name":"realm_roles"},"name":"User Realm Roles","protocolMapper":"oidc-usermodel-realm-role-mapper"}' | /opt/jboss/keycloak/bin/kcadm.sh create -r ${KEYCLOAK_REALM:-master} clients/$CLIENT_ID/protocol-mappers/models --config $CONFIG_PATH -f -
+    echo Creating group lagoon project IDs mapper for service-api
+    echo '{"protocol":"openid-connect","config":{"id.token.claim":"false","access.token.claim":"true","userinfo.token.claim":"false","multivalued":"true","aggregate.attrs":"true","user.attribute":"group-lagoon-project-ids","claim.name":"group_lagoon_project_ids","jsonType.label":"String"},"name":"Group Lagoon Project IDs","protocolMapper":"oidc-usermodel-attribute-mapper"}' | /opt/jboss/keycloak/bin/kcadm.sh create -r ${KEYCLOAK_REALM:-master} clients/$CLIENT_ID/protocol-mappers/models --config $CONFIG_PATH -f -
+}
+
+function configure_token_exchange {
```
rocketeerbkw, line 1717: As mention in a previous comment, you can move this logic related to `service-api` to also be in `configure_service_api_client`.

rocketeerbkw (changes requested):

services/api/src/models/group.ts
smlx: Ugh, painful. Thanks for letting me know.

I guess you'd need to get to ~100 projects to get near this limit though?

tests/tests/features/remote-ssh-portal.yaml
smlx: Thanks, I applied this change

services/keycloak/start.sh
smlx: Ah okay thanks for explaining the existing structure of this file.

I did try to keep the changes as small as possible, but the problem I ran into was [here](https://github.com/uselagoon/lagoon/pull/2941/files#diff-8a14c757e38a4f4906e6dd229db8406de31fec05d5b8c5e1c6b0514e5a8c1f19R1717) where I need to pass the IDs of both the clients which can do token exchange to the policy. In the one-function-per-client model I don't think I can do this because I don't have the IDs of both clients.

Do you have any suggestions for a better way to structure the file that allows multiple clients to be added to the token exchange policy?

services/keycloak/start.sh
smlx: As mentioned above I did initially try to do it this way, but the problem I ran into is that whichever function ran last set the token-exchange policy on that client only, and removed the token exchange policy from the other client :disappointed: 

Can you see a better way to do the same thing with less intrusive changes to the api client config?

rocketeerbkw: I haven't tested but would it also work if you just keep the `configure_token_exchange` logic but don't modify `configure_api_client`? That's basically saying 1) set token exchange for `auth-server` only 2) update token exchange for both `auth-server` and `service-api`. The `api` will know it's ready after `1` and when the policy is updated later it doesn't break anything? You can still grab the IDs of both clients, without setting the policy for both only the one time.

And if you move the logic of `configure_token_exchange` into `configure_service_api_client` you don't have to "check if policies already exist" (line 1696), because you'll know that `configure_api_client` was just run or was run previously, and you can assume if the `service-api` doesn't exist (you're already checking that in `configure_service_api_client`) you need to update the token exchange for both services.

services/keycloak/start.sh
rocketeerbkw: answered in the other thread

services/keycloak/start.sh
smlx: I refactored as you suggested, and CI is passing so :crossed_fingers: it didn't break anything.

Could you take another look when you get a chance?

smlx:
CI is passing except for the documentation link checks. Ready for review again.

services/keycloak/start.sh
```diff
@@ -1691,6 +1691,50 @@ function update_openshift_view_permission {
...
+    fi
+
+    # create client service-api policy
+    SERVICE_API_CLIENT_ID=$(/opt/jboss/keycloak/bin/kcadm.sh get  -r lagoon clients?clientId=service-api --config $CONFIG_PATH | python -c 'import sys, json; print json.load(sys.stdin)[0]["id"]')
+    echo '{"type":"client","logic":"POSITIVE","decisionStrategy":"UNANIMOUS","name":"Client service-api Policy","clients":["'$SERVICE_API_CLIENT_ID'"]}' | /opt/jboss/keycloak/bin/kcadm.sh create clients/$REALM_MANAGEMENT_CLIENT_ID/authz/resource-server/policy/client --config $CONFIG_PATH -r lagoon -f -
+    SERVICE_API_CLIENT_POLICY_ID=$(/opt/jboss/keycloak/bin/kcadm.sh get -r lagoon clients/$REALM_MANAGEMENT_CLIENT_ID/authz/resource-server/policy?name=Client+service-api+Policy --config $CONFIG_PATH | python -c 'import sys, json; print json.load(sys.stdin)[0]["id"]')
+
+    echo Enable token exchange for auth-server and service-api
+    # The decision strategy is affirmative since only one policy has to pass.
+    # NOTE: this assumes that configure_api_client has set the global AUTH_SERVER_CLIENT_POLICY_ID variable.
```
rocketeerbkw: I'm afraid that assumption will cause problems. There are 3 scenarios each function has to account for:

1. A fresh install ✅: `configure_api_client` will see that it needs to get run and set `AUTH_SERVER_CLIENT_POLICY_ID` needed in this function
2. An already fully configured keycloak ✅: `configure_api_client` [will not run](https://github.com/uselagoon/lagoon/blob/main/services/keycloak/start.sh#L100) and will not set `AUTH_SERVER_CLIENT_POLICY_ID` but `configure_token_exchange` [also won't run](https://github.com/uselagoon/lagoon/pull/2941/files#diff-8a14c757e38a4f4906e6dd229db8406de31fec05d5b8c5e1c6b0514e5a8c1f19R1723) so there will be no issues
3. An upgrade from lagoon <= 2.4.1 ❌: `configure_api_client` [will not run](https://github.com/uselagoon/lagoon/blob/main/services/keycloak/start.sh#L100) and will not set `AUTH_SERVER_CLIENT_POLICY_ID` but `configure_token_exchange` will see that it needs to create the client policy and will probably set the impersonate permission so that it has a null/empty value. Might throw an error? Might fail silently?

You can just copy/paste the `AUTH_SERVER_CLIENT_POLICY_ID` querying code in here so that this function has no dependencies.

rocketeerbkw (changes requested):

services/keycloak/start.sh
smlx: Oh nicely spotted, thanks!

I'll fix this up as you suggest.

smlx: Okay, made the change. Does that look better?

rocketeerbkw (approved):

services/keycloak/start.sh
rocketeerbkw: LGTM!

PR #2946, opened 2021-11-26
Feature: remove associated variables from api when environments and projects are deleted

shreddedbacon:
# Checklist

When environments and projects are deleted, the environment variables remain in the api-db.
These no longer need to exist if the associated project or environment are removed.


services/api/src/resources/environment/helpers.ts
```diff
@@ -27,6 +27,19 @@ export const Helpers = (sqlClientPool: Pool) => {
   return {
     aliasOpenshiftToK8s,
     getEnvironmentById,
+    deleteEnvironment: async (name: string, pid: number) => {
+      // clean up environment variables
+      await query(
+        sqlClientPool,
+        'DELETE FROM `env_vars` WHERE `environment` = :pid',
```
rocketeerbkw: I don't think `environment = pid` is correct? You'll need to do an extra sql query to load the environment id, or use a `join` to the `environment` table, or change the function parameter to accept environment id instead of name, since in the resolver there is already a call to [load the full environment](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/environment/resolvers.ts#L460)

rocketeerbkw (changes requested):
Do we also want to add a query to migrations that deletes existing left over env vars?

services/api/src/resources/environment/helpers.ts
shreddedbacon: Ahh you are probably right, I'll have another look when I get a chance.

shreddedbacon:
> Do we also want to add a query to migrations that deletes existing left over env vars?

I'm thinking that is probably a good idea

services/api/src/resources/environment/helpers.ts
```diff
@@ -27,6 +27,22 @@ export const Helpers = (sqlClientPool: Pool) => {
   return {
     aliasOpenshiftToK8s,
     getEnvironmentById,
+    deleteEnvironment: async (name: string, eid: number, pid: number) => {
+      // clean up environment variables
+      await query(
+        sqlClientPool,
+        'DELETE FROM `env_vars` WHERE `environment` = :eid',
+        { eid }
+      );
```
rocketeerbkw: This looks like the same query as below. Forgot to delete after a refactor perhaps?

services/api/src/resources/project/sql.ts
```diff
@@ -29,6 +29,26 @@ export const Sql = {
     knex('project as p')
       .whereIn('p.id', projectIds)
       .toString(),
+  deleteEnvironmentVariables: (id: number) =>
+    knex('env_vars')
+      .where('project', '=', id)
+      .delete()
+      .toString(),
+  deleteDeployTargeTConfigs: (id: number) =>
```
rocketeerbkw: nitpick: weird capitalization `deleteDeployTargeTConfigs` > `deleteDeployTargetConfigs`. Also needs fixed in services/api/src/resources/project/helpers.ts

rocketeerbkw (changes requested):
Looks like it would work fine, I have a few small change requests too.

rocketeerbkw (approved):
There's still the question of adding a "delete currently orphaned env variables" migration but it's fine if that ends up as a follow up PR

shreddedbacon:
I can probably squeeze in the cleanup as part of this still, just need to put some time towards it.

shreddedbacon:
I've added the clean up migration, would be good to get some :eyes: to sanity check

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -1737,6 +1737,16 @@ CREATE OR REPLACE PROCEDURE
   END;
 $$
 
+CREATE OR REPLACE PROCEDURE
+  clean_stale_project_data()
+
+  BEGIN
+    DELETE FROM env_vars WHERE project NOT IN (SELECT id FROM project);
```
tobybellwood, line 1744: do we also need to delete env_vars where the environment no longer exists (but the project still does?)

shreddedbacon: Yes :doh:

shreddedbacon:
Also, not sure if need to check with other teams. But if a project is delete, do the associated environments also need to be deleted?? Even though they've been soft deleted, the rows will still exist

Edit: lets do this later, there is more to this part

Issue #2789, opened 2021-07-28
StorageCalc should not run on RWO PVs that are already mounted


seanhamlin:
# Describe the bug

I occasionally see storagecalc pods for PVs that are ReadWriteOnce (RWO). StorageCalc attempts to mount the PV in order to do the `du` but fail to get the claim.

![image](https://user-images.githubusercontent.com/10605846/127395906-84b1dec3-b05b-4c5b-ace8-715b683d6fe3.png)

## To Reproduce

1. Create RWO PV, e.g. Solr
2. Wait for storageCalc to run (daily)
3. See error above

## Expected behavior

Just skip over RWO PVs that are mounted.

vincenzodnp:
I saw the same issue and I was prepping a PR by setting an `interpod-affinity` rule and be sure `storagecalc` pod is scheduled in the same node where `solr` pod is.
The idea is to modify the helmchart and add something like 
```
podAntiAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
```
I'd use `preferredDuringSchedulingIgnoredDuringExecution` instead of `requiredDuringSchedulingIgnoredDuringExecution` because `storagecalc` must run even if  any Solr pod scheduled 

@seanhamlin

tobybellwood:
new storagecalc fixed this

Issue #2245, opened 2020-10-14
Lagoon API lists `bytesUsed` when it is actually `kiloBytesUsed`


seanhamlin:
# Describe the bug

The storage calculator incorrectly (and confusingly) captures KB instead of Bytes. It looks like this was introduced in https://github.com/amazeeio/lagoon/commit/e59e3a05985de4ac4bacaf3a508904204a98ab2f but never reflected in the API.

## To Reproduce

Install the latest version of `du`, on Mac, this is GNU `gdu`. The command used to count the storage space is [listed in the code](https://github.com/amazeeio/lagoon/blob/main/services/storage-calculator/calculate-storage.sh#L122).

```
du -s /storage/${PVC} | cut -f1
```

Reading up on the `-s` switch

>        -s, --summarize
>             display only a total for each argument

Here is an example from my local laptop of a large directory of log files:

1. Current command

```
$ gdu -s ~/largedir/ | cut -f1
109611500
```

2. Forcing Bytes

```
$ gdu -b ~/largedir/ | cut -f1
111982025079
```

3. Forcing KB

```
$ gdu -sk ~/largedir/ | cut -f1
109611500
```

4. Human readable

```
$ gdu -h ~/largedir/ | cut -f1
105G
```

## Expected behavior

One of 2 options:

1. That the `bytesUsed` matches up to actual bytes used. 
2. The API is changed to `kiloBytesUsed` and the `-k` flag added for clarity

Either way, this is likely a breaking API change, and possibly better suited to Lagoon 2.0.

shreddedbacon:
The [storage calculator is being updated](https://github.com/uselagoon/storage-calculator/pull/11) to include the kilobyte field, while retaining the old one for backwards compatability. The data in both fields remains the same though, just simply making it clear that the data is kilobytes not bytes. We will need to update the actions-handler to consume the new field if it is present, and announce a deprecation of the old one (minimum lagoon-remote chart version required to get the new field)

The API is a different story, we can expose the data in a new field that just displays the same data with `kiloBytesUsed` and deprecate `bytesUsed`

PR #2947, opened 2021-11-27
Refactor Storage Calculator to be more robust

seanhamlin:
# Checklist

It was noticed that `storage-calculator` was not running for all projects in a large dedicated cluster with it's own Lagoon Core. This PR refactors the storage calculator to:

* Use consistent code style
* Easier to read debug messages
* Move guard statements to the top to avoid unnecessary nesting of code
* Remove `exit 1` which potentially halts the storage calculator before it runs over all projects
* Use of `--ignore-not-found=true` and less piping to `/dev/null`
* Have a timeout of **30 seconds** for the `storageCalc` pod rollout. To prevent `RWO` PVs from stalling the script for too long. The current default is 10 minutes.
* Add a new environment variable `LAGOON_STORAGE_IGNORE_REGEX` to which can be used to selectively skip certain PVCs by name. e.g. it is easy now to skip reading all the `solr` PVCs now.

I also added README.md file to explain a lot of the things.

Things that would make great follow up PRs:

* Move away from `python-yq` to Go `yq`.
* Fixing #2789

# Related issues

* Related to #2245

### DIFF
--- a/services/storage-calculator/README.md
+++ b/services/storage-calculator/README.md
@@ -1,0 +1,37 @@
+# Storage Calculator
+
+This service is responsible for:
+
+* Recording the used size of any Persistent-Volume (PV)
+* Recording the used size of any database (`data` and `index`)
+
+The result of the size calculations is sent to the Lagoon API.
+
+The storage is measured in `KB`.
+
+If you want to retrieve the storage for a given project, you can use GraphQL:
+
+```graphql
+query {
+  projectByName(name: "PROJECT") {
+    id
+    name
+    gitUrl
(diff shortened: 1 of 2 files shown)


rocketeerbkw:
Adding more details about the issue and reason the changes here will help from internal comms:

- there are a lot of projects (each with 1+ environments)
- projects are enumerated through in id order, meaning new projects are done last
- if there is a RWO PV in use, then the process stalls for 10 minutes to wait for it to fail to spawn
- the cron process physically cannot complete in 24 hours, and thus new projects are just skipped.

PR #2958, opened 2021-12-07
Document additional requirements for collecting logs of the NGINX ingress controller

rocketeerbkw:
# Checklist

The lagoon-logging chart expects a specific namespace that isn't documented anywhere. Even though the nginx ingress controller [example install command](https://kubernetes.github.io/ingress-nginx/deploy/#quick-start) is correct, it's possible for a user to install to a namespace that would not allow collecting the logs

# Closing issues

n/a

### DIFF
--- a/docs/using-lagoon-advanced/installing-lagoon-into-existing-kubernetes-cluster.md
+++ b/docs/using-lagoon-advanced/installing-lagoon-into-existing-kubernetes-cluster.md
@@ -10,7 +10,7 @@ description: >-
 
 * Kubernetes 1.18+
 * Familiarity with [Helm](https://helm.sh) and [Helm Charts](https://helm.sh/docs/topics/charts/#helm), and [kubectl](https://kubernetes.io/docs/tasks/tools/).
-* Ingress controller (ingress-nginx)
+* [Ingress controller (ingress-nginx)](#logging-nginx-ingress-controller)
 * Cert manager (for TLS) - We highly recommend using letsencrypt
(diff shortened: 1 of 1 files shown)


tobybellwood:
It might be worth explicitly saying this in [line 13](https://github.com/uselagoon/lagoon/pull/2958/files#diff-bda755e04e150480ef0082a390e36d2225340833c8bdf1710199e8efd154cbf3R13) too - ingress-nginx installed in the ingress-nginx namespace

rocketeerbkw:
My reading of the docs (which could be wrong) is that you'd also need to add the contents of the gist mentioned in the section I edited. If both are needed, do we mention both at the top of the doc (line 13)? Or we could change that to `Ingress controller (ingress-nginx) <link to below section>configured for lagoon</link>`

tobybellwood:
I think you can add the logging config later, but can't change the namespace - but just making people aware that there is a caveat in case they don't read the whole doc at once

smlx:
> My reading of the docs (which could be wrong) is that you'd also need to add the contents of the gist mentioned in the section I edited.

The `lagoon-logging` chart just expects to parse this as JSON. So the linked gist is an example that can be customised but needs to remain valid JSON.

rocketeerbkw:
I made the top level requirement link to the section that explains how it should be configured. It is possible to fix an incorrect namespace install, which we can document in a new section/PR

tobybellwood:
docs should cover it now, I think?

PR #2973, opened 2021-12-21
fix: resolver for environment openshift needs to do a project lookup

### LINKED ISSUE #2972: Permission issue resolving kubernetes for environments
# Describe the bug

When querying an environments kubernetes resolver, a permission denied is encountered if you are not an admin
```
projectByName(name: $name) {
  id
  name
  environments {
    id
    name
    environmentType
    openshiftProjectName
    kubernetes {
      id
    }
  }
}
```

## Expected behavior

It should resolve the same for querying the kubernetes at the project level.

shreddedbacon:
# Checklist

# Closing issues

closes #2972

### DIFF
--- a/services/api/src/resources/openshift/resolvers.ts
+++ b/services/api/src/resources/openshift/resolvers.ts
@@ -1,6 +1,7 @@
 import * as R from 'ramda';
 import { ResolverFn } from '../';
 import { query, isPatchEmpty } from '../../util/db';
+import { Helpers as projectHelpers } from '../project/helpers';
 import { Sql } from './sql';
 
 const attrFilter = async (hasPermission, entity) => {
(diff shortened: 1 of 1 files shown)


services/api/src/resources/openshift/resolvers.ts
```diff
@@ -114,7 +114,22 @@ export const getOpenshiftByEnvironmentId: ResolverFn = async (
   args,
   { sqlClientPool, hasPermission }
 ) => {
-  await hasPermission('openshift', 'viewAll');
+  // get the project id for the environment
+  const projectrows = await query(
+    sqlClientPool,
+    `SELECT e.project
```
rocketeerbkw: There is a preference to reduce raw sql queries in resolvers. For permissions checks you could  re-use a knex/sql object `selectProjectByEnvironmentId` (first [import](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/environment/resolvers.ts#L11), then use) or avoid the `query` all together by using a project helper like `getProjectByEnvironmentId` (first [import](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/environment/resolvers.ts#L12), then [use](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/environment/resolvers.ts#L244-L246))

rocketeerbkw (changes requested):

services/api/src/resources/openshift/resolvers.ts
shreddedbacon: refactored with the helper

rocketeerbkw (approved):

PR #2989, opened 2022-01-07
Add python-persistent helm chart.

### LINKED ISSUE #2988: Create new Lagoon service type `python-persistent`
Like `python` but with a PV.

steveworley:
# Checklist

## Explain the **details** for making this change. What existing problem does the pull request solve?

Some python applications are currently deployed with the `node-persistent` service type to allow binding a persistent volume to the deployment. This introduces a `python-persistent` service type for semantic separation of deployments. 

The main driver behind this is CKAN deployments as they currently rely on a PV. The `node-persistent` service type uses a TCP liveness probe rather than HTTP. This means that we cannot include Layer7 validations for CKAN to get Kubernetes to restart the pods. 

The python liveness probe is configured for HTTP, so we can add additional listeners and application lookups to determine if a scaling event has moved the PG pod and have kubernetes restart the CKAN pod.

## Closing issues

Closes #2988


tobybellwood:
Thanks for this - it looks pretty good! Only pickup from me is that the cronjob isn't configured to have the volume mounts, so any cronjobs that run as separate containers won't have them. It should be easy enough to copy across from another service (e.g. node-persistent)

steveworley:
Ah! Yeah - forgot to add the PV to the cron pod, will update the PR thanks!

images/kubectl-build-deploy-dind/helmcharts/python-persistent/templates/deployment.yaml
```diff
@@ -0,0 +1,80 @@
...
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "python-persistent.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "python-persistent.labels" . | nindent 8 }}
+      annotations:
+        {{- include "python-persistent.annotations" . | nindent 8 }}
```
tobybellwood, line 19: also, if we add the configMapSha as an annotation, it will redeploy on a configMap update
```suggestion
        {{- include "python-persistent.annotations" . | nindent 8 }}
        lagoon.sh/configMapSha: {{ .Values.configMapSha | quote }}
```

images/kubectl-build-deploy-dind/helmcharts/python-persistent/templates/cronjob.yaml
```diff
@@ -0,0 +1,86 @@
...
+                {{- toYaml $.Values.resources | nindent 16 }}
+              volumeMounts:
+                - mountPath: /var/run/secrets/lagoon/sshkey/
+                  name: lagoon-sshkey
+                  readOnly: true
+                - name: {{ $.Values.persistentStorage.name }}
+                  mountPath: {{ $.Values.persistentStorage.path | quote }}
+          restartPolicy: Never
+          {{- with $.Values.pythonSelector }}
+          pythonSelector:
```
tobybellwood: these two lines should go back to be nodeSelector (as they refer to the k8s node)
```suggestion
          {{- with $.Values.nodeSelector }}
          nodeSelector:
```

tobybellwood:
Tested locally by lablelling existing python docker-compose
```
    labels:
      lagoon.type: python-persistent
      lagoon.persistent: /files
```
Works as expected :tada:

PR #2994, opened 2022-01-17
Prefer to segregate workloads by environment type

smlx:
# Checklist

This change makes kubernetes prefer to schedule pods of a particular environment type together on the same node, and to keep pods of differing environment types on separate nodes.

The idea is that rapidly idled and/or noisy-neighbour development pods are less likely to affect prod pods this way. For example when development environments are idled the cluster autoscaler will remove a node containing only development pods, instead of also disrupting production pods.

Note that this is only a scheduling preference, and does not guarantee that production pods will not be disrupted.

# Closing issues

n/a


smlx:
After some discussion there are a few problems with this approach:
* more chance that HA deployments are placed on the same node
* pod affinity is namespaced, so to have affinity work between pods of different Lagoon projects we would have to use `namespaceSelector: {}` which only appeared in k8s 1.22.

In summary it seems that inter-pod affinity is not a good way to achieve workload segregation.

PR #2997, opened 2022-01-18
insights-handler service

timclifford:
Adds Insights handler service


services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,713 @@
...
+				if key == "lagoon.sh/insightsService" {
+					resource.Service = value
+				}
+				if key == "lagoon.sh/insightsFormat" {
+					resource.Format = value
+				}
+			}
+		}
+
+		if incoming.Payload == nil && incoming.BinaryPayload == nil {
```
bomoko, line 315: If we hit this, do we actually process anything - should it not just kill the message with a log?

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,713 @@
...
+		if err != nil {
+			fmt.Errorf(err.Error())
+		}
+
+		project, environment, apiErr := determineResourceFromLagoonAPI(apiClient, resource)
+		if apiErr != nil {
+			return apiErr
+		}
+		source := fmt.Sprintf("image-inspect:%s", resource.Service)
+		apiErr = h.deleteExistingFactsBySource(apiClient, environment, source, project)
```
bomoko: I'm wondering if this means we're deleting the existing facts before we know whether or not we can successfully unmarshall/process the new ones? Should just be just before we push the new facts?

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,722 @@
...
+						LagoonType: Facts,
+					}
+				}
+				if key == "lagoon.sh/insightsType" && value == "image-inspect-gz" {
+					insights = InsightsData{
+						InputType:  ImageInspectGz,
+						LagoonType: ImageInspectFacts,
+					}
+				}
+				if key == "lagoon.sh/insightsProject" {
```
tobybellwood:
```suggestion
				if key == "lagoon.sh/project" {
```

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,722 @@
...
+					resource.Project = value
+				}
+				if key == "lagoon.sh/insightsEnvironment" {
```
tobybellwood:
```suggestion
				if key == "lagoon.sh/environment" {
```

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,722 @@
...
+					resource.Environment = value
+				}
+				if key == "lagoon.sh/insightsService" {
```
tobybellwood:
```suggestion
				if key == "lagoon.sh/service" {
```

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,722 @@
...
+					resource.Service = value
+				}
+				if key == "lagoon.sh/insightsFormat" {
```
tobybellwood: What's this label for?

timclifford: This was to allow control for different formats of insights data that we might get before we decided on gzipping it all. So if the type of insight data comes in as raw json or csv or whatever we can handle it differently.

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,722 @@
...
+	return res, nil
+}
+
+func processFactsFromSBOM(facts *[]cdx.Component, environmentId int, source string) []lagoonclient.AddFactInput {
+	var factsInput []lagoonclient.AddFactInput
+	if len(*facts) == 0 {
+		return factsInput
+	}
+
+	keyFacts, err := scanKeyFactsFile("./key_facts.txt")
```
tobybellwood, line 697: this file isn't currently in the image - so the handler fails here - we either copy it into the dockerfile, or mount it via a configMap

services/insights-handler/main.go
```diff
@@ -0,0 +1,178 @@
...
+	jwtTokenSigningKey = getEnv("JWT_SECRET", jwtTokenSigningKey)
+	jwtAudience = getEnv("JWT_AUDIENCE", jwtAudience)
+	jwtSubject = getEnv("JWT_SUBJECT", jwtSubject)
+	jwtIssuer = getEnv("JWT_ISSUER", jwtIssuer)
+	insightsQueueName = getEnv("INSIGHTS_QUEUE_NAME", insightsQueueName)
+	insightsExchange = getEnv("INSIGHTS_EXCHANGE", insightsExchange)
+	secretAccessKey = getEnv("S3_FILES_SECRET_ACCESS_KEY", secretAccessKey)
+	s3Origin = getEnv("S3_FILES_HOST", s3Origin)
+	accessKeyID = getEnv("S3_FILES_ACCESS_KEY_ID", accessKeyID)
+	bucket = getEnv("S3_FILES_BUCKET", bucket)
```
tobybellwood: If we're using the same bucket as the task and build logs (which we probably should for now), we need to make sure we're creating a subfolder for sboms - e.g. https://github.com/uselagoon/lagoon/blob/main/services/logs2s3/src/readFromRabbitMQ.ts#L47

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,749 @@
...
+					gz := gzip.NewWriter(&buf)
+					gz.Write(resultJson)
+					gz.Close()
+					err = ioutil.WriteFile(tempFilePath, buf.Bytes(), 0644)
+					if err != nil {
+						fmt.Errorf(err.Error())
+					}
+				}
+
+				s3FilePath := strings.ToLower(fmt.Sprintf("%s/%s/%s/%s/%s", insights.InputType, resource.Project, resource.Environment, resource.Service, objectName))
```
tobybellwood: maybe call the top level folder `insights` and store all insights in the same folder?
```suggestion
				s3FilePath := strings.ToLower(fmt.Sprintf("insights/%s/%s/%s", resource.Project, resource.Environment, objectName))
```

services/insights-handler/internal/handler/main.go
```diff
@@ -0,0 +1,749 @@
...
+		log.Printf("Successfully created %s\n", h.S3Config.Bucket)
+	}
+
+	if incoming.Payload != nil {
+		b, err := json.Marshal(incoming)
+		if err != nil {
+			return err
+		}
+
+		objectName := strings.ToLower(fmt.Sprintf("%s-%s-%s-%s.json", insights.InputType, resource.Project, resource.Environment, resource.Service))
```
tobybellwood: I think this can be changed to `ci-python-control-k8s-python37-python-sbom.json` (and likeewise the gz, but I don't think we need to keep that one)

image

tobybellwood:
As discussed, I recommend splitting this out into a separate repo (at least whilst it's under development) at uselagoon/insights-handler, but still adding it to the lagoon-core chart, just with an imageTag override (:latest) and Always Pull.

tobybellwood:
Let's close this pr and continue in uselagoon/insights-handler

tobybellwood:
Closing in favour of https://github.com/uselagoon/insights-handler/pull/4

PR #3006, opened 2022-01-26
Removes stored proc calls from resolvers

### LINKED ISSUE #3003: Rework the stored procedures in the DB to be in line with existing resolvers
Currently there are a handful of stored procedures used by the API, for instance [here](https://github.com/uselagoon/lagoon/blob/51778508bb7650df5b2ae6cb005c2c5408066451/services/api/src/resources/environment/resolvers.ts#L418).

Until these stored procs are expunged from the DB we're going to have a hard dependency on mysql/mariadb for the API.

The resolvers using these SPs should be reworked to perform the manipulation via the query builder.

bomoko:
Removes references from the api to Stored Procs as described in #3003 

# Checklist

# Closing issues

closes #3003

### DIFF
--- a/services/api/src/resources/environment/resolvers.ts
+++ b/services/api/src/resources/environment/resolvers.ts
@@ -317,10 +317,6 @@ export const addOrUpdateEnvironment: ResolverFn = async (
   { input },
   { sqlClientPool, hasPermission, userActivityLogger }
 ) => {
-  const inputDefaults = {
-    deployHeadRef: null,
-    deployTitle: null
-  };
 
   // @ts-ignore
   const pid = input.project.toString();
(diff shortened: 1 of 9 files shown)


services/api/src/resources/environment/resolvers.ts
```diff
@@ -410,20 +418,32 @@ export const addOrUpdateEnvironmentStorage: ResolverFn = async (
     ...unformattedInput,
     updated: unformattedInput.updated
       ? unformattedInput.updated
-      : convertDateToMYSQLDateTimeFormat(new Date().toISOString())
+      : convertDateToMYSQLDateFormat(new Date().toISOString())
   };
 
-  const rows = await query(
+  let updatedDate = knex.fn.now();
```
rocketeerbkw: This looks like an unused variable?

services/api/src/resources/notification/sql.ts
```diff
@@ -33,7 +33,7 @@ export const Sql = {
         { type: type, content_type: contentType }
       )
       .where('nt.name', '=', name)
-      .select('nt.*', 'pn.*', knex.raw('? as type', [type]))
+      .select('nt.*', 'pn.*', knex.raw('? as orig_type', [type]))
```
rocketeerbkw, line 36: This looks like you're just clearing up a "confusing variable name" issue, but I don't see `orig_type` anywhere else so want to make sure

rocketeerbkw (dismissed):

tobybellwood:
Tested quite extensively on a fully loaded local cluster - can't see any adverse impacts, all appears to work correctly!

services/api/src/resources/notification/sql.ts
bomoko: @rocketeerbkw - it actually has to do with the fact that it was bombing out because of duplication of the name "type", which actually appears in project_notification. The two column names were causing issues with ambiguity.

services/api/src/resources/environment/resolvers.ts
bomoko: ooh, I'll remove in that case :) Thanks

services/api/src/resources/notification/sql.ts
bomoko: I'll double check this to make 100% sure though

bomoko: Yep - it's a bug in the existing query.

services/api/src/resources/environment/resolvers.ts
bomoko: Done

Issue #2844, opened 2021-09-22
Improve Opensearch compatibility


smlx:
Opensearch and Opensearch Dashboards is the open source successor of Opendistro Elasticsearch and Kibana. Since Opendistro is no longer maintained Lagoon will have to migrate to the Opensearch stack going forward.

Lagoon has a couple of compatibility issues dealing with the Opensearch stack vs Opendistro. Here are the problems that I've noticed from running the current `opendistro-sync` script:

## Creating index patterns

```
[2021-09-21 15:30:29] [error]: Kibana Error during setup of index pattern "application-logs-*": HTTPError: Response code 400 (Bad Request)
```
I think this one is due to the rename of the `kbn-xsrf` header to `osd-xsrf`? This `curl` command works:
```
curl -sSL -u "$USER_AUTH" 'https://opensearch.example.com/api/saved_objects/index-pattern' -XPOST -H 'Content-Type: application/json' -H 'securitytenant: exampletenant' -H 'osd-xsrf: true' --data-raw '{"attributes":{"title":"router-logs-*","timeFieldName":"@timestamp"}}'
```

## Configuring default index (?)

I'm not actually sure what this is supposed to do, but this error is logged:
```
[2021-09-21 15:30:29] [error]: Kibana Error during config of default Index: HTTPError: Response code 404 (Not Found)
```

## Backend role mapping

Opensearch needs the backend role mapped to the regular role otherwise logging in via Keycloak gives no permissions. In practice for the way that Lagoon uses roles this means adding the name of the role as a backend role. e.g. this curl command works:
```
curl -ksSL -u "$USER_AUTH" -XPUT 'https://localhost:9200/_plugins/_security/api/rolesmapping/example-lagoon-group' -H 'Content-Type: application/json' --data-raw '{"users":[],"backend_roles":["example-lagoon-group"],"hosts":[]}'"
```

tobybellwood:
so - looking back at this - and checking the Lagoon that is running Opensearch

1. Has been done
2. Needs some investigation (but I don't think is a deal breaker, as a default index is selected as a backup anyway)
3. Has been done

PR #3017, opened 2022-02-03
Assorted OpenDistro and OpenSearch fixes

tobybellwood:
This PR does the following:

* adds `rejectUnauthorized: false` to the got commands to interact with Kibana and OpenDistro/Opensearch security - Whilst the elasticsearch plugin can handle the varied authentication, got can't - so we need this. Given this comminication is within the cluster, and to endpoints defined within the API service, the risk of setting this is minimal
* adds the `'osd-xsrf': 'true'` header to the kibanaClient to handle OpenSearch Dashboards, as well as Kibana
* adds `indices:monitor/settings/get` to intial index permissions for groups to enable reporting


local-dev/lagoon-logging.values.yaml
```diff
@@ -2,6 +2,9 @@ logsDispatcher:
   serviceMonitor:
     enabled: false
   replicaCount: 1
+  image:
+    repository: testlagoon/logs-dispatcher
+    tag: pr-3009
```
smlx: I guess this should be removed before merge?

services/api/src/resources/group/opendistroSecurity.ts
```diff
@@ -45,7 +45,7 @@ export const OpendistroSecurityOperations = (
         index_permissions: [
           {
             index_patterns: [],
-            allowed_actions: ['read']
+            allowed_actions: ['read','indices:monitor/settings/get']
```
smlx, line 48: What does this change allow?

smlx (approved):
I have no way to test this, but LGTM otherwise. Also goes some way towards fixing https://github.com/uselagoon/lagoon/issues/2844

services/api/src/resources/group/opendistroSecurity.ts
tobybellwood: This is one of the core permissions required to add reporting access to user accounts in ODFE. Without this permissions, it appears that users are unable to select the data needed to create reports.

local-dev/lagoon-logging.values.yaml
tobybellwood: ah yes - that's the merged PR for fluent plugin updates #3009

PR #3021, opened 2022-02-10
Fixes rogue semicolon in custom task creation

bomoko:
This Fixes an error with custom tasks with no arguments. We were inserting a semicolon by default, which was breaking simple commands.


tobybellwood:
Replicated and tested locally, much success!
image

PR #3031, opened 2022-02-17
Adjust permissions on openshift:view

shreddedbacon:
# Checklist

The UI exposes some openshift fields to users now, but permissions errors occur for certain users

This makes `openshift:view` available to Guest and above now to allow the UI to display this information, but also implements some features to omit fields unless the user has `openshift:viewToken` permission

### DIFF
--- a/docs/administering-lagoon/rbac.md
+++ b/docs/administering-lagoon/rbac.md
@@ -109,7 +109,7 @@ Here is a table that lists the roles and the access they have:
 | deleteOpenshift | openshift | delete |  | Yes | Yes | No | No | No | No | No |  |
 | deleteAllOpenshifts | openshift | deleteAll |  | Yes | Yes | No | No | No | No | No |  |
 | getAllOpenshifts | openshift | viewAll |  | Yes | No | No | No | No | No | No |  |
-| getOpenshiftByProjectId | openshift | view | projectID | Yes | Yes | Yes | Yes | No | No | No |  |
+| getOpenshiftByProjectId | openshift | view | projectID | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
 | addNotificationToProject | project | addNotification | projectID | Yes | Yes | Yes | Yes | No | No | No |  |
 | removeNotificationFromProject | project | removeNotification | projectID | Yes | Yes | Yes | Yes | No | No | No |  |
 | addProject | project | add |  | Yes | Yes | Yes | Yes | Yes | Yes | Yes |  |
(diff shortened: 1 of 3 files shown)


rocketeerbkw:
I have some reservations about releasing changes that make permissions more permissive in a patch release, is guest required for the fix or just a bonus for this PR?

The changes in the `getOpenshiftByEnvironmentId` resolver are an obvious bug, but it would be nice if we could also rename the `view:token` scope and refactor to not use `attrFilter` anymore. If we split the changes, the former could get merged now and fix the UI, and the later could go in the next minor(?) release.

shreddedbacon:
> I have some reservations about releasing changes that make permissions more permissive in a patch release, is guest required for the fix or just a bonus for this PR?
> 
> The changes in the `getOpenshiftByEnvironmentId` resolver are an obvious bug, but it would be nice if we could also rename the `view:token` scope and refactor to not use `attrFilter` anymore. If we split the changes, the former could get merged now and fix the UI, and the later could go in the next minor(?) release.

Guest is required for displaying `openshift|kubernetes` information per environment in the UI, since currently maintainer and above is required, but that means anyone else will get a permission error when using the UI, not great.

tobybellwood:
When the openshift permission was conceived, there was little data there of use to the general user. Now that we store (specifically customer-facing) metadata against the object, I think a re-ratification of the permission is warranted. Given that this PR also includes additional safeties (consoleURL and alertmanager configs), I think this is appropriate for inclusion in a release - whether a patch or minor release.

shreddedbacon:
Once we finally remove the requirement for `token` and `consoleUrl` in the API (I think only ssh and storage-calculator are the remaining services using these), then we can remove the need for the token view scope and attrFilter checks, monitoringConfig will eventually move to the remote-controller instead

rocketeerbkw:
> anyone else will get a permission error when using the UI

That would only be the case if the UI components don't properly check for fields that have permission errors, it doesn't automatically error the whole page if a single field in a query has an error of some kind.

I further researched my security concerns and I don't feel as nervous about it. the `openshift:view` permission is only used for resolvers that are project based, meaning it's not possible for guests to get information about _any_ openshift, just ones they have access to via a project role. There also aren't many fields that other lagoon users could stuff potentially sensitive information into so it should be low risk of exposing anything to "guests" in that scenario.

Why do we consider the `consoleUrl` to need higher permissions, but `sshHost`/`sshPort` are still visible to guests?

shreddedbacon:
> Why do we consider the `consoleUrl` to need higher permissions, but `sshHost`/`sshPort` are still visible to guests?

consoleUrl exposes the kubernetes API endpoint

sshHost and sshPort are used mainly in tests from what I can see now, but they could be utilised more in the future once we move to ssh-portal and would be a way to expose the ssh endpoint for customers to use. Guests may not necessarily need to know these, but if they have access to a project, they probably already know the ssh endpoint. RBAC would still prevent them from doing anything via SSH even if they tried though too.

PR #3045, opened 2022-02-23
Feature/workflows

bomoko:
The following PR introduces the "Workflows" feature to Lagoon.

Essentially this is an extension of the custom task system that allows users to specify tasks that respond to standard Lagoon events. Post deployment, for instance, we may need to run some custom image based task in the namespace.

# API Changes

This introduces changes to the api to add/remove/resolve workflows, as well as making some changes to the custom task system (as well as their associated type changes):

* addWorkflow
* updateWorkflow
* deleteWorkflow
* workflowsForEnvironment
* getWorkflowsByEnvironmentId
* resolveAdvancedTaskDefinitionsForWorkflow

Perhaps most important is the structure of the workflow.

```
 type Workflow {
    id: Int
    name: String
    event: String
    project: Int
    advancedTaskDefinition: AdvancedTaskDefinition
  }
```

The important point here is the "event" type. Currently the events recognized are defined here https://github.com/uselagoon/lagoon/blob/14fd77347ad940e3aee2a4d8ce2d878fec328f0b/services/workflows/internal/lagoonclient/eventClassification.go

# New service introduced

This PR also introduces the service ("workflows") that watches the "lagoon-log" queue for incoming events. On an event, workflows will query the API to determine if there is an associated workflow for the project and, if there's a match, it will invoke workflow.

# Checklist

# Closing issues


services/api/src/resources/task/task_definition_resolvers.ts
```diff
@@ -616,3 +671,74 @@ const getAdvancedTaskTarget = advancedTask => {
   }
 };
 
+// const advancedTaskFunctions = sqlClientPool => {
```
timclifford, line 674: FYI: commented out code

PR #3056, opened 2022-03-08
Introducing s3 insights file handling to the API

timclifford:
WIP: not sure an additional table is required for storing insights data since we could just retrieve directly from s3


tobybellwood:
I would definitely not be storing all insights in the db! We should probably consider keeping the key facts there for ease of access (given they're to be used all over the place), but that would be limited to 5-6 per environment.

services/api/src/resources/insight/resolvers.ts
```diff
@@ -0,0 +1,169 @@
...
+
+const convertDateFormat = R.init;
+
+
+// Get insights files directly from the bucket
+export const getInsightsBucketFiles = async ({ prefix }) => {
+	try {
+    const data = await s3Client.listObjects({ Bucket: bucket, Prefix: prefix }, function(err, data) {
+      if (err) {
+        console.log("Failed to get items: ", err);
```
bomoko: Is this actually an error?

bomoko: That is, should it come out in the graqhQL response?

services/api/src/resources/insight/resolvers.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  try {
+    let insightsFile = 'insights/'+projectData.name+'/'+environmentName+'/'+file
+    const data = await s3Client.getObject({Bucket: bucket, Key: insightsFile}).promise();
+
+    if (!data) {
+      return null;
+    }
+
+    if (data.ContentEncoding == 'gzip') {
```
bomoko: Just to be clear - the idea is that we don't send non-textual data via this resolver?

services/api/src/resources/insight/resolvers.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (!eid) {
+    throw "No Environment ID given.";
+  }
+
+  const environmentData = await environmentHelpers(
+    sqlClientPool
+  ).getEnvironmentById(parseInt(eid));
+
+  if (!environmentAuthz) {
+    await hasPermission('environment', 'view', {
```
bomoko, line 136: I wonder if this shouldn't require elevated perms?

timclifford: this check only happens if a permissions check hasn't already happened in the environment resolver

services/api/src/resources/insight/resolvers.ts
timclifford: yeah - will remove this, no need for it

services/api/src/resources/insight/resolvers.ts
```diff
@@ -0,0 +1,162 @@
...
+  );
+  const environmentName = getEnvironmentName(environmentData, projectData);
+
+  const insightsItems = await getInsightsBucketFiles({ prefix: 'insights/'+projectData.name+'/'+environmentName+'/'});
+  const files = await Promise.all(insightsItems.map(async (file, index) => {
+
+    const fileName = file.Key.split("/").pop();
+    return {
+      id: index+1,
+      fileId: file.Owner.ID,
```
tobybellwood: is this causing this error?
image

services/api/src/resources/insight/resolvers.ts
```diff
@@ -0,0 +1,162 @@
...
+import { getEnvironmentName, convertBytesToHumanFileSize } from './helpers';
+import { Helpers as environmentHelpers } from '../environment/helpers';
+import { Helpers as projectHelpers } from '../project/helpers';
+
+import S3 from 'aws-sdk/clients/s3';
+
+// s3 config
+const accessKeyId =  process.env.S3_FILES_ACCESS_KEY_ID || 'minio'
+const secretAccessKey =  process.env.S3_FILES_SECRET_ACCESS_KEY || 'minio123'
+const bucket = process.env.S3_INSIGHTS_BUCKET || 'lagoon-insights'
```
tobybellwood: The insights handler just assumes that the S3_INSIGHTS_BUCKET is the S3_FILES_BUCKET - maybe we should here too?

services/api/src/resources/insight/resolvers.ts
```diff
@@ -0,0 +1,164 @@
...
```
tobybellwood:
```suggestion
const bucket = process.env.S3_FILES_BUCKET || 'lagoon-files'
```

PR #3061, opened 2022-03-11
Migration to make fact unqiue by env-name-source

timclifford:
# Checklist

# Closing issues

### DIFF
--- a/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
+++ b/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
@@ -1600,7 +1600,7 @@ CALL add_fact_source_and_description_to_environment_fact();
 CALL add_fact_type_to_environment_fact();
 CALL add_fact_category_to_environment_fact();
 CALL add_fact_key_to_environment_fact();
--- CALL update_unique_environment_fact();
+CALL update_unique_environment_fact();
 CALL add_metadata_to_project();
 CALL add_content_type_to_project_notification();
 CALL convert_project_production_routes_to_text();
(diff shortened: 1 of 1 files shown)


services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -1094,6 +1094,24 @@ CREATE OR REPLACE PROCEDURE
...
+CREATE OR REPLACE PROCEDURE
+  update_unique_environment_fact()
+
+  BEGIN
+    IF NOT EXISTS(
+      SELECT NULL
+      FROM INFORMATION_SCHEMA.COLUMNS
+      WHERE
+        table_name = 'environment_fact'
+        AND table_schema = 'infrastructure'
```
bomoko: Hey! I don't think this is doing what we need it to do.
If you run a `show index from environment_fact` you'll see that the unique index's name we're interested in is called "environment"
So we want to look for `environment`, and if we find it drop it, 
i.e. 
ALTER TABLE environment_fact DROP INDEX environment;
Then, let's name the new index as well, by creating it something like this
ALTER TABLE environment_fact ADD CONSTRAINT environment_fact_unique UNIQUE(environment, name, source);

bomoko (changes requested):

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
bomoko: Did it :)

bomoko (approved):

bomoko:
Closing this because these commits are in another PR

PR #3064, opened 2022-03-14
environment_fact table updates - service column and unique constraint

timclifford:
This PR updates the environment_fact table to add a column for "service" to store the lagoon service that the fact relates to.

It also adds an additional migration to modify the uniqueness constraint for that table to accomodate the concept of multiple services returning similar facts (as with Drupal -which reports in the cli, nginx and php services usually)


bomoko (commented):
LGTM

PR #3074, opened 2022-03-18
chore: remove random wait from test suite

smlx:
# Checklist

We now have proper readiness signalling in keycloak so the random wait should no longer be necessary.

This will make test runs slightly faster.

# Closing issues

n/a


tobybellwood:
Fine, you win, merging!

PR #3080, opened 2022-03-24
add rootless rsync commands to drush rsync task

Schnitzel:
With rootless systems running a normal `drush rsync` command causes some issues as rootless systems (like openshift or k8s rootless) do not allow the executing user to do some of the standard things that rsync wants to do (change owner and group to what the source had).
Therefore we tell `drush rsync` to run `rsync` with some additional parameters that will prevent it to fail

This is also documented here: https://docs.lagoon.sh/drupal/drush-9/#drush-rsync-from-local-to-remote-environments

 

# Checklist

# Closing issues


timclifford:
LGTM 👍

Issue #3082, opened 2022-03-28
Insights version 2 work - remove Lagoon Core Harbor/Trivy/Drutiny integration


bomoko:
With the move to Lagoon 2 and remote Harbors, the current Harbor/Trivy/Drutiny integration needs to be reworked, since there is presently no mechanism for the Harbor webhook setup talking back to Lagoon core that guarantees webhook integration - further, Core's Harbor/Trivy integration simply doesn't support multiple harbors. 

We are in the process of completely reworking the insights system, and so the previous iteration needs to be removed from Core. This requires

1 - a first version that deprecates but still allows the integration to be enabled if required.
2 - actually removing the Harbor/Trivy/Drutiny integration.

bomoko:
This issue didn't address Drutiny integration in closing. Reopening.

tnelson-doghouse:
Hi @bomoko !  
Is there a ticket I can follow (or documentation to read) where there's discussion of giving Trivy access to Lagoon customers?  By which I mean we can see the insights generated, etc

Thanks!

bomoko:
Hi @tnelson-doghouse 

So SBOM scans are already part of Lagoon - driven primarily by https://github.com/uselagoon/insights-remote and https://github.com/uselagoon/insights-handler.

These are enabled at a cluster level - you can bring this up with your TAM.
With it you get scanning and SBOM downloads, which are potentially very useful.

There are some important caveats here
* with Insights remote scanning enabled, builds will be slightly longer, as the SBOM generation will happen as the last step of the build deploy process
* You'll need to enable the facts and problems UI if you want menu items to show up in the UI (by default they're hidden).

tnelson-doghouse:
Thanks @bomoko !  

I'm already familiar with facts/problems/metadata; we don't use them at the moment, but I've considered adding them to our site setup process, and would if we were using them.  

As for the rest, I'll take it to our TAM.  

Thanks!

bomoko:
It's a pleasure @tnelson-doghouse 

The problems system, once insights are enabled, is essentially the output of a trivy scan. You'll also, as mentioned, be able to access SBOMs from the UI if you want to do further processing.

Feel free to pull me into whatever chats you have if you need anything further.

PR #3083, opened 2022-03-28
Dep. trivy integration from core

bomoko:
This addresses point 1 of #3082 by marking as deprecated the Harbor/Trivy integration in core but making it optional.


timclifford:
LGTM :D

Issue #2907, opened 2021-11-03
Remove Harbor functionality from lagoon-core


Schnitzel:
Harbor is now handled by lagoon-remote and therefore we can remove it from lagoon-core.
Also we should make sure that we have release notes that explain what is to do when upgrading.

shreddedbacon:
Part of this is solved in #3086

PR #3086, opened 2022-03-28
Feature/expunge harbor trivy from core

### LINKED ISSUE #3082: Insights version 2 work - remove Lagoon Core Harbor/Trivy/Drutiny integration

bomoko:
This PR addresses point 2 of #3082 and SHOULD NOT BE MERGED until we are comfortable that the Harbor/Trivy integration dependencies are done with.

closes #3082


shreddedbacon:
@tobybellwood @bomoko would be good to try and get this out of core to help address #2907. Especially after the issues we hit today.

All our tests etc don't use the core integration of harbor anymore, and our docs say not to configure core to talk to harbor. 

Do we have some blocker on the IO side that we need to resolve?

tobybellwood:
@bomoko to tidy up and move forwards with

bomoko:
Closing in favor of #3672

PR #3089, opened 2022-03-29
feature: add a timeout on the rollout status watch

### LINKED ISSUE #2861: Prevent long running (failed) deployments where practical
It would be useful if there was a way to be notified of long running deployments once they reach a certain time limit. Or for a central service desk to to be notified (for example GovCMS)

image

shreddedbacon:
# Checklist

Sometimes `kubectl rollout status` can get stuck indefinitely waiting for a rollout to complete due to an issue with the previous pod in the existing replicaset possibly not terminating.

Adding a `--timeout=1200s` to this status check is double what the `progressDeadlineSeconds` value is, and if a pod rollout is taking more than 20 minutes to complete, there is probably something wrong anyway.

# Closing issues

Partially closes #2861 as it won't actually report on a long running notification, but will prevent the condition where a long running deployment would occur. With intermittent build logs now, it can be more obvious where a build has stalled, before calling it "long running"

### DIFF
--- a/images/kubectl-build-deploy-dind/scripts/exec-monitor-deploy.sh
+++ b/images/kubectl-build-deploy-dind/scripts/exec-monitor-deploy.sh
@@ -32,7 +32,10 @@ stream_logs_deployment &
 STREAM_LOGS_PID=$!
 
 ret=0
-kubectl rollout --insecure-skip-tls-verify -n ${NAMESPACE} status deployment ${SERVICE_NAME} --watch || ret=$?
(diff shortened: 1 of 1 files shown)


seanhamlin:
Confirming this does what we expect (altered the timeout to be lower for testing):

```
$ kubectl -n ns-dev rollout status deployment solr --watch --timeout=1m
Waiting for deployment "solr" rollout to finish: 0 out of 1 new replicas have been updated...
error: timed out waiting for the condition

$ echo $?
1
```

Issue #2189, opened 2020-09-08
Support ECDSA SSH keys and FIDO/U2F key types in Lagoon


smlx:
> The YubiKey 5 Series supports the following algorithms on the PIV smart card application.
>
>    RSA 1024
>    RSA 2048
>    ECC P-256
>    ECC P-384

https://support.yubico.com/support/solutions/articles/15000014219-yubikey-5-series-technical-manual

I'd like to use my Yubikey to interact with Lagoon via SSH, but only RSA and Ed25519 keys are supported by Lagoon. For various reasons RSA 2048-bit keys aren't great so it would be nice to add support for at least `ecdsa-sha2-nistp256`, and possibly `ecdsa-sha2-nistp384`.

smlx:
Lagoon should also support the FIDO/U2F key types `ecdsa-sk` and `ed25519-sk`.
https://www.openssh.com/txt/release-8.2

tobybellwood:
As long as the key type is parsable by [sshpk](https://github.com/joyent/node-sshpk) - it should be easyish to add in.

I'm not sure that the *-sk are supported by sshpk, so we may need to look to validate somehow else?

smlx:
Looks like it isn't supported https://github.com/joyent/node-sshpk/issues/72

smlx:
Even the non `sk` versions will make it easier to use Yubikeys.

tobybellwood:
ECDSA key support is now in Lagoon

tobybellwood:
Going to reopen this to track the -sk key types

PR #3099, opened 2022-03-31
ECDSA ssh key type support

cdchris12:
This PR adds support for newer variations of SSH keys to Lagoon. As of this PR, the following ssh key types are supported:

- `ssh-rsa`
- `ssh-ed25519`
- `ecdsa-sha2-nistp256`
- `ecdsa-sha2-nistp384`
- `ecdsa-sha2-nistp521`

# Checklist

Given that the only limitation of our stack in supporting newer ssh key types is our GraphQL API, UI, and the `api-db` component, it is relatively trivial to enable support for some newer SSH key types. Unfortunately, the JS module we use for ssh key verification, `sshpk`, does not support FIDO based ssh keys currently, per https://github.com/joyent/node-sshpk/issues/72 , so we are unable to add support for these key types quite as easily.

# Affected issues

Partially satisfies the requirements of #2189


services/api/src/resources/sshKey/resolvers.ts
```diff
@@ -10,6 +10,9 @@ const formatSshKey = ({ keyType, keyValue }) => `${keyType} ${keyValue}`;
 const sshKeyTypeToString = R.cond([
   [R.equals('SSH_RSA'), R.always('ssh-rsa')],
   [R.equals('SSH_ED25519'), R.always('ssh-ed25519')],
+  [R.equals('ECDSA-SHA2-NISTP256'), R.always('ecdsa-sha2-nistp256')],
+  [R.equals('ECDSA-SHA2-NISTP384'), R.always('ecdsa-sha2-nistp384')],
+  [R.equals('ECDSA-SHA2-NISTP521'), R.always('ecdsa-sha2-nistp521')],
```
seanhamlin: Should this be `ECDSA_SHA2_NISTP521` with underscores?

cdchris12: Good catch; it certainly should. I've pushed a fix :)

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
```diff
@@ -552,7 +552,7 @@ CREATE OR REPLACE PROCEDURE
         AND column_name = 'key_type'
     ) THEN
       ALTER TABLE `ssh_key`
-      CHANGE `keyType` `key_type` ENUM('ssh-rsa', 'ssh-ed25519') NOT NULL DEFAULT 'ssh-rsa';
+      CHANGE `keyType` `key_type` ENUM('ssh-rsa', 'ssh-ed25519','ecdsa-sha2-nistp256','ecdsa-sha2-nistp384','ecdsa-sha2-nistp521') NOT NULL DEFAULT 'ssh-rsa';
```
rocketeerbkw: Our stored proc migrations are designed to be run only once, so this will never get run on production and the ENUM won't be updated. Instead of modifying an existing migration you'll need to create a new one. An existing migration that's very similar would be [`add_environment_type_to_advanced_task_argument`](https://github.com/uselagoon/lagoon/blob/main/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql#L1335-L1355).

services/ui/src/components/SshKeys/AddSshKey.js
```diff
@@ -80,7 +83,7 @@ const AddSshKey = ({me: { id, email }}) => {
                   type="text"
                   onChange={handleChange}
                   value={values.sshKey}
-                  placeholder="Begins with 'ssh-rsa', 'ssh-ed25519'"/>
+                  placeholder="Begins with 'ssh-rsa', 'ssh-ed25519', 'ecdsa-sha2-nistp256', 'ecdsa-sha2-nistp384', 'ecdsa-sha2-nistp521'"/>
```
rocketeerbkw, line 86: Did you happen to run the UI to look at this change? It's a really long placeholder now and I'm not sure how well it scales with length.

rocketeerbkw (changes requested):

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
tobybellwood: I'm building this locally to review the same!

services/ui/src/components/SshKeys/AddSshKey.js
cdchris12: That text field is configured to wrap on dashes or spaces, so it should be good for most modern browser window sizes, I'd expect?

Screen Shot 2022-04-01 at 12 01 09 PM

services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
cdchris12: This should be resolved by https://github.com/uselagoon/lagoon/blob/684a583420db3fd78d20fef91021f3e4a5c68b55/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql#L1579-L1599

services/ui/src/components/SshKeys/AddSshKey.js
rocketeerbkw: Not gonna win any design awards but as long as it's functional...

rocketeerbkw (approved):

tobybellwood:
ok - I've E2E tested this, and picked up a few things, but it's now looking good!

I've added a few commits here - there were two important fixups:
* [`e5e431f` (#3099)](https://github.com/uselagoon/lagoon/pull/3099/commits/e5e431f27acbbb8a0f3a3a4bf1fd27a4251b46a5) to make sure the temporary column used to transition the ENUM was large enough to hold the new value
* [`346c6f0` (#3099)](https://github.com/uselagoon/lagoon/pull/3099/commits/346c6f0a6dace773f787c283242a75a8aab0f6a5) where the UI would only lowercase the first `-` in ecdsa-sha2-nistpXXX, causing a mismatch in the ENUMs and resolvers

I've configured the test data pushers to create an additional ecdsa user to ensure the process flows (and added this user's key into the local-dev as cli_id_ecdsa.

Also updated a chunk of docs, stories and mocks for ecdsa keys.

We will need to update the lagoon-cli to support these keys (https://github.com/uselagoon/lagoon-cli/blob/main/cmd/users.go#L40-L41) - https://github.com/uselagoon/lagoon-cli/issues/217 raised

cdchris12:
The corresponding lagoon-cli PR is https://github.com/uselagoon/lagoon-cli/pull/218

PR #3112, opened 2022-04-08
feat: API changes to support cancelling tasks

### LINKED ISSUE #3106: Ability to cancel tasks
Currently it isn't possible to cancel tasks. Sometimes tasks could appear to be long running and a user may want to cancel the task.

We should handle this in a similar way to how cancelling a deployment works

timclifford:
# Checklist

This adds support to the API to be able to cancel tasks. There is a permission split to do by production and development environment, and the roles are configured by default to be `developer` can cancel development environment tasks, and `maintainer` can cancel production and development environment tasks.

The cancelTask mutation can be called like so
```
mutation cancelTask {
  cancelTask(input:{
    task: {
      taskName: "lagoon-task-pt2pp5"
      environment: {
        id: 3
        project:{
          id: 18
        }
      }
    }
  })
}
```

Changes to the UI will be required to support the cancellation of tasks in https://github.com/uselagoon/lagoon-ui

# Closing issues

closes #3106


shreddedbacon:
I like, but theres some things I think we can do better to better handle cancelling tasks (the controller needs to be updated too, but we should add the initial support into core before we expose task cancelling in the api)

rocketeerbkw:
Would close #3106

tobybellwood:
Now that the controller can cancel tasks - the approach needs revisiting. @shreddedbacon to assess refactor/rewrite

services/api/src/resources/task/resolvers.ts
```diff
@@ -15,7 +15,9 @@ import { Validators as envValidators } from '../environment/validators';
 import S3 from 'aws-sdk/clients/s3';
 import sha1 from 'sha1';
 import { generateTaskName } from '@lagoon/commons/dist/util/lagoon';
-import { logger } from '../../loggers/logger';
+// import { logger } from '../../loggers/logger';
```
bomoko: If we're not using this, we should remove this line

services/api/src/resources/task/helpers.ts
```diff
@@ -1,113 +1,39 @@
...
-        service,
-        command,
-        deployTokenInjection,
-        projectKeyInjection,
-        adminOnlyView,
-        remoteId,
-      }),
-    );
+import { Helpers as environmentHelpers } from '../environment/helpers';
+// import convertDateToMYSQLDateTimeFormat from '../../util/convertDateToMYSQLDateTimeFormat';
```
bomoko: Deleting is probably better if we're not using this

services/api/src/resources/task/helpers.ts
```diff
@@ -168,54 +140,174 @@ export const Helpers = (sqlClientPool: Pool) => ({
...
+            }
 
-    return taskData;
-  },
-});
+            return task;
+          }
+        );
+
+        const taskFromNameEnv = async input => {
```
bomoko, line 266: I'm going to assume that this function and the related ones are the new code? The diff is difficult to read ...

services/api/src/resources/task/helpers.ts
```diff
@@ -168,54 +140,174 @@ export const Helpers = (sqlClientPool: Pool) => ({
...
+          const environments = await environmentHelpers(
+            sqlClientPool
+          ).getEnvironmentsByEnvironmentInput(R.prop('environment', input));
+          const activeEnvironments = R.filter(
+            R.propEq('deleted', '0000-00-00 00:00:00'),
+            environments
+          );
+
+          if (activeEnvironments.length < 1 || activeEnvironments.length > 1) {
```
bomoko: `activeEnvironments.length != 1` maybe?

Also, I'm a little puzzled by the error - is "Unauthorized" appropriate? (it might be, of course! Just wanted to check)

services/api/src/resources/task/resolvers.ts
```diff
@@ -329,6 +331,71 @@ export const deleteTask: ResolverFn = async (
...
+    }
+  },
+  { sqlClientPool, hasPermission, userActivityLogger }
+) => {
+
+  const task = await Helpers(sqlClientPool, hasPermission).getTaskByTaskInput(taskInput);
+  const environment = await environmentHelpers(sqlClientPool).getEnvironmentById(task.environment);
+  const project = await projectHelpers(sqlClientPool).getProjectById(environment.project);
+
+  if (!task) {
```
bomoko, line 344: This could be moved under line 344 and skip the two other calls on 345 and 346

services/api/src/resources/task/resolvers.ts
```diff
@@ -329,6 +331,71 @@ export const deleteTask: ResolverFn = async (
...
+    return null;
+  }
+
+  const envPerm = await environmentHelpers(sqlClientPool).getEnvironmentById(
```
bomoko: we already pull the environment on line 345 aren't we? Is there a side effect here or something that requires this to be pulled again?

bomoko (commented):
Looks good, and should all work - just added a couple comments/questions. Nice PR!

services/api/src/resources/task/helpers.ts
shreddedbacon: Yep

services/api/src/resources/task/resolvers.ts
shreddedbacon: nah, just duped code

PR #3117, opened 2022-04-19
Fix `add_ecdsa_ssh_key_types` migration

### LINKED ISSUE #3116: ECDSA ssh key `api-db` migration never runs
# Describe the bug

We added new ssh key types in #3099 which required updating the `api-db` schema. The migration procedure has a typo which causes it to never be run.

## To Reproduce

1. Remove the new enums from the [create table statement](https://github.com/uselagoon/lagoon/blob/main/services/api-db/docker-entrypoint-initdb.d/00-tables.sql#L9)
2. Spin up the `api-db`
3. See that the enum hasn't been updated

## Expected behavior

The database should be in a correct state for fresh and upgraded states.

rocketeerbkw:
# Checklist

The migration stored procedure has a superfluous `space` which causes it to think it doesn't need to be run. I ran the migration test query to double check the exact value that's required:

```
MariaDB [infrastructure]> select COLUMN_TYPE from information_schema.columns where table_name = 'ssh_key' and table_schema = 'infrastructure' and column_name = 'key_type';
+-------------------------------+
| COLUMN_TYPE                   |
+-------------------------------+
| enum('ssh-rsa','ssh-ed25519') |
+-------------------------------+
```

# Closing issues

closes #3116

### DIFF
--- a/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
+++ b/services/api-db/docker-entrypoint-initdb.d/01-migrations.sql
@@ -1623,7 +1623,7 @@ CREATE OR REPLACE PROCEDURE
       AND column_name = 'key_type';
 
     IF (
-      column_type_argument_type = "enum('ssh-rsa', 'ssh-ed25519')"
+      column_type_argument_type = "enum('ssh-rsa','ssh-ed25519')"
     ) THEN
       ALTER TABLE ssh_key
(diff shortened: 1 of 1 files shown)


rocketeerbkw:
Normally a migration shouldn't be altered once release. It's acceptable to modify the already released migration instead of writing a new one because the current version will have never run before, and after the new run is run, it will still never get run twice. This change doesn't effect the idempotents of the migration

PR #3137, opened 2022-05-12
fixup ingress and persistant storage for node-based tests

tobybellwood:
A number of node-based examples copy/pasted the .lagoon.yml files, and as a result, there may be some race conditions around ingress creation - the same ingress may be "already taken" by another competing test,

This PR removes the ingresses from all tests except features - where they are actually used

### DIFF
--- a/tests/files/features-autogenerated-routes-disabled/docker-compose.yml
+++ b/tests/files/features-autogenerated-routes-disabled/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     labels:
       lagoon.type: node-persistent
       lagoon.persistent: /files
+      lagoon.persistent.size: 500Mi
     volumes:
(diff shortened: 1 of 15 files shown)


shreddedbacon:
Worth dropping these to 100Mi too?
```
tests/files/drupal8-mariadb-single/docker-compose.yml:      lagoon.persistent.size: 1Gi
tests/files/drupal9-postgres/docker-compose.single.yml:      lagoon.persistent.size: 1Gi
tests/files/drupal9-postgres/docker-compose.yml:      lagoon.persistent.size: 1Gi
tests/files/drupal9-mariadb/docker-compose.single.yml:      lagoon.persistent.size: 1Gi
tests/files/drupal9-mariadb/docker-compose.yml:      lagoon.persistent.size: 1Gi
```

PR #3155, opened 2022-05-27
Stop using deprecated keycloak "upload scripts" for authorization policies

rocketeerbkw:
# Checklist

In Keycloak 18, the deprecated "upload scripts" features is finally being removed. In order to to import our authorization scripts we need to convert them to a custom provider. This is a proof of concept to show that we can make the change when the need arises.

# Closing issues

n/a


tobybellwood:
ok - tested this many which ways, and all checks out. Even situations where permission has been modified are saved and then changed to match new Policy format.

PR #3160, opened 2022-05-28
Improve project page performance

### LINKED ISSUE #3159: Projects with deploy targets may fail to load
# Describe the bug

The query used for the project page may cause excessive load to the api/keycloak when it is configured to use deploy targets which results in an error page.

## Additional context

* Projects without deploy targets don't seem to be affected
* It's most prevalent when a user has never loaded resources for that project before, and therefor has an empty permissions cache.

rocketeerbkw:
# Checklist

This PR improves the project page by:

* Reducing the amount of nested data being queried
* Removing unused fields from queries
* Reducing the number of `hasPermission` calls from 17 to 8
* Removing uncacheable openshift permission checks when not required

Additionally it cleans up no longer used `lib/query` files and openshift fields.

# Closing issues

closes #3159


services/api/src/apolloServer.js
```diff
@@ -191,6 +191,26 @@ const apolloServer = new ApolloServer({
     };
   },
   plugins: [
+    // Debug plugin for logging resolver execution order
```
bomoko, line 194: any way we can inject this with a startup flag rather than leaving in the comments?

bomoko (approved):
LGTM - just one question about the chunk of commented out debug code - but all make sense.

services/api/src/apolloServer.js
rocketeerbkw: I'm open to suggestions. I don't think we have anything in place to do something like that? I briefly considering keeping it uncommented and using `logger.silly` but 1) I'm not willing to YOLO some troubleshooting code into production that runs on every graphql request and 2) we default to `silly` in [production](https://github.com/uselagoon/lagoon/blob/main/node-packages/commons/src/local-logging.ts#L53).

This is the second time I've written a similar plugin and had to go read the apollo docs both times. Alternatives would be to add this in the docs somewhere?

Issue #2930, opened 2021-11-14
Reduce debug noise in API pod logs


seanhamlin:
# Describe the bug

There is a lot of log noise in the `api` container logs of  `lagoon-core`:

```
[2021-11-14 20:19:50] [debug]: Keycloak legacy auth failed: Error decoding token: invalid algorithm
[2021-11-14 20:19:50] [debug]: Keycloak token auth failed: Error decoding token: Grant validation failed. Reason: invalid token (expired)
```

I roughly guess 99% is the above debug noise, and < 1% is actual useful log information

## To Reproduce

1. Tail `api` container logs
2. Notice log noise

## Expected behavior

No debug information emitted

shreddedbacon:
This is becoming a problem when trying to actually debug issues

smlx:
Fixing this will also make it practical to increase Lagoon logs retention.

tobybellwood:
closing on #3164 - will address specific logging later

PR #3164, opened 2022-06-01
Reduce logging noise and allow log level to be configured in Node.js services

rocketeerbkw:
# Checklist

The following changes are made:
- Add `CONSOLE_LOGGING_LEVEL` environment variable with the default set to `info` instead of `debug` and standardize the levels used in all the loggers
- First pass at reducing the amount of useless log messages and/or changing severity where appropriate

Some back of the napkin guesses on how this may improve things:

1) Fun with the command line (not very useful info though):
    `grep --only-matching --recursive --no-messages --no-filename --exclude-dir=node_modules --exclude-dir=dist --context=0 --extended-regexp --regexp="logger\.\w+" services/ | sort | uniq -c`
    Before:
    ```
    53 logger.debug
    86 logger.error
    38 logger.info
    21 logger.verbose
    47 logger.warn
    ```
    After:
    ```
    52 logger.debug
    84 logger.error
     2 logger.fatal
    36 logger.info
    25 logger.verbose
    34 logger.warn
    ```
2) Fun with kibana (by excluding logs that closely match what have been removed in code, **~50% reduction**)
    Before:
    before
    After:
    after

Additionally, while testing log messages, I found a bug in `addOrUpdateEnvironment` resolver which is also added as a rider to this PR

Would love to see follow up PRs to do a full log message audit and the ability to change the log level at runtime, but this should help with the annoyances everyone has in production right now.

# Closing issues

First step towards #2930


node-packages/commons/src/logs/local-logger.ts
```diff
@@ -0,0 +1,41 @@
...
+    format.colorize(),
+    format.splat(),
+    format.timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }),
+    format.printf(
+      ({ timestamp, level, message }) => `[${timestamp}] [${level}]: ${message}`
+    )
+  ),
+  transports: [
+    new transports.Console({
+      level: getConfigFromEnv('LOGGING_LEVEL', 'info'),
```
rocketeerbkw: Naming things is hard, is `LOGGING_LEVEL` the right play? I also thought about `LOCAL_LOGGING_LEVEL` since this only affects `console.log` right now, but we discussed the possibility of having different severity levels for stdout and opensearch in the future.

services/api/src/authMiddleware.ts
```diff
@@ -33,108 +35,99 @@ const parseBearerToken = R.compose(
...
+        `Legacy authentication granted for '${username}' from '${source}'`,
+        {
+          user: legacyCredentials ? legacyCredentials : null,
+          headers: req.headers
+        }
+      );
+
+      return next();
+    } catch (e) {
+      res.status(401).send({
```
rocketeerbkw, line 87: I changed the HTTP error code in a few places from `403 Forbidden` to `401 Unauthorized` when there was an error evaluating the supplied token, is that a BC break we care about or not?

shreddedbacon: I don't think we need to care about this. 

For lagoon-cli, it will just capture all errors anyway so a change of 403 to 401 here is not a big issue for it. But things I don't know about too much being the UI, would it have an issue with this?

node-packages/commons/src/logs/local-logger.ts
rocketeerbkw: Decided on `CONSOLE_LOGGING_LEVEL`

services/api/src/authMiddleware.ts
rocketeerbkw: Tested a bunch of stuff, couldn't find any errors in our tools

tobybellwood (approved):
I've run this locally, and all seems to work. Will merge and test further on infra.

PR #3176, opened 2022-06-09
Add update deploy target for environment mutation

### LINKED ISSUE #3173: discussion + error: updateEnvironment unable to change openshift
# Describe the bug

On running the following mutation
```

mutation updateOpenshift ($envid: Int!, $targetID: Int!) {
  updateEnvironment(input:{
    id: $envid,
    patch: {
      openshift: $targetID
    }
  }) {
    id
    name
    openshift {
      id
      name
    }
  }
}
```
we receive the following error

`Unauthorized: You don't have permission to \"update:undefined\" on \"environment\": {\"project\":`

In speaking with @tobybellwood about this, he suggested that simply allowing users with the correct privileges to update the openshift field may be potentially problematic.

As a safety net, he suggests that we only allow the `openshift` to be updated if it is in accord with the current deploy target configuration. This would set up at least some minimal safety net in that changing the deploy target will at least be a two step process (i.e. update deploy target config, then update the field on the environment itself).

## To Reproduce

1. As a maintainer of the project, run the graphql given above with the appropriate values set
2. get the error back from the API

## Expected behavior

The openshift field value is updated.

## Additional context

This is an issue that is blocking writing cluster migration scripts

bomoko:
Discussion in issue #3173 

# Checklist

Adds new mutation updateEnvironmentDeployTarget to allow maintainers to, if the deploy targets are correctly set, update the deploy target for their environment.

# Closing issues

closes #3173

### DIFF
--- a/services/api/src/resources/deploytargetconfig/resolvers.ts
+++ b/services/api/src/resources/deploytargetconfig/resolvers.ts
@@ -145,6 +145,10 @@ export const updateEnvironmentDeployTarget: ResolverFn = async (
       }
     })
   );
+
+  return await environmentHelpers(
+    sqlClientPool
+  ).getEnvironmentById(parseInt(environment));
 }
 
 
--- a/services/api/src/resources/deploytargetconfig/resolvers.ts
+++ b/services/api/src/resources/deploytargetconfig/resolvers.ts
@@ -1,8 +1,13 @@
 import * as R from 'ramda';
 import { ResolverFn } from '../';
+import { Environment } from '../../resolvers';
+import { Helpers as environmentHelpers } from '../environment/helpers';
 import { isPatchEmpty, query, knex } from '../../util/db';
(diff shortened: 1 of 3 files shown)


services/api/src/resources/deploytargetconfig/resolvers.ts
```diff
@@ -79,6 +84,73 @@ export const getDeployTargetConfigsByDeployTarget: ResolverFn = async (
...
+  await query(
+    sqlClientPool,
+    EnvironmentSql.updateEnvironment({
+      id: environment,
+      patch: {
+        openshift: deployTarget,
+      }
+    })
+  );
+
```
tobybellwood, line 149: can we get an activity logger in here to report that the environment had it's deployTarget updated?

bomoko: there we go.

shreddedbacon (approved):
:shipit:

PR #3181, opened 2022-06-12
update build-deploy-tool and add docker-compose validation step

shreddedbacon:
# Checklist

A recent release of the build-deploy-tool added the ability to consume the docker-compose file. But due to some issues with some customers having broken or invalid docker-compose files this caused the `LAGOON_ROUTE` to not be displayed and the build-deploy-tool silently failed. Previous versions of Lagoon builds were less strict with this file so these errors went unnoticed for a lot of people.

This PR uses the updated build-deploy-tool v0.9.1 which uses the `compose-spec/compose-go` library to read the docker-compose file, but has some additional functionality to be able ignore some errors with validating the docker-compose file.

An additional step at the end of builds will now run to validate the docker-compose file. If the build-deploy-tool discovers an issue with the file, it will report a warning at the end of a build, this is the same as running `docker compose config` against your docker-compose file.

Additionally, we now also capture the `docker-compose.yml` into pre and post deploy fields in a new configmap.
This way we can analyse docker-compose files if required.

# Closing issues

https://github.com/uselagoon/build-deploy-tool/issues/42

### DIFF
--- a/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
+++ b/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
@@ -1431,3 +1431,29 @@ if [ "$(featureFlag INSIGHTS)" = enabled ]; then
   previousStepEnd=${currentStepEnd}
 fi
 set -x
+
+set +x
+##############################################
+### RUN docker compose config check against the provided docker-compose file
+### use the `build-deploy-tool` built in validater to run over the provided docker-compose file
+##############################################
+dcc=$(bash -c 'build-deploy-tool validate docker-compose --docker-compose '${DOCKER_COMPOSE_YAML}' --ignore-non-string-key-errors=false; exit $?' 2>&1)
+dccExit=$?
+if [ "${dcc}" != "0" ]; then
+  echo "
+##############################################
+Warning!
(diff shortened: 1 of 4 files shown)


images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh
```diff
@@ -1431,3 +1457,29 @@ if [ "$(featureFlag INSIGHTS)" = enabled ]; then
...
+dccOutput=$(bash -c 'build-deploy-tool validate docker-compose --docker-compose '${DOCKER_COMPOSE_YAML}' --ignore-non-string-key-errors=false; exit $?' 2>&1)
+dccExit=$?
+if [ "${dccExit}" != "0" ]; then
+  echo "
+##############################################
+Warning!
+There are issues with your docker compose file that lagoon uses that should be fixed.
+This does not currently prevent builds from proceeding, but future versions of Lagoon *will* be more strict on issues shown here.
+The following is output from \`build-deploy-tool validate docker-compose --docker-compose '${DOCKER_COMPOSE_YAML}' --ignore-non-string-key-errors=false\` which you can run locally to verify and fix.
+##############################################
```
shreddedbacon, line 1475: @tobybellwood I don't know if this is the best wording for this? Thoughts?

PR #3184, opened 2022-06-12
Admin and non-UI tasks

### LINKED ISSUE #3182: Explicitly setting Admin level Custom Tasks
https://github.com/uselagoon/remote-controller/pull/133 makes the loading of ssh keys and deploy tokens explicit.

This requires us to explicitly request these to be injected into image based advanced tasks.

Proposal:
Add an `adminTask` boolean that is set to FALSE by default and can only be set to TRUE to those with the `create:advanced` permission for advanced tasks.

### LINKED ISSUE #3183: Non-UI custom tasks
We want to create custom tasks that do not appear in the UI - these are, for instance, tasks that might only ever be kicked off by automated processes or by admins.

Proposal: add a `showUi` field to the custom task definitions that is set to TRUE by default. This will allow explicitly hiding tasks from the UI.

bomoko:
This PR introduces two new flags to the advanced task definition system
* `adminTask`: This marks a task as requiring slightly elevated privileges within the task system (i.e. having the deployer token mounted into the task container). By default, tasks will not be set as adminTask. Details in #3182 
* `showUi`: if set to 0, the custom task will not be shown in the UI. Details in #3183 

# Checklist

# Closing issues

closes #3183  and closes #3182

### DIFF
--- a/services/api/src/resources/task/helpers.ts
+++ b/services/api/src/resources/task/helpers.ts
@@ -113,6 +113,7 @@ export const Helpers = (sqlClientPool: Pool) => ({
       payload = {},
       remoteId,
       execute,
+      adminTask,
     }: {
       id?: number,
       name: string,
@@ -127,6 +128,7 @@ export const Helpers = (sqlClientPool: Pool) => ({
       payload: object,
       remoteId?: string,
       execute: boolean,
+      adminTask: boolean,
     },
   ) => {
     let rows = await query(
@@ -175,7 +177,9 @@ export const Helpers = (sqlClientPool: Pool) => ({
       environment: environmentData,
       advancedTask: {
         RunnerImage: image,
-        JSONPayload: new Buffer(JSON.stringify(payload).replace(/\\n/g, "\n")).toString('base64')
+        JSONPayload: new Buffer(JSON.stringify(payload).replace(/\\n/g, "\n")).toString('base64'),
+        deployerToken: adminTask, //an admintask will have a deployer token and ssh key injected into it
+        sshKey: adminTask,
       }
     }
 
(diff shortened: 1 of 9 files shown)


services/api/src/resources/task/helpers.ts
```diff
@@ -175,7 +177,9 @@ export const Helpers = (sqlClientPool: Pool) => ({
       environment: environmentData,
       advancedTask: {
         RunnerImage: image,
-        JSONPayload: new Buffer(JSON.stringify(payload).replace(/\\n/g, "\n")).toString('base64')
+        JSONPayload: new Buffer(JSON.stringify(payload).replace(/\\n/g, "\n")).toString('base64'),
+        deployerToken: adminTask, //an admintask will have a deployer token and ssh key injected into it
```
shreddedbacon, line 181: The support for this isn't in any released controllers yet, it has only been merged into main pending a proper release.

shreddedbacon (commented):
I think this is ok

One question though, with `show_ui` I assume this is to hide the admin tasks in the dropdown menu in the UI, but task logs will still show for these tasks. Do we want to hide the logs too?

bomoko:
@shreddedbacon its a good question. I guess if the logs are still available via the api we should be good.

shreddedbacon:
> @shreddedbacon its a good question. I guess if the logs are still available via the api we should be good.

So if I run an admin task that I didn't want the user to be able to execute, it will show up in the list of tasks that have been run against the environment. And the logs for this task will be accessible to all users that have permission to view tasks.
image

Is this ok? Or would there be benefits of hiding that the task had been run, and thus hiding the logs from non-admin users?

bomoko:
@shreddedbacon I've given some more thought to what you were asking - I think, actually, that it would be good if we suppressed the log output from the UI generally for admin tasks - or, at least, made some kind of filter that suppressed admin tasks by default.

I'm thinking here that if there are, say, platform admin tasks that get run (I don't know, say cleaning backups or running analyses on the environment), we wouldn't necessarily want or need that information to be available as _standard_.

shreddedbacon:
Can this get updated and merge conflicts fixed so I can review this properly. I think the time to get this merged is sooner than later.

services/api/src/resources/task/helpers.ts
shreddedbacon: This is now supported in the controller

shreddedbacon:
Also either with or after this is merged, getting the imaged based advanced task permission re-scoped to the same as normal advanced tasks. With these adminTask and showUi features, it means that general users should be able to run imaged based tasks if they need to

tobybellwood:
closing, replaced with new split out PRs

PR #3196, opened 2022-06-21
[DOCS] Updated $BRANCHNAME with actual env variable names $LAGOON_GIT_BRANCH and $LAGOON_GIT_SAFE_BRANCH used in the script.

AlexSkrypnyk:
https://github.com/uselagoon/lagoon-images/blob/main/images/commons/lagoon/entrypoints/50-dotenv.sh#L32

### DIFF
--- a/docs/using-lagoon-advanced/environment-variables.md
+++ b/docs/using-lagoon-advanced/environment-variables.md
@@ -9,12 +9,12 @@ As there can be environment variables defined in either the Dockerfile or during
 1. Environment variables \(defined via Lagoon API\) - environment specific.
 2. Environment variables \(defined via Lagoon API\) - project-wide.
 3. Environment variables defined in Dockerfile \(`ENV` command\).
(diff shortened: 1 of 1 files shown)


shreddedbacon (commented):
Good to go from me. The linked references to the images entrypoint script is how we do it, so the documentation should also represent this.

~One thing we don't have in the docs at the moment is mentioning that points 4-7 are referenced by the entrypoint script, should we maybe put this in there too as a note?~ we do :facepalm:

tobybellwood:
> One thing we don't have in the docs at the moment is mentioning that points 4-7 are referenced by the entrypoint script, should we maybe put this in there too as a note?

that's in https://github.com/uselagoon/lagoon/pull/3196/files#diff-b7a3703dc5a94973227659ffdcf0d9470b46b9f4b69c374ada50aa4d5c276f92R17 no?

shreddedbacon:
> > One thing we don't have in the docs at the moment is mentioning that points 4-7 are referenced by the entrypoint script, should we maybe put this in there too as a note?
> 
> that's in [#3196 (files)](https://github.com/uselagoon/lagoon/pull/3196/files#diff-b7a3703dc5a94973227659ffdcf0d9470b46b9f4b69c374ada50aa4d5c276f92R17) no?

:facepalm:

PR #3197, opened 2022-06-21
Deprecate Drutiny integration

bomoko:
This PR deprecates Drutiny integration from Insights V1.

The Drutiny/problems integration will be revisited in Insights V2 in a more generic fashion.

The next PR will remove it completely.

Addresses #3082


tobybellwood:
@bomoko to recheck and progress

bomoko:
@tobybellwood - this should still work. But at this point it may just make sense to completely remote it

tobybellwood:
@bomoko - should we just close this now insights has moved on, or is it still needed?

PR #3201, opened 2022-06-24
chore: update storage-calculator for kubernetes

### LINKED ISSUE #3200: Update storage-calculator to use kubectl instead of oc
Storage calculator still uses the OC binary, we should stop using this and use kubectl

shreddedbacon:
# Checklist

The storage-calculator still used the OC binary from openshift 3.x, it now uses kubectl.

Other improvements are that the previous version would create the deployment, pause it, modify it with volumes and envvars and then resume the deployment. This process created 2 pods.
This now templates out the deployment prior to creating it so that only 1 pod is created.

# Closing issues

closes #3200


tobybellwood (commented):
@shreddedbacon - I rewrote the dockerfile to remove the dependency on the kubectl image (which we will be removing) - and straightened up the yq/JWT deps

shreddedbacon:
Is the base of this also up to date with the rest of main?

tobybellwood:
This has been running in our test cluster and functions similarly to the previous one. Happy to merge and continue observation

PR #3202, opened 2022-06-24
update drupal used in tests

tobybellwood:
This should remove _some_ of the warnings about composer vulnerabilities, and ensure we're testing what could actually be run.


tobybellwood:
no longer pursuing, will update in different PR

PR #3203, opened 2022-06-27
Feature/add metadata to environment filters

bomoko:
This PR introduces a small extension to the project/environmentsByFactSearch queries, allowing both to search on project metadata.

The queries now support the following

```query environmentbymetadata {
  environmentsByFactSearch(input: {
    filterConnective: AND
    filters: [
      {lhsTarget: PROJECT_METADATA, name: "exampleKey", contains: "exampleValue" }
    ]
  }) {
		environments {
      id
      name
    }
  }
}
```

It introduces a `lhsTarget` of `PROJECT_METADATA` and, as with the other target types, allows one to specify a key and value for the search. The above will match any environments with the metadata key/value pair "exampleKey:exampleValue".
This can, further, be combined with the existing filters.

### DIFF
--- a/services/api/src/resources/fact/resolvers.ts
+++ b/services/api/src/resources/fact/resolvers.ts
@@ -527,7 +527,6 @@ const buildConditionsForFactSearchQuery = (filterDetails: any, factQuery: any, p
           innerBuilder = innerBuilder.orWhere(builderFactory(filter, i));
         }
       });
-      console.log(innerBuilder.toString());
       return innerBuilder;
     })
   }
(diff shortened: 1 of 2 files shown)


services/api/src/resources/fact/resolvers.ts
```diff
@@ -525,6 +527,7 @@ const buildConditionsForFactSearchQuery = (filterDetails: any, factQuery: any, p
           innerBuilder = innerBuilder.orWhere(builderFactory(filter, i));
         }
       });
+      console.log(innerBuilder.toString());
```
timclifford: console log

bomoko: fixed :)

tobybellwood:
@bomoko - should we just close this if we're looking at re-doing metadata, or is it ok to merge now and modify later?

tobybellwood:
Closing pending metadata rewrite

PR #3204, opened 2022-06-29
Feature: new environment variable resolvers

### LINKED ISSUE #1667: build environment variables get escaped incorrectly

### LINKED ISSUE #1677: Allow users to update environment variables value and scope
**Describe the bug**

At the moment with Lagoon, you can create new environment variables. You however can't update the value or scope (e.g. `global` or `runtime`) of the environment variable in the future. If you have the ability to create, surely this would mean you can update as well.

The impact of this bug is that you are forced to:

1. delete the current environment variable
2. create a new environment variable with the same name but with the new value/scope

This however is 2 GraphQL API requests, when 1 would be safer (no messing up the name), atomic (no chance of a build with no environment variable), easier for everyone (1 API request) and less confusing.

This also will mean that customers with heavy use (read - automated) of environment variables will end up hitting the API a lot more than necessary, which could impact scaling in the future.

**Expected behaviour**

If you have the ability to create environment variables, you should also be able to update them. At least the value of the variable and and scope of the variable should be updateable.

**Impacted customers**

At least 2 large customers in APJ are impacted by this.

### LINKED ISSUE #2556: Lagoon API for Environment Variables should allow upsert (insert or update)
The current API for environment variables is sub-optimal, if you are looking to ensure a variable matches a particular value.

## Current process

1. Get the value (for comparison)
2. If the value exists, and is wrong, delete it
3. Insert the new value

This is a rather long process to do, to just set an environment variable.

## Compounding factors

There is a large customer in region that synchronises a lot of environment variables, to ensure they are consistent with a configuration database that they maintain. The process takes around 30 minutes to run end to end.

Not only is this slow, the Lagoon API also experiences more load than is actually necessary.

## Ideal solution

The environment variable add function should insert or update the value, and hide this detail from the end user.

## Why this should not go in Lagoon CLI only

The Lagoon CLI project attempts to keep API parity with Lagoon, so features like this should be upstreamed to Lagoon, so all users benefit from it.

shreddedbacon:
# Checklist

This adds 3 new resolvers to the API for interacting with environment variables
* addOrUpdateEnvVariableByName
* getEnvVariablesByProjectEnvironmentName
* deleteEnvVariableByName

These resolvers are mostly designed to be used by lagoon-cli to make less calls against the API to retrieve this information.

Examples for adding or updating a variable on an environment and a project:
```
# add `environment` to add to environment only
mutation addOrUpdateVariableInEnvironment {
  addOrUpdateEnvVariableByName(
    input: { 
      project: "high-cotton"
      environment:"Develop"
      name: "MY_VARIABLE"
      value: "my variable value"
      scope: GLOBAL 
    }
  ) {
    id
    scope
    name
    value
  }
}
# omit `environment` to add to project only
mutation addOrUpdateVariableInProject {
  addOrUpdateEnvVariableByName(
    input: { 
      project: "high-cotton"
      name: "MY_VARIABLE"
      value: "my variable value"
      scope: GLOBAL
    }
  ) {
    id
    scope
    name
    value
  }
}
```

# Closing issues

closes #2556 
closes #1677


services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -168,7 +168,7 @@ function configure_api_client {
...
+    resource_scope_names=(add add:development add:production addGroup addNoExec addNotification addOrUpdate:development addOrUpdate:production addUser delete delete:development delete:production deleteAll deleteNoExec deploy:development deploy:production drushArchiveDump:development drushArchiveDump:production drushCacheClear:development drushCacheClear:production drushRsync:destination:development drushRsync:destination:production drushRsync:source:development drushRsync:source:production drushSqlDump:development drushSqlDump:production drushSqlSync:destination:development drushSqlSync:destination:production drushSqlSync:source:development drushSqlSync:source:production environment:add:development environment:add:production environment:view:development environment:view:production environment:delete:development environment:delete:production getBySshKey invoke:guest invoke:developer invoke:maintainer create:advanced delete:advanced project:add project:view project:delete removeAll removeGroup removeNotification removeUser ssh:development ssh:production storage update update:development update:production view view:token view:user viewAll viewPrivateKey)
```
rocketeerbkw: Every update function is designed to be run once and only once, and in order. Since functions "later" in the order were made with the assumptions about what was done in previous functions, we can't safely make changes to any update functions once they're released (excepting logic bugs in the update). Since everything is taken care of in `add_delete_env_var_permissions` you can revert the other changes.

rocketeerbkw (dismissed):
If these are "are mostly designed to be used by lagoon-cli" what if put them behind a `cli` field? The purpose would be to wall off a section were we can be looser with BC concerns since we control both sides. I suggested a similar tactic for the UI parts. I'm a bit hesitant to add these new resolvers to the main `query` when we already have issues in the queue that would likely require BC breaks for the proposed schema changes.

shreddedbacon:
> If these are "are mostly designed to be used by lagoon-cli" what if put them behind a `cli` field? The purpose would be to wall off a section were we can be looser with BC concerns since we control both sides. I suggested a similar tactic for the UI parts. I'm a bit hesitant to add these new resolvers to the main `query` when we already have issues in the queue that would likely require BC breaks for the proposed schema changes.

They might be mostly designed for the CLI, but they are easier to use than the existing ones are in terms of user convenience. I flagged the existing ones as deprecated, but not removed them so users that like to do things the hard way still can.

I'm not sure what putting them behind a cli field means here though? Can you reference the ui parts you mentioned?
I'm struggling to see why putting them under something specifically for the cli would make a difference otherwise? I haven't changed any of the existing functionality, these are just new resolvers that store data the same as the old resolvers do. They just make adding, updating, and retrieving this data easier.

rocketeerbkw:
> I'm not sure what putting them behind a cli field means here though? Can you reference the ui parts you mentioned? I'm struggling to see why putting them under something specifically for the cli would make a difference otherwise?

https://github.com/uselagoon/lagoon/pull/3229#discussion_r930446151

So something like:

```graphql
query {
  cli {
    getEnvVariablesByProjectEnvironmentName {
    }
  }
}
```

This schema change only addresses some of the env var issues we have. It's easier to use, but not more flexible. Moving to a sub field is one way to say "hey this section of the api is not public/guaranteed/stable." Or we mark these as deprecated/alpha right now. It's also possible we don't get to anything before v3 and worrying about deprecation doesn't matter. Just wanting to avoid a situation where we have `addEnvVariable @deprecated`, `addOrUpdateEnvVariableByName @deprecated`, `addOrUpdateEnvVariableByEnvironmentType @deprecated`, `addOrUpdateEnvVariableNew` (that's exaggerating to be fair).

services/keycloak/startup-scripts/00-configure-lagoon.sh
rocketeerbkw: Also needed for
https://github.com/uselagoon/lagoon/pull/3204/files#diff-93954129f290faff8e798f2ed40db083f354749196d9e0ed41c30838a0d3d392L185-R185 

and https://github.com/uselagoon/lagoon/pull/3204/files#diff-93954129f290faff8e798f2ed40db083f354749196d9e0ed41c30838a0d3d392R1278-R1314

shreddedbacon:
> > I'm not sure what putting them behind a cli field means here though? Can you reference the ui parts you mentioned? I'm struggling to see why putting them under something specifically for the cli would make a difference otherwise?
> 
> [#3229 (comment)](https://github.com/uselagoon/lagoon/pull/3229#discussion_r930446151)
> 
> So something like:
> 
> ```graphql
> query {
>   cli {
>     getEnvVariablesByProjectEnvironmentName {
>     }
>   }
> }
> ```
> 
> This schema change only addresses some of the env var issues we have. It's easier to use, but not more flexible. Moving to a sub field is one way to say "hey this section of the api is not public/guaranteed/stable." Or we mark these as deprecated/alpha right now. It's also possible we don't get to anything before v3 and worrying about deprecation doesn't matter. Just wanting to avoid a situation where we have `addEnvVariable @deprecated`, `addOrUpdateEnvVariableByName @deprecated`, `addOrUpdateEnvVariableByEnvironmentType @deprecated`, `addOrUpdateEnvVariableNew` (that's exaggerating to be fair).

Are there more than these issues that I can find?
https://github.com/uselagoon/lagoon/issues/1538 - IMHO this is a bad idea, would work better under my proposed organization structure.
https://github.com/uselagoon/lagoon/issues/2556 - This PR addresses this issue, and makes it easy
https://github.com/uselagoon/lagoon/issues/1677 - This PR addresses this issue, and makes it easy
https://github.com/uselagoon/lagoon/issues/3208 - Part of this was a bug, but unsetting variables could be useful, these resolvers or any others that are existing can't solve this directly either.
https://github.com/uselagoon/lagoon/issues/3286 - This could be solved by using new enums with the env type in the scope `global_development` or such, if not with new scopes, I'm sure these resolvers could be extended easily to support environment types..

If there are other issues, I'd love to know about them to see how these resolvers need to be more flexible to support potential changes?

tobybellwood:
I don't see these as being purely-CLI (or even purely-UI, as they would be used there too) - hence no need to put them in a sub-section.

I think it's ok to add them now (given we're not rewriting existing functions - they will act as they currently do, but these ones will bring Lagoon closer to the CRUD tools (in this space anyway) that our users expect it to be?

grahamethompson:
As a consumer it would be more convenient to be able to **update** an env var than the process to **get / delete / add**. 

Our use case currently is secrets rotations.

services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -1781,6 +1782,57 @@ function update_add_env_var_to_project {
...
+
+  /opt/jboss/keycloak/bin/kcadm.sh create clients/$CLIENT_ID/authz/resource-server/permission/scope --config $CONFIG_PATH -r lagoon -f - <<EOF
+{
+  "name": "Delete Environment Variable from Project",
+  "type": "scope",
+  "logic": "POSITIVE",
+  "decisionStrategy": "UNANIMOUS",
+  "resources": ["env_var"],
+  "scopes": ["project:delete"],
+  "policies": ["Users role for project is Maintainer","User has access to project"]
```
tobybellwood: As per #3155 |
```suggestion
  "policies": ["[Lagoon] Users role for project is Maintainer","[Lagoon] User has access to project"]
```

tobybellwood: as per #3155

services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -1781,6 +1782,57 @@ function update_add_env_var_to_project {
...
+
+    /opt/jboss/keycloak/bin/kcadm.sh create clients/$CLIENT_ID/authz/resource-server/permission/scope --config $CONFIG_PATH -r lagoon -f - <<EOF
+{
+  "name": "Delete Environment Variable from Development Environment",
+  "type": "scope",
+  "logic": "POSITIVE",
+  "decisionStrategy": "UNANIMOUS",
+  "resources": ["env_var"],
+  "scopes": ["environment:delete:development"],
+  "policies": ["Users role for project is Developer","User has access to project"]
```
tobybellwood: As per #3155 

```suggestion
  "policies": ["[Lagoon] Users role for project is Developer","[Lagoon] User has access to project"]
```

services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -1781,6 +1782,57 @@ function update_add_env_var_to_project {
...
+
+    /opt/jboss/keycloak/bin/kcadm.sh create clients/$CLIENT_ID/authz/resource-server/permission/scope --config $CONFIG_PATH -r lagoon -f - <<EOF
+{
+  "name": "Delete Environment Variable from Production Environment",
+  "type": "scope",
+  "logic": "POSITIVE",
+  "decisionStrategy": "UNANIMOUS",
+  "resources": ["env_var"],
+  "scopes": ["environment:delete:production"],
+  "policies": ["Users role for project is Maintainer","User has access to project"]
```
tobybellwood: As per #3155 
```suggestion
  "policies": ["[Lagoon] Users role for project is Maintainer","[Lagoon] User has access to project"]
```

tobybellwood (approved):
tested ok, issues resolved

PR #3221, opened 2022-07-21
Feature: basic and worker template types

### LINKED ISSUE #2525: Create new lagoon type "worker" for long operating processes
Currently, Lagoon projects that have long running (non-web) processes use the `cli` Lagoon type to run these processes in.  This is not optimal, as the cli type should be reserved for short-running, instant spin-up type uses.

We are proposing to add an additional type `worker` and corresponding `worker-persistent` type to Lagoon.  The main aim of these types is to accommodate projects that have long-running, persistent, and non-web-facing processes.  Where a project contains a worker type and needs cronjobs - these cronjobs would be run inside the worker pod instead of a separate cron pod.

The main benefit of this is that cli pods can be idled when they are not in use, reducing load on a cluster, and can be treated appropriately during scaling or node-shuffling events.  This will also mean that projects can operate their web-facing containers as lean as possible, offloading backend processing or interactive tasks to an additional pod, with specialist tooling.

shreddedbacon:
# Checklist

Previously we have recommended people use the `node` or `cli` templates when building and deploying more generic or basic applications. Instead of suggesting these, adding two new types:

* `basic` and `basic-persistent` - almost identical to `node` but allows for the service port to be changed using `lagoon.service.port` label on the docker-compose service. This is good for other languages like ruby, or go.
* `worker` and `worker-persistent` - almost identical to `cli` but does not mount the project ssh key. This type is good for when you just need to process data without exposing a service or ingress.

# Closing issues

closes #2525


tobybellwood:
How will aergia handle these new types? - https://github.com/amazeeio/aergia-controller/blob/main/resources/selectors.yaml

Presumably, we don't want worker pods idling? There's no way we can set a label on a deployment to stop it idling at a namespace level though?

shreddedbacon:
> How will aergia handle these new types? - [https://github.com/amazeeio/aergia-controller/blob/main/resources/selectors.yaml](https://github.com/amazeeio/aergia-controller/blob/main/resources/selectors.yaml?rgh-link-date=2022-07-28T00%3A56%3A31Z)
> 
> Presumably, we don't want worker pods idling? There's no way we can set a label on a deployment to stop it idling at a namespace level though?

Aergia will just idle them like any other pod, the selectors work on these labels which mostly exclude certain types from idling: https://github.com/amazeeio/aergia-controller/blob/main/resources/selectors.yaml#L1-L21

If someone did want to exclude these, the selectors are configurable in aergia via the helm chart, so someone could create their own selectors for idling rather than use our pre-defined one

Why would you want to keep a worker pod around and idle the others though? If the worker needs to work on something another pod has, it would probably crash (if built right). I see no problems with idling these like other pods. If a worker pod does need to not be idled, then there is the option to just exclude the environment from idling entirely too.

tobybellwood:
I think as long as these pods idle the same way as (non-cli) services, that's perfect - and reading the selectors confirms that!

PR #3229, opened 2022-07-26
WIP: UI Get Projects resolver

timclifford:
This will speed up the retrieval of basic project data in the UI by replacing the expensive 'allProjects' query. The intention is to use this resolver to return all the projects the user has access to (currently only returns name and id) directly from the access token since a new 'projects' attribute has been defined.


services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -52,6 +52,9 @@ function configure_lagoon_realm {
...
 
+    echo Creating projectIds mapper for lagoon-ui "projects"
+    echo '{"protocol":"openid-connect","config":{"script":"var ArrayList = Java.type(\"java.util.ArrayList\");\nvar projects = new ArrayList();\nvar forEach = Array.prototype.forEach;\n\nforEach.call(user.getGroups().toArray(), function(group) {\n if(group.getFirstAttribute(\"type\") == \"role-subgroup\") {\n var parent = group.getParent();\n if(parent.getFirstAttribute(\"type\") == \"project-default-group\") {\n var projectIds = parent.getFirstAttribute(\"lagoon-projects\");\n if(projectIds !== null) {\n forEach.call(projectIds.split(\",\"), function(g) {\n projects.add(g); \n });\n return;\n }\n }\n }\n \n return;\n});\n\nexports = projects;","id.token.claim":"true","access.token.claim":"true","userinfo.token.claim":"true","multivalued":"true","claim.name":"projects","jsonType.label":"String"},"name":"projects","protocolMapper":"oidc-script-based-protocol-mapper"}' | /opt/jboss/keycloak/bin/kcadm.sh create -r ${KEYCLOAK_REALM:-master} clients/$CLIENT_ID/protocol-mappers/models --config $CONFIG_PATH -f -
```
rocketeerbkw: This file is designed to be run as an init/update style (each function runs once, and in order). So each function, once released, should be considered "final" and not changed (misspellings aside).

The new mapper will need to be part of a new function (added at the end of the list) that should check if it has been run already.

services/api/src/typeDefs.js
```diff
@@ -1154,6 +1154,10 @@ const typeDefs = gql`
     """
     deployTargetConfigsByDeployTarget(deployTarget: Int!) : [DeployTargetConfig]  @deprecated(reason: "Unstable API, subject to breaking changes in any release. Use at your own risk")
     allDeployTargetConfigs: [DeployTargetConfig]  @deprecated(reason: "Unstable API, subject to breaking changes in any release. Use at your own risk")
+    """
+    Returns all user Projects for associated user
+    """
+    uiProjects(limit: Int, skip: Int): [Project]
```
rocketeerbkw: I expect we'll have even more UI specific queries and mutations, should we put them under a new `ui` field now so we don't have to bc break anything later?

```graphql
{
  ui {
    projects(limit: 10, skip: 10)
  }
}
```

rocketeerbkw (changes requested):
I don't know much about the protocol mappers, so I have some questions about this method.

How often are they run? It's included in the token itself, so that means it only runs once on token creation? If it's run more often (like for every page load/query) how fast is it? Executing JS in Java isn't very performant. Our production tokens are valid for 10 hours, so any user<>project access changes won't be reflected until after 10 hours (the users token expired) or until the user manually logs out/logs in?

timclifford:
Thanks @rocketeerbkw - great points.

**How often will the projects mapper be run?**
- Every time the page is refreshed since a new access token is generated
- Or every 5mins without a page refresh since tokens are now refreshed server-side

This means that updating user projects on keycloak will only reflect in the ui when a new token is requested.

**How fast is it?**
- dashboard.amazeeio.cloud/projects - retrieves jwt in 40ms
- localhost (test with project mapper) - retrieves jwt in `150-200ms (has around 40 project ids attached)

So its around 4x-5x slower having to compute JS in Java. We could re-write the mapper in Java? 


**10h Production tokens**
Our production tokens are valid for 10 hours but after updating user attribute fields such as name, these changes were instantly reflected on a token refresh.

I also tested adding a test 'User Realm Role' protocol mapper to the 'lagoon-ui' client and refreshed the token on the UI which showed the update. So I think a token refresh will always retrieve the latest attribute on request rather than needing to re-authenticate or wait for expiry. With SSR now updating tokens every 5mins a user should only ever have outdated project id data for a maximum of 5minutes. 


**Other**
Also need to ensure we move the configuration logic for the new mapper to the end of the file, in a new function.

smlx:
Hey @timclifford FYI there is already a client which uses a built-in mapper to achieve what you're trying to do here - no need to inject JS or Java dynamic code. It was added in https://github.com/uselagoon/lagoon/pull/2941. Happy to show you how it works if you like :slightly_smiling_face:

timclifford:
> Hey @timclifford FYI there is already a client which uses a built-in mapper to achieve what you're trying to do here - no need to inject JS or Java dynamic code. It was added in #2941. Happy to show you how it works if you like 🙂

Thanks for showing me this, I've now updated the mapper to use the `group lagoon project IDs mapper`.

tobybellwood:
superseded by #3397

PR #3233, opened 2022-07-29
fix: add custom dnsConfig to all cronjobs

smlx:
# Checklist

This is an attempt to fix the problem where cronjobs intermittently fail
with getaddrinfo() error EAI_AGAIN. The problem appears to be that the
lookup times out due to dropped UDP packets.

The default for timeout is 5 seconds and attempts is 2 for both musl and
glibc. This change bumps the configured values to the maximum values for
musl: 60s timeout and 10 attempts.

For any custom cronjobs running glibc-based images the maximum values
are silently capped at half those values: 30s timeout and 5 attempts.

References:
https://git.musl-libc.org/cgit/musl/tree/src/network/resolvconf.c#n17
https://manpages.debian.org/buster/manpages/resolv.conf.5.en.html

# Closing issues

There is no issue for this - we have seen the problem on managed clusters.

### DIFF
--- a/images/kubectl-build-deploy-dind/helmcharts/basic-persistent/templates/cronjob.yaml
+++ b/images/kubectl-build-deploy-dind/helmcharts/basic-persistent/templates/cronjob.yaml
@@ -76,4 +76,10 @@ spec:
           tolerations:
             {{- toYaml $ | nindent 12 }}
         {{- end }}
+          dnsConfig:
+            options:
+              - name: timeout
+                value: "60"
+              - name: attempts
+                value: "10"
 {{- end }}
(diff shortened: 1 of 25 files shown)


smlx:
Follow-up will be to add the same configuration to our other templates. The same issue has been seen there - just very rarely.

shreddedbacon (approved):
All good from me

smlx:
Just a note because there was a question about it: the *total timeout is 60s*. Each attempt is retried at 6 second intervals. It is **not** 10 attempts x 60s == 10 minutes.

See https://git.musl-libc.org/cgit/musl/tree/src/network/res_msend.c#n111 for the logic.

tobybellwood (approved):

smlx:
Thinking about this some more, should we shorten the timeout to 30s? Keep the 10 attempts so the period is 3s, which is closer to the default period?

shreddedbacon:
> Thinking about this some more, should we shorten the timeout to 30s? Keep the 10 attempts so the period is 3s, which is closer to the default period?

I'm not fussed on this, we can always adjust it or make it configurable in some way in the future when it is added to `build-deploy-tool`

PR #3241, opened 2022-08-03
Additional keycloak permissions scripts/hook

### LINKED ISSUE #3240: Introduce hook mechanism for Keycloak customization
If a user wants to customize their Keycloak permissions, they have few options - the only real one being using the UI.

It would be useful if these changes could be persisted as code in some way, such as a configMap that's volume mounted into the Keycloak container.

bomoko:
This POC introduces a simple mechanism for adding final steps to the keycloak service's init process.

It simply searches in `/opt/jboss/startup-scripts/extra/` for any `.sh` files and sources them.

The example of how this can be used is given in `extra_permission_example.sh.example` which updates the default "Add Project" permission.

# Closing issues

closes #3240


rocketeerbkw (commented):
Since `/opt/jboss/startup-scripts` is "owned" by keycloak, if we went with `configure-lagoon-extra` or something similarly lagoon specific that would make me feel better about not accidentally stepping on future keycloak upgrades.

Otherwise LGTM

bomoko:
@rocketeerbkw Thanks for the feedback, makes sense - so I've moved the script out of the directory completely (although I'm not sure if you just meant I should just rename it).

Let me know if this works for you!

services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -1811,6 +1811,15 @@ function configure_keycloak {
     configure_token_exchange
     update_add_env_var_to_project
 
+    # Here we source in any extra configuration
+    echo "Checking for any custom startup scripts"
+    if ls /configure-lagoon-extra/*.sh 1> /dev/null 2>&1; then
```
smlx: Instead of this logic you could do something like this that might be simpler?
```bash
     for FILE in /configure-lagoon-extra/*.sh; do
        [ -r "$FILE" ] || continue
        echo "Found $FILE - executing in current shell"
        source "$FILE";
      done
```

services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -1811,6 +1811,15 @@ function configure_keycloak {
...
+      for FILE in /configure-lagoon-extra/*.sh; do
+        echo "Found $FILE - executing in current shell"
+        source $FILE;
```
smlx: This should be quoted or it will break if someone puts a space in the name of their script.

services/keycloak/startup-scripts/00-configure-lagoon.sh
bomoko: Thanks @smlx!

rocketeerbkw (approved):

tobybellwood:
May also be worth looking at https://github.com/adorsys/keycloak-config-cli/ to see if it can be used.

tobybellwood:
Not planned to implement this particular implementation

PR #3244, opened 2022-08-08
Feature: Configure remote target build image in API

shreddedbacon:
# Checklist

This adds support for being able to define the build image to use on a particular remote in the API.
This allows for a specific target to use a different image without having to apply a change directly to the remote-controller in that target.

This image will override whatever the remote-controller has defined.

Two environment variables are introduced in this
* `DEFAULT_BUILD_DEPLOY_IMAGE`
  * This will define the default image to use in lagoon, this will by default be set in the `lagoon-core` chart matching the core version
* `EDGE_BUILD_DEPLOY_IMAGE`
  * This will define an edge image to use, this will override the default image and will be a more frequently released version of the build image but still constrained to a supported core version. The idea with edge is that it is a helm option that will set this to an edge image version in `lagoon-core` that is also matched to the core version, but this image tag will receive more frequent updates for the supported version of core, where the default image will not.

Eventually the image that is defined in the `remote-controller` helm chart will only be used as a fall back if no other image determined by core is applied. The goal being that core will be the source of image truth.


shreddedbacon:
Add the default environment variable image configuration

shreddedbacon:
@tobybellwood I added the support for the default and edge stuff to this.

Of note, if you don't set either of the vars, and you don't configure any remotes in the API, then the fallback (current) behaviour is to use what is in remote. And this will always be the case.

So once we get the support for configuring those vars as part of `lagoon-core` chart, it won't matter what people have configured in their remotes as the API will always win.

PR #3245, opened 2022-08-09
Go svc updates to 1.18

CGoodwin90:
Updates to the Go Services - Active Standby, Actions Handler, Backup Handler, to go 1.18 and common structure.


services/actions-handler/Dockerfile
```diff
@@ -2,7 +2,7 @@
 ARG GO_VERSION
 ARG UPSTREAM_REPO
 ARG UPSTREAM_TAG
-FROM golang:${GO_VERSION:-1.16.4} AS builder
+FROM golang:${GO_VERSION:-1.18} AS builder
```
tobybellwood: I think we're ok to remove the variable here (and the ARG in line 2), and just call it like
`FROM golang:1.18-alpine AS builder`

services/backup-handler/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # build the binary
 ARG UPSTREAM_REPO
 ARG UPSTREAM_TAG
-FROM golang:${GO_VERSION:-1.16.4} AS builder
+FROM golang:${GO_VERSION:-1.18} AS builder
```
tobybellwood: same here - I think we're ok to remove the variable here, and just call it like
`FROM golang:1.18-alpine AS builder`

services/workflows/Dockerfile
```diff
@@ -2,7 +2,7 @@
 ARG GO_VERSION
 ARG UPSTREAM_REPO
 ARG UPSTREAM_TAG
-FROM golang:${GO_VERSION:-1.16.4} AS builder
+FROM golang:${GO_VERSION:-1.18} AS builder
```
tobybellwood: I think we're ok to remove the variable here (and the ARG in line 1), and just call it like
`FROM golang:1.18-alpine AS builder`

taskimages/activestandby/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the binary
```
tobybellwood, line 1: it may also be worth looking at the structure of the builder stage in this dockerfile - and make it match the others - just purely a neatness thing?

services/backup-handler/Dockerfile
```diff
@@ -1,7 +1,7 @@
...
 
 # bring in all the packages
 # COPY main.go /go/src/github.com/uselagoon/lagoon/services/backup-handler/
```
tobybellwood: Let's also clean out the commented COPY lines - we don't need them

services/actions-handler/go.mod
```diff
@@ -1,17 +1,26 @@
...
 	github.com/matryer/try v0.0.0-20161228173917-9ac251b645a2 // indirect
+	github.com/pborman/uuid v1.2.1 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
+	github.com/rabbitmq/amqp091-go v1.4.0 // indirect
+	github.com/streadway/amqp v1.0.0 // indirect
 	github.com/tiago4orion/conjure v0.0.0-20150908101743-93cb30b9d218 // indirect
-	gopkg.in/matryer/try.v1 v1.0.0-20150601225556-312d2599e12e
 )
 
 // Fixes for AppID
```
tobybellwood, line 26: we'll have to check whether these AppID fixes have applied, given the replace is for a different version

tobybellwood (changes requested):
looks pretty good - couple of things to look into

services/actions-handler/go.mod
shreddedbacon: they aren't :D we should maybe just stop using the appid stuff, worth investigating as part of `logs2notifications` more than solving here

services/actions-handler/Dockerfile
CGoodwin90: Updated to `FROM golang:1.18-alpine3.16 AS builder`

services/backup-handler/Dockerfile
CGoodwin90: Updated to `FROM golang:1.18-alpine3.16 AS builder`

services/workflows/Dockerfile
CGoodwin90: Updated to `FROM golang:1.18-alpine3.16 AS builder`

taskimages/activestandby/Dockerfile
CGoodwin90: Updated the structure to match the other Dockerfiles

services/backup-handler/Dockerfile
CGoodwin90: Cleaned up the comments

tobybellwood (approved):

PR #3251, opened 2022-08-12
refactor: feature flag backup configurations

### LINKED ISSUE #3250: Custom backup-retention is "overwritten" by other enviornments retention periods
# Describe the bug
Lagoon uses a shared k8up/restic S3 repo for multiple namespaces, this improves the de-duplication capabilities of restic and greatly reduces the amount of backup space that is used.

With the `backup-retention.production` setting in the `.lagoon.yml` a developer can set a custom retention period for production environments, see https://docs.lagoon.sh/using-lagoon-the-basics/lagoon-yml/#backup-retention 

Unfortunately the restic prune system does not just prune for one specific environment, but prunes the whole repo, this means a prune command that runs in a development environment still has the normal backup retention used and will therefore remove snapshots of production environments.

In order to fix this we should start using the restic tag system that can be defined via `Schedule` objects (https://k8up.io/k8up/2.3/references/api-reference.html#k8s-api-github-com-k8up-io-k8up-v2-api-v1-backupspec) basically it will tag snapshots with tags and when then in the prune retention setting we can define the same tag: https://k8up.io/k8up/2.3/references/api-reference.html#k8s-api-github-com-k8up-io-k8up-v2-api-v1-retentionpolicy which means the prune will only prune backups from the same tag and the issue should not happen anymore.

Another way to mitigate this is to have a custom S3 backup location defined for the production environment as then the development environment prune task cannot access the production backups anymore and cannot prune them.
This would need a change though in the `backup-generation.sh` (https://github.com/uselagoon/lagoon/blob/acede1cb7929bea9281b2fbdf7d9829ca6e22bbf/images/kubectl-build-deploy-dind/scripts/exec-backup-generation.sh#L142) script to not only look for `LAGOON_PROJECT_VARIABLES` but also for `LAGOON_ENVIRONMENT_VARIABLES` in order to be able to set custom S3 backups locations for the production environment

## To Reproduce

1. create `.lagoon.yml` with:
```
backup-retention:
  production:
    daily: 999
```

2. deploy a production and development environment
3. wait for 3 weeks
4. see that daily backups for production environment have been removed after 7 days (the default retention period for daily backups)

## Expected behavior

production environment backups are kept for 999 days

shreddedbacon:
# Checklist

This is to featureflag wrap experimental backup configurations. If you think this is useful, ensure you speak to the Lagoon team before you enable this to find out if it will work for your specific environments, as there is potential for issues with backups if incorrectly configured.

# Closing issues

closes #3250

### DIFF
--- a/images/kubectl-build-deploy-dind/scripts/exec-backup-generation.sh
+++ b/images/kubectl-build-deploy-dind/scripts/exec-backup-generation.sh
@@ -4,6 +4,9 @@ set +x
 ##############################################
 #   it is possible to override the retention using a variable defined in the api
 #
+#   if you want to use a different retention period for production branches, you can use the following
+#   LAGOON_BACKUP_PROD_RETENTION="H:D:W:M"
+
 #   if you want to use a different retention period for development branches, you can use the following
 #   LAGOON_BACKUP_DEV_RETENTION="H:D:W:M"
 #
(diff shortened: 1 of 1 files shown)


shreddedbacon:
This is experimental because it relies on newer versions of k8up to work properly, which is not fully supported in Lagoon yet. Support is being added while we work on supporting newer versions of k8up, which is why this is feature flagged.

tobybellwood (approved):
Tested backwards and forwards - works as expected

PR #3252, opened 2022-08-12
fix: use restore bucket protocol from the restore url

shreddedbacon:
# Checklist

When working locally if a restore is in a local minio for example, the minio may not have https on it. As https is hardcoded, the restore fails to work, instead use the protocol from the restore itself.


tobybellwood (approved):
tested local and in test cluster

PR #3254, opened 2022-08-15
chore: use bridge for s3_files_host for local

shreddedbacon:
# Checklist

When using `make ui-logs-development` locally, the default value for `S3_FILES_HOST` is localhost which doesn't resolve to the minio instance correctly. This fixes it to use the docker bridge network like we do for registry etc.


shreddedbacon:
Pretty simple one, only impacts `make ui-logs-development` locally

PR #3256, opened 2022-08-16
Inital pass at disallowing invalid characters via project search

### LINKED ISSUE #3255: Add a trim to the project search feature
# Describe the bug

If you copy and paste whitespace before or after the project name, then the project filter will find no projects 

image

This is annoying.

## To Reproduce

1. Go to the projects page
2. Type ` project`
3. Notice how it finds no project

## Expected behavior

Seeing as whitespace is not a valid project identifier, I would expect it to be stripped before searching

CGoodwin90:
Relates to issue [3255](https://github.com/uselagoon/lagoon-ui/issues/49) - updated input field to disallow invalid characters.
# Checklist

# Closing issues

closes uselagoon/lagoon-ui#49

### DIFF
--- a/services/ui/src/components/Projects/index.js
+++ b/services/ui/src/components/Projects/index.js
@@ -52,7 +52,7 @@ const Projects = ({ projects = [] }) => {
           className="searchInput"
           type="text"
           value={searchInput}
-          onChange={e => setSearchInput(e.target.value.replace(/[^0-9a-zA-Z-]/g,''))}
+          onChange={e => setSearchInput(e.target.value.replace(/[^0-9a-zA-Z.-]/g,''))}
           placeholder="Type to search"
           disabled={projects.length === 0}
         />
--- a/services/ui/src/components/Projects/index.js
+++ b/services/ui/src/components/Projects/index.js
@@ -28,13 +28,13 @@ const Projects = ({ projects = [] }) => {
   const filteredProjects = projects.filter(key => {
     const sortByName = key.name
       .toLowerCase()
-      .includes(searchInput.toLowerCase());
+      .includes(searchInput.trim().toLowerCase());
     let sortByUrl = '';
     if (key.environments[0] !== void 0) {
       if (key.environments[0].route !== null) {
         sortByUrl = key.environments[0].route
           .toLowerCase()
-          .includes(searchInput.toLowerCase());
+          .includes(searchInput.trim().toLowerCase());
       }
     }
     return ['name', 'environments', '__typename'].includes(key)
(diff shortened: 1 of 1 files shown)


services/ui/src/components/Projects/index.js
```diff
@@ -52,7 +52,7 @@ const Projects = ({ projects = [] }) => {
           className="searchInput"
           type="text"
           value={searchInput}
-          onChange={e => setSearchInput(e.target.value)}
+          onChange={e => setSearchInput(e.target.value.replace(/[^0-9a-zA-Z-]/g,''))}
```
tobybellwood:
```suggestion
          onChange={e => setSearchInput(e.target.value.replace(/[^0-9a-zA-Z.-]/g,''))}
```
There may be people who use the search to search for their route URLs, so adding the period separator may be prudent here.

smlx: The hyphen needs to be last (or first) otherwise it's interpreted as a range, so maybe: `/[^0-9a-zA-z.-]/g`.

CGoodwin90: Great idea - updated now.

services/ui/src/components/Projects/index.js
```diff
@@ -52,7 +52,7 @@ const Projects = ({ projects = [] }) => {
           className="searchInput"
           type="text"
           value={searchInput}
-          onChange={e => setSearchInput(e.target.value)}
+          onChange={e => setSearchInput(e.target.value.replace(/[^0-9a-zA-Z.-]/g,''))}
```
rocketeerbkw: Real-time frontend frameworks like React allow us to do things that weren't really possible before and this is a clever and easy technical solution to the problem, which is why I've tried it on past projects. Unfortunately it results in a bad user experience for form input.

This implementation creates a militant type of form validation where invalid user input is ignored. But to the user whet they see is that they'll type an `_` but nothing on the page changes. The way 99% of websites do form validation is by telling the user what they did wrong, and many of the HTML5 client-side validations still allow invalid input. To the user, it will seem like the site is broken, not that we are validating their input (during UAT we had one user ask for a spare keyboard because they thought a key was broken).

IMHO the bare minimum to close the original issue would be to add a new `onPaste` handler that trims whitespace only on paste.

The current form doesn't use any form validation at all, we just depend on the "no results" message. If we wanted to change that we could implement full/proper form validation by showing error messages and error styling with the underlying principle that user input must always be kept (including if the user types a `space`).

rocketeerbkw (changes requested):

services/ui/src/components/Projects/index.js
CGoodwin90: Really solid points, makes sense! I've briefly discussed this with @bomoko and instead of mutating the input value, we're now trimming the whitespace in the filtering step. This won't restrict user input like the previous iteration, but will ignore invalid whitespace.

rocketeerbkw (approved):

services/ui/src/components/Projects/index.js
rocketeerbkw: Thanks, LGTM!

bomoko (approved):

tobybellwood:
@CGoodwin90 this should be ported to uselagoon/lagoon-ui

PR #3259, opened 2022-08-17
chore: remove remaining harbor integration from core

shreddedbacon:
# Checklist

As #2907 outlines, we need to remove harbor support from core as this integration is no longer required.
There is already some work in #3086 to remove other harbor integrations

This is an actual breaking change, but as outlined in Lagoon v2.18.0, harbor support in core would be removed in a future release.

This pullrequest also completely removes the harbor integration and all API queries and mutations that were changed to return errors in 2.18.0 as part of #3672. Anyone that was still using any of these would have been getting errors since v2.18.0.

# Closing issues

partially #2907


tobybellwood:
@shreddedbacon to flag this for deprecation in the next release - and then this PR will remove it in a future release.

tobybellwood (approved):
This looks good to me. Can't see any remnant code in there and all checks out.

PR #3260, opened 2022-08-17
Feature: add lagoon-core version to build payload

### LINKED ISSUE #3257: Send lagoon version in build and task messages
When a build or task is created, we should send the lagoon-core version in the build or task payloads so that the images that run the builds or tasks are aware of which version of lagoon-core triggered it.

This way we can build compatibility into builds and tasks.

shreddedbacon:
# Checklist

As #3257 describes, sending the lagoon-core version in builds could be very useful for builds to be able to consume.
This PR uses a scope of `internal_system` to inject `LAGOON_SYSTEM_CORE_VERSION` which is not able to be defined in the API by users, it is only used by internal systems and is a simple way to pass variables/values from core to builds without having to change CRDs or other resources. Builds are be able to consume these without fear that a user will have changed them.

# Closing issues

closes #3257


shreddedbacon:
https://github.com/uselagoon/build-deploy-tool/issues/85 is to track consuming this in the build-deploy-tool

bomoko (approved):
Looks good, makes sense.

Issue #3262, opened 2022-08-19
disable opensearch integration if no opensearch defined


shreddedbacon:
When running locally or in tests with no opensearch configured, the API still attempts to reach out to an opensearch (that doesn't exist) to create roles and index patterns etc.

It would be nice to be able to disable this integration if there is no opensearch configured.

smlx:
Is there any update on this issue?

shreddedbacon:
https://github.com/uselagoon/lagoon/blob/f519a90cb4b2ec28ec1f6c7283ba4a24951e7e60/services/api/src/resources/group/opendistroSecurity.ts#L16 there is this envvar that can disable the integration

PR #3263, opened 2022-08-22
Revert "chore: remove random wait from test suite"

tobybellwood:
This reverts commit 56a3bd5549f1b72847f0ff401e06ea40bf757a16.

api-db lockouts are being observed in CI when multiple projects process in the same second. Not replicatable in a prod env.

### DIFF
--- a/tests/ansible.cfg
+++ b/tests/ansible.cfg
@@ -8,3 +8,4 @@ callback_whitelist = profile_tasks
 
 [persistent_connection]
 command_timeout = 60
+connect_timeout = 60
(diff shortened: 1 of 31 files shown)


shreddedbacon:
I don't think this is api-db related, but api. It looks like when a number of `addProject` requests are made at the same time, because we have no harbor or opensearch configured in core, there are 2 operations that are attempted, and because these requests appear to be blocking or delaying subsequent requests long enough to hit the 30s ansible request timeout. If this 30s timeout was increase (to who knows, 60/90s?) this may be enough to avoid the failures.

The best option though is to just remove the harbor integration from core (https://github.com/uselagoon/lagoon/issues/2907) and feature wrap the opensearch integration (https://github.com/uselagoon/lagoon/issues/3262) so that these timeouts and stuff are not required.

shreddedbacon:
Seems even the random wait hits the issue too :\

tobybellwood:
superseded by https://github.com/uselagoon/lagoon/pull/3264 - which adds retry logic instead of a hopeful wait

PR #3264, opened 2022-08-23
move test api retries into api stage

tobybellwood:
previously, if the api encountered a delay or an error, it wouldn't attempt a retry, as the retries/delay code was added to the debug stage. This PR moves the logic into the correct stage, and introduces a slight jitter into the delay to avoid more potential race conditions.

### DIFF
--- a/tests/tasks/api/add-deploytarget.yaml
+++ b/tests/tasks/api/add-deploytarget.yaml
@@ -26,8 +26,9 @@
         body:
           query: '{{ lookup("template", "./add-deploytarget.gql") }}'
       register: apiresponse
+      until: apiresponse.status == 200
+      retries: 10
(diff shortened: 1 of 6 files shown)


shreddedbacon:
Thinking about this more, I'm not sure if retries are the best thing to do here. If the first request processes but times out after 30s, the second request that is tried could return an error still, but errors are still `200` responses?
Which means `until: apiresponse.status == 200` will just waltz on through even on genuine failures, since we aren't checking the apiresponse for an error condition?

It might be better to set the timeout parameter in the URI request from 30 to longer, and then checking the response for errors (but this is also not ideal as some things error in a successful request)https://docs.ansible.com/ansible/latest/collections/ansible/builtin/uri_module.html#parameter-timeout

tobybellwood:
no longer required

PR #3268, opened 2022-08-23
Makes the opensearch api integration feature flaggable

bomoko:
This PR makes the opensearch api integration for the lagoon api service flaggable via an environment variable OPENSEARCH_INTEGRATION_ENABLED

if set to "false" it returns a mock object that simply logs the calls to the functions used for opensearch integration - this prevents us from getting integration errors in the logs, as well as preventing startup slowness in some circumstances (such as local docker-compose development and in testing).

# Checklist

### DIFF
--- a/services/api/src/resources/group/opendistroSecurity.ts
+++ b/services/api/src/resources/group/opendistroSecurity.ts
@@ -8,11 +8,42 @@ import { Helpers as projectHelpers } from '../project/helpers';
 export const OpendistroSecurityOperations = (
   sqlClientPool: Pool,
   GroupModel
-) => ({
-  syncGroup: async function(groupName, groupProjectIDs) {
-    return this.syncGroupWithSpecificTenant(groupName, groupName, groupProjectIDs);
-  },
-  syncGroupWithSpecificTenant: async (groupName, tenantName, groupProjectIDs) => {
+) => {
+
+  // In certain cases (some testing, local env), we want to disable the opensearch integration
+  // Setting OPENSEARCH_INTEGRATION_ENABLED to "false" will return an object that simply logs
+  // the requested operations, rather than performs them
+  const opensearchIntegrationEnabled = getConfigFromEnv('OPENSEARCH_INTEGRATION_ENABLED', 'true');
+
+  if (opensearchIntegrationEnabled == "false") {
+    logger.info("OPENSEARCH_INTEGRATION_ENABLED is 'false' - disabling and returning log only object");
+    return {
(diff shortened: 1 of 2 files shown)


services/api/src/resources/group/opendistroSecurity.ts
```diff
@@ -8,11 +8,42 @@ import { Helpers as projectHelpers } from '../project/helpers';
...
-  syncGroupWithSpecificTenant: async (groupName, tenantName, groupProjectIDs) => {
+) => {
+
+  // In certain cases (some testing, local env), we want to disable the opensearch integration
+  // Setting OPENSEARCH_INTEGRATION_ENABLED to "false" will return an object that simply logs
+  // the requested operations, rather than performs them
+  const opensearchIntegrationEnabled = getConfigFromEnv('OPENSEARCH_INTEGRATION_ENABLED', 'true');
+
+  if (opensearchIntegrationEnabled == "false") {
+    logger.info("OPENSEARCH_INTEGRATION_ENABLED is 'false' - disabling and returning log only object");
```
rocketeerbkw: Thinking about #2930 and #3164, the new default logging level would be `info`. The currently proposed level of logging would print _something_ to stdout every time an opensearch related action (project/group changes) is made in the api, and I'm not sure that's necessary?

Maybe all of these could be `debug` instead of `info` (would result in nothing being logged in production)? Or keep `info` but only outputting the "OPENSEARCH_INTEGRATION_ENABLED" log once on application boot instead of on every instantiation of `OpendistroSecurityOperations` (and get rid of the logging for each function call)?

The logging changes aren't implemented yet though, so wouldn't hurt anything to get merged as is.

rocketeerbkw (approved):

services/api/src/resources/group/opendistroSecurity.ts
bomoko: I was half and half about the level of logging on this one. `info` was an explicit choice, because I though that given the context that this is used, it might be useful for it to be more verbose by default - but you're right, I think that the changes you're suggesting make a lot of sense.

bomoko: making 'em all debug. Easiest fix.

shreddedbacon: Missed one: https://github.com/uselagoon/lagoon/blob/4d117442a1e048c550dfe291b27524ee65d0c397/services/api/src/resources/group/opendistroSecurity.ts#L38

PR #3293, opened 2022-09-19
Adds new deploymentByFilter query to api

bomoko:
This adds a new query `deploymentsByFilter` that allows one to show and filter the current state of deployments

By default it returns all deployments for a user currently in states `["NEW", "PENDING", "RUNNING"]` (these can be customized) - additionally it can take an array of deploy target ids if one wanted to only see deployments on a particular set of remotes.

# Checklist


services/api/src/resources/deployment/resolvers.ts
```diff
@@ -143,6 +143,59 @@ export const getDeploymentsByBulkId: ResolverFn = async (
...
+      id: keycloakGrant.access_token.content.sub
+    });
+  }
+
+  let queryBuilder = knex.select("deployment.*").from('deployment').
+      join('environment', 'deployment.environment', '=', 'environment.id').
+      join('project', 'environment.project', '=', 'project.id');
+
+  if(openshifts) {
+    queryBuilder = queryBuilder.whereIn('deployment.remote_id', openshifts);
```
shreddedbacon: Should this be `whereIn('environment.openshift', openshifts)` ?

The `remote_id` is just a uuid that is used for tracking logs mainly.

EDIT: not exactly this query, but one that can check all the deployments associated environments openshift id

bomoko: Well, this query works :D Resolved.

shreddedbacon (approved):
:shipit:

PR #3300, opened 2022-10-05
Filters out deleted environments in getDeploymentsByFilter

bomoko:
Currently getDeploymentsByFilter returns all deployments, including those of deleted environments. This filters out the latter

It's an addendum to #3293


shreddedbacon (approved):
:rocket:

PR #3309, opened 2022-10-11
Don't update environment routes/services for cancelled/failed deployments

rocketeerbkw:
# Checklist

When a deployment is manually cancelled or fails, the `controllerhandler` would still try to update the routes and services for the deployments environment. This could result in the routes being set to the string "undefined" for example.

We don't want to change environment data if a deployment wasn't successful for any reason.

# Closing issues

Closes https://github.com/uselagoon/lagoon-ui/issues/55

### DIFF
--- a/node-packages/commons/src/api.ts
+++ b/node-packages/commons/src/api.ts
@@ -68,9 +68,32 @@ interface RestorePatch {
   restoreLocation?: string;
 }
 
+interface EnvironmnetPatch {
+  autoIdle?: number;
+  created?: string;
+  deployBaseRef?: string;
+  deployHeadRef?: string;
+  deployTitle?: string;
+  deployType?: DeployType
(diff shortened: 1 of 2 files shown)


shreddedbacon:
There are valid cases to update routes in a failed build. Like if a post-deploy task fails, the routes are still valid, and the `remote-controller` does ship routes for failed builds if it fails if they are available, same for cancelled builds.

There is a condition though where the routes are not injected, and that is primarily on cancelled builds that the cluster has no resource tracking for, like if the build is super old and the associated pod and lagoonbuild resource no longer exist. This case results in no routes being provided in the response to the controllerhandler.

It could be worth checking that routes is actually provided, and adding them if they are.

tobybellwood:
This is now handled in the actions-handler

PR #3310, opened 2022-10-11
fix: add additional metadata to pullrequests triggered by the API

shreddedbacon:
# Checklist

An issue discovered with logs2notifications is that now that it is processing the `meta` data of messages to create notifications, it has uncovered that some fields that would normally be in the `message` that is created in the resolvers, are not provided in the metadata of the log message itself.

When triggering a pullrequest deployment from the API, it is missing the `branchName` in the metadata, for PRs this is just the `pr-#`, but it is also missing other meta that could be useful. This adds additional metadata for pullrequests so that logs2notifications can consume it.


bomoko (approved):
LGTM - all data should be in scope, given we use it right above.

Issue #3161, opened 2022-05-30
Cannot log in to Lagoon logs UI


smlx:
# Describe the bug

The Lagoon logs UI (Kibana or Opensearch Dashboards) gets a cookie from Keycloak which passes role information to the logs UI in order to enforce index viewing permissions.

If the user is in too many groups, this cookie can get so large that it reaches the [de facto browser limit of 4096 bytes](https://stackoverflow.com/q/52203972) and is rejected by the browser.

This causes a redirect loop when trying to log in to the logs UI.

## To Reproduce

1. Add a user to lots of projects in Lagoon
2. User visits the logs UI
3. User cannot login and encounters a "too many redirects" error.

## Expected behavior

Lagoon should avoid creating such a large cookie.

## Screenshots

n/a

## Additional context

In order to add users to a project a user must be `OWNER` within a project group. This means that a Lagoon user who is `OWNER` in their main customer group which contains all their projects will often also want to be `OWNER` in individual project groups which will increase the likelihood of hitting this bug.

The code creating the cookie looks roughly like this:

```javascript
var ArrayList = Java.type("java.util.ArrayList");
var groupsAndRoles = new ArrayList();
var forEach = Array.prototype.forEach;
// add all groups the user is part of
forEach.call(user.getGroups().toArray(), function(group) {
	// remove the group role suffixes
	// lets check if the group has a parent if this is a child
	if(group.getFirstAttribute("type") == "role-subgroup") {
		var parent = group.getParent();
		if(parent.getFirstAttribute("type") == "project-default-group") {
			var projectIds = parent.getFirstAttribute("lagoon-projects");
			if(projectIds !== null) {
				forEach.call(projectIds.split(","), function(g) {
					groupsAndRoles.add("p" + g);
				});
				return;
			}
		}
	}
	var groupName = group.getName().replace(/-owner|-maintainer|-developer|-reporter|-guest/gi,"");
	groupsAndRoles.add(groupName);
	return;
});
// add all roles the user is part of
forEach.call(user.getRoleMappings().toArray(), function(role) {
	var roleName = role.getName();
	groupsAndRoles.add(roleName);
});
exports = groupsAndRoles;
```

This adds roles for all the projects that the user is an `OWNER` of even if the user is already an `OWNER` of a group that contains the project.

[...]

PR #3319, opened 2022-10-31
create new mappers for groups-and-roles

tobybellwood:
As discussed in https://github.com/uselagoon/lagoon/issues/3161#issuecomment-1141937412

This PR adds a replacement groups-and-roles mapper to the keycloak-opensearch connector. It improves the current solution by adding all the groups the user is in, then checking any additional projects and excluding them if they're already covered by a group access. This reduces the size of the passed object to Keycloak.


services/keycloak/javascript/mappers/groups-and-roles.js
```diff
@@ -1,33 +1,50 @@
...
-        }
-    }
-  }
-
-  var groupName = group.getName().replace(/-owner|-maintainer|-developer|-reporter|-guest/gi,"");
-  groupsAndRoles.add(groupName);
-  return;
+	// remove the group role suffixes
+	// lets check if the group has a parent if this is a child
+	var groupName = group.getName().replace(/-owner|-maintainer|-developer|-reporter|-guest/gi,"");
```
smlx: This should probably have an anchor on the end of the group name, and remove the `gi` flags. Otherwise couldn't users create a group in Lagoon like `i-am-the-Owner` which will mess up this logic?

smlx (commented):
I don't think I'm the right person to review this. Maybe @rocketeerbkw (as the Keycloak SME) or @shreddedbacon (as the original author)?

shreddedbacon (approved):
This was running in our test infra for some time prior and it provides the right group/project reductions as required.

PR #3325, opened 2022-11-03
chore: support consuming deployer-token from service account

shreddedbacon:
# Checklist

In kubernetes 1.24, service accounts no longer get automatically generated token secrets. It is possible to generate long lived tokens, but using the TokenRequest API is the recommended way forward.

The remote-controller will now use the serviceaccount on all pods started by it, which uses volume projection to mount the token into the pod automatically.

The default will be to try the existing token location, but if one has been provided the service account, it will use this this.

Same as https://github.com/uselagoon/build-deploy-tool/pull/140


tobybellwood (approved):
agreed - same as BDT change

PR #3327, opened 2022-11-03
fix: some queries retrieve a deleted environment

shreddedbacon:
# Checklist

Some of the sql queries used in environment resolvers don't retrieve environments in the right order, this can result in a deleted environment being returned in the first result.

This adjusts some of the querires to use to order by id, or select the environment that isn't deleted.

I discovered this when I had created an environment called `main2` and then deleted it. I then re-created this environment and set it up as an active/standby environment.
When triggering the active/standby switch, the task was created against the deleted environment, not the new one, so it never showed up in the API.

### DIFF
--- a/services/api/src/resources/env-variables/sql.ts
+++ b/services/api/src/resources/env-variables/sql.ts
@@ -40,6 +40,7 @@ export const Sql = {
     knex('environment')
       .where('name', '=', name)
       .andWhere('project', '=', projectId)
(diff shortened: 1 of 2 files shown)


tobybellwood:
@rocketeerbkw - can you think of anywhere where we'd expect `selectEnvironmentByNameAndProject` to specifically return a deleted environment

rocketeerbkw (approved):
I checked all the callers of the changed functions and it won't break anything that I can see.

PR #3336, opened 2022-11-13
Adding validation to addOrUpdateEnvVariableByName resolver

timclifford:


shreddedbacon:
Should we add validations for more than this variable too? There are quite a few called out in the docs

tobybellwood:
This adds a layer of complexity that will become increasingly hard to maintain and provide little benefit

PR #3346, opened 2022-11-28
Add facts by project and environment name

timclifford:
Introduces project and environment args to addFacts input so that bulk facts can be added without needing to know the project's environment id.

Note: Changes have been made to existing addFact resolver also that's worth a 👀


services/api/src/resources/fact/resolvers.ts
```diff
@@ -186,6 +188,11 @@ export const addFact: ResolverFn = async (
   { input: { id, environment: environmentId, name, value, source, description, type, category, keyFact, service } },
   { sqlClientPool, hasPermission, userActivityLogger }
 ) => {
+  if (environmentId == undefined) {
+    logger.error(`No environment ID given for fact: ${name}`)
+    throw new Error(`No environment ID given for fact: ${name}`)
+  }
+
```
timclifford, line 244: Needed to add this since I had to remove the 'required' environmentID check from the schema input and handle that here instead.

services/api/src/typeDefs.js
```diff
@@ -329,6 +329,8 @@ const typeDefs = gql`
   }
 
   input AddFactsInput {
+    project: String
+    environment: String
```
timclifford, line 337: Would be good to have the ability to allow multiple input types for `addFacts`, but AFAIK unions only work for Object types currently?

bomoko: As discussed, I think introducing multiple mutations is probably right

bomoko (changes requested):

bomoko (approved):
This looks good!

PR #3351, opened 2022-11-29
feature: add queued status type and build step field

### LINKED ISSUE #3352: Change QoS/Bulk deployments "status" from "New" to "Queued"
Change QoS/Bulk deployments "status" from "New" to "Queued", so that Qos/Bulk deploys can be easily identified from standard "new" deployments.

image__11__png

shreddedbacon:
# Checklist

To make it clear when a build or task is queued, add a queued status. This will make it so that when a build or task has been received by the remote-controller and marked as in a queue for processing, Lagoon will be able to display it as `queued` rather than `new` or `pending`

This also updates the enums to be the same for tasks and builds now that the remote-controller has supported using the same status types for both builds and tasks for some time.

Also adds support for `buildStep` to the deployment, currently Lagoon sends this data in the payload back to core, but it is not consumed by the API. This means we will be able to expose the current build step in the UI, as this build step contains useful information about which step the build is currently progressing through, or other information that could be useful at a quick glance in an overview page.

closes #3352


shreddedbacon:
@CGoodwin90 just want to get you to :eyes: the migration part to make sure it looks ok

CGoodwin90:
Looks good to me!

CGoodwin90 (approved):

shreddedbacon:
> Looks good to me!

I buggered something up, also turns out you can't do enum the way I thought, so basically now just dropping the column and recreating it seems to do the trick.

Also, I'm not 100% sure about the rollback, do I need to do the column check on rollback?

CGoodwin90:
> > Looks good to me!
> 
> I buggered something up, also turns out you can't do enum the way I thought, so basically now just dropping the column and recreating it seems to do the trick.
> 
> Also, I'm not 100% sure about the rollback, do I need to do the column check on rollback?

Ah I see, turns out Knex can't alter enums after creation. An alternate approach would be to use Knex.raw - https://knexjs.org/guide/raw.html.

Don't necessarily need a column check for the rollback, but I've been adding them just incase somethng goes awry.

services/api/database/migrations/20221129221152_queuedstatus.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+ exports.up = async function(knex) {
+    // cant alter enums in place, so drop the column first :D
+    buildStep = await knex.schema.hasColumn('deployment', 'build_step');
+    if (!buildStep) {
```
rocketeerbkw: knex migrations always run once and in sequence, so this form of protection that we had to do in our stored procedures shouldn't be necessary?

services/api/src/typeDefs.js
```diff
@@ -67,15 +68,14 @@ const typeDefs = gql`
   }
 
   enum TaskStatusType {
-    ACTIVE
-    SUCCEEDED
```
rocketeerbkw, line 71: Removing enum values is a BC break. Even if we don't use them anywhere in lagoon we can't remove them yet. Adding an enum value is considered "dangerous" but it didn't cause problems [last time](https://github.com/uselagoon/lagoon/pull/3114).

services/api/database/migrations/20221129221152_queuedstatus.js
```diff
@@ -0,0 +1,49 @@
...
+        return knex.schema
+        .alterTable('deployment', (table) => {
+            table.dropColumn('status');
```
rocketeerbkw: Dropping the column wipes out all status data for all deployments, we should not do that. I tested knex and I'm pretty sure all you need to do is add `.alter()` at the end of your enum statement instead of doing a drop/add.

`table.enu('status', ['new', 'pending', 'running', 'cancelled', 'error', 'failed', 'complete', 'queued']).notNullable().alter()`

If that really doesn't work, we need to do some raw SQL to alter the column like we [used to do](https://github.com/uselagoon/lagoon/blob/main/services/api-db/legacy-migration-scripts/01-migrations.sql#L1687-L1688) in order to retain data correctly.

services/api/src/resources/deployment/resolvers.ts
```diff
@@ -15,6 +15,7 @@ import {
   pubSub,
   createEnvironmentFilteredSubscriber
 } from '../../clients/pubSub';
+// @ts-ignore
 import { getConfigFromEnv, getLagoonRouteFromEnv } from '../../util/config';
```
rocketeerbkw: The api container would fail to build if there were type errors, I think your local development environment isn't resolving the imports correctly. I would very much like to not commit these ignores.

rocketeerbkw (changes requested):

services/api/database/migrations/20221129221152_queuedstatus.js
shreddedbacon: I don't know, @CGoodwin90 @tobybellwood ?

services/api/src/resources/deployment/resolvers.ts
shreddedbacon: vscode, do I need some extension or something installed to resolve these?

services/api/database/migrations/20221129221152_queuedstatus.js
shreddedbacon: Sure I'll swap to raw to save things

services/api/src/typeDefs.js
shreddedbacon: Ugh, fine

services/api/src/resources/deployment/resolvers.ts
rocketeerbkw: discussed elsewhere but, on the extremely improbable chance that someone sees this... you need to run `yarn build` in the api at least once to resolve packages in `@lagoon/commons`

services/api/database/migrations/20221129221152_queuedstatus.js
bomoko: can confirm

shreddedbacon: Sweet as, will rip that out then

shreddedbacon:
Would be good to get this in soon so that clusters with QoS enabled get some visibility of when builds are queued.

tobybellwood (approved):
I've reviewed the migration, and now with a safe rollback, I'm happy. I would like @rocketeerbkw to re-review the code logic

services/actions-handler/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/cheshir/go-mq v1.2.0
-	github.com/uselagoon/machinery v0.0.7
+	github.com/uselagoon/machinery v0.0.8-0.20221129235134-66a9c5a0ca7b
```
rocketeerbkw, line 7: This is a reference to a specific commit? Is that a datestamp 2022-11-29? Is there a stable release we could reference instead?

This is outside my knowledge area, I'll defer entirely to your judgement.

rocketeerbkw (approved):
Everything api/api-db wise LGTM.

One question about the actions-handler dependencies, but feel free to ignore if it doesn't make sense.

services/actions-handler/go.mod
shreddedbacon: Yes, this is a reference to a specific commit, and that is the timestamp and commit sha. I can bump this to a stable release but the feature that it implements isn't in the Lagoon API until this PR is merged.

rocketeerbkw: ah, right. `machinery` is the api client, and if you try to query a non-existent field graphql will error. I'd be interested in figuring out a better way to handle this chicken/egg problem, but not right now

PR #3356, opened 2022-11-30
fix: check for disabledDeployments in webhooks too

shreddedbacon:
# Checklist

When `deploymentsDisabled` was introduced, the webhooks were missed.

I verified this by setting deploymentsDisabled on a project in local lagoon `lagoon update project -p high-cotton --deploymentsDisabled 1` and then running a [curl against the webhook-handler to try start a build](https://gist.github.com/shreddedbacon/fd806546f843f1d7d0f0cbdc4646962b)

### DIFF
--- a/node-packages/commons/src/api.ts
+++ b/node-packages/commons/src/api.ts
@@ -9,6 +9,7 @@ interface Project {
   slack: any;
   name: string;
   openshift: any;
+  deploymentsDisabled: number;
 }
 
 interface GroupPatch {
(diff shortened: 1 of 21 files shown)


shreddedbacon:
I'm not sure how to write a test case for this. The webhook-handler just accepts the payload and sends it to the queue for processing. Since there is no feedback provided to the client on if it succeeds or not beyond a log message when webhooks2tasks processes it (that also goes to notifications) writing a test case would be quite complex.

tobybellwood:
The webhook tests are in dire need of re-examination anyway, happy to proceed without a dedicated test.

PR #3369, opened 2022-12-15
Feature: Organizations

shreddedbacon:
To be merged with https://github.com/uselagoon/lagoon-ui/pull/94

 

# Checklist

This PR introduces support for organizations to Lagoon. Organizations are a way to logically assign projects and groups together and allows owners of organizations to have more control over their projects, groups, and notifications.

There are 2 new user roles, organization owner, and organization viewer.
* Owner
  * allows creation of projects within the organization
  * allows creation of groups and users within those groups within the organization
  * allows creation of top level notifications (slack, teams, rocketchat, email, webook...) within the organization
  * allows adding notifications to projects within the organization
  * ability to add other owners and viewers
* Viewer
  * view all groups, users, projects, and notifications within the organization
  * no ability to modify or create anything within the organization

# Benefits of organizations
* Projects created within an organization will only be able to deploy to targets defined within that organization. Even when utilizing deploy target configurations. Only platform admins can add additional deploytargets to an organization.
* Groups within an organization can only be added to projects in the same organization
* Notifications within an organization can only be added to projects in the same organization
* Organization owners can add any user within a Lagoon to a group within their organization without that user needing to be in their organization, this user can be removed from groups easily by the organization owner whenever they want.
* Organization owners don't have the ability to deploy any environments in projects without being a member of a group associated to that project first.

# Screenshots
Here are some screenshots of how the UI makes organizations visible to users. These screenshots are subject to change in the released version but provide a general view of what is possible
## Main overview
This is the quick overview of the organization page, you can see the quotas available, which targets projects can be configured to deploy against, and the users associated to this organization management
image
## Groups overview
This is where owners can create new groups
image
## Group/User management
Owners can associate users to these groups here too, selecting which role they should have. Removal of users is also available
image
## Projects overview
Overview of projects within the organization
image
## Project group/notification management
This is where owners can associate groups and notifications to their projects
image
## Notification overview
This is where owners can create new notifications for their organization to then be able to assign to projects
image


services/api/src/clients/redisClient.ts
```diff
@@ -91,12 +98,32 @@ export const saveProjectGroupsCache = async (projectId, groupIds) =>
...
+export const saveOrganizationGroupsCache = async (organizationId, groupIds) =>
+  sadd(`organization-groups:${organizationId}`, groupIds);
+export const deleteOrganizationGroupsCache = async (organizationId, groupIds) =>
+  del(`organization-groups:${organizationId}`);
+
+export const getUsersOrganizationCache = async organizationId =>
+  smembers(`organization-users:${organizationId}`);
+export const saveUsersOrganizationCache = async (organizationId, userIds) =>
+  sadd(`organization-users:${organizationId}`, userIds);
+export const deleteUsersOrganizationCache = async (organizationId, userIds) =>
```
bomoko: Should this function take in `userIds`?

services/api/src/models/group.ts
```diff
@@ -286,6 +310,52 @@ export const Group = (clients: {
...
+          : attribute.value;
+        return R.test(new RegExp(`\\b${organizationId}\\b`), value);
+      }
+
+      return false;
+    };
+
+    let groupIds = [];
+
+    if (R.isEmpty(groupIds)) {
```
bomoko: unless I'm reading this wrong, isn't this always going to be empty?

node-packages/commons/src/util/func.ts
```diff
@@ -21,3 +22,24 @@ export const jsonMerge = function(a, b, prop) {
   });
   return reduced.concat(b);
 }
+
+// will return only what is in a1 that isn't in a2
+// eg:
+// a1 = [1,2,3,4]
+// a2 = [1,2,3,5]
+// arrayDiff(a1,a2) = [4]
+export const arrayDiff = (a1, a2) => {
```
bomoko: So I think a more functional/idiomatic way of calculating set difference is something like this
```
const arrayDiff = (a:Array<any>, b:Array<any>) =>  a.filter(e => !b.includes(e));
```

I've also added typing here to enforce arrays

services/api/src/models/user.ts
```diff
@@ -202,6 +262,58 @@ export const User = (clients: {
...
+          : attribute.value;
+        return R.test(new RegExp(`\\b${organizationId}\\b`), value);
+      }
+
+      return false;
+    };
+
+    let userIds = [];
+
+    if (R.isEmpty(userIds)) {
```
bomoko: Again, this will be empty every time - we can drop this conditional

services/api/src/models/user.ts
```diff
@@ -202,6 +262,58 @@ export const User = (clients: {
...
+      const keycloakUsers = await keycloakAdminClient.users.find();
+      // @ts-ignore
+      userIds = R.pluck('id', keycloakUsers);
+    }
+
+    let fullUsers = [];
+    for (const id of userIds) {
+      const fullUser = await keycloakAdminClient.users.findOne({
```
bomoko, line 298: We have an earlier instance of this call "findone" for groups wrapped in an empty try/catch block - in this case we don't want to catch any exceptions? (Just raising it because I noticed it)

services/api/src/resources/deploytargetconfig/resolvers.ts
```diff
@@ -102,6 +103,21 @@ export const updateEnvironmentDeployTarget: ResolverFn = async (
     project: environmentObj.project
   });
 
+  // check the project has an organization id, if it does, check that the organization supports the requested deploytarget
+  const projectdata = await projectHelpers(sqlClientPool).getProjectById(environmentObj.project)
+  if (projectdata.organization != null) {
+    const deploytargets = await organizationHelpers(sqlClientPool).getDeployTargetsByOrganizationId(projectdata.organization);
+    let validDeployTarget = false
```
bomoko: Given this is repeated verbatim in "UpdateDeployTargetConfig" and "AddDeployTargetConfig" - it's a good candidate for factoring out - not a blocker or necessary, but in this case it could be a clean single function call that returns a bool used by all three of the resolvers.

services/api/src/resources/group/resolvers.ts
```diff
@@ -232,7 +268,43 @@ export const addGroup: ResolverFn = async (
...
+    const orgGroups = await models.GroupModel.loadGroupsByOrganizationId(input.organization);
+    let groupCount = 0
+    for (const pGroup in orgGroups) {
+      // project-default-groups don't count towards group quotas
+      if (orgGroups[pGroup].attributes["type"] != "project-default-group") {
+        groupCount++
+      }
+    }
+
+    if (groupCount >= organizationData.quotaGroup) {
```
bomoko, line 309: Just as an aside - what might be interesting and/or useful is to move org rules into their own object type - this would let us add and validate org events against (potentially) arbitrary sets of rules.

services/api/src/resources/notification/resolvers.ts
```diff
@@ -35,8 +55,25 @@ export const addNotificationEmail: ResolverFn = async (
   { input },
   { sqlClientPool, hasPermission }
 ) => {
-  await hasPermission('notification', 'add');
+  if (input.organization != null) {
```
bomoko: Again, I think that these checks where we have repeated boilerplate could probably be factored out _very_ cleanly.

services/api/src/resources/notification/resolvers.ts
```diff
@@ -166,9 +279,18 @@ export const deleteNotificationEmail: ResolverFn = async (
   { input },
   { sqlClientPool, hasPermission }
 ) => {
-  await hasPermission('notification', 'delete');
-
   const { name } = input;
+  const check = await query(
```
bomoko, line 225: Repeated code could also usefully be factored out.

services/api/src/resources/notification/resolvers.ts
```diff
@@ -345,18 +506,87 @@ export const getNotificationsByProjectId: ResolverFn = async (
...
   { sqlClientPool, hasPermission }
 ) => {
-  await hasPermission('notification', 'update');
-
-  const { name } = input;
-
   if (isPatchEmpty(input)) {
     throw new Error('input.patch requires at least 1 attribute');
   }
+  const { name } = input;
```
bomoko, line 503: Again, a nice easy refactor.

services/api/src/resources/organization/resolvers.ts
```diff
@@ -0,0 +1,676 @@
...
+  // let oid = args.organization;
+  if (args.organization) {
+    oid = args.organization;
+  }
+
+  await hasPermission('organization', 'view', {
+      organization: oid,
+  });
+
+  const rows = await query(
```
bomoko: this seems like it should be simple enough to knex. Not necessarily a blocker, but since we're trying to move all our queries that way, might make sense to do it now (I'm well aware that some of this code predates us doing that :) )

services/api/src/resources/organization/resolvers.ts
```diff
@@ -0,0 +1,676 @@
...
+      oid = organization;
+    }
+
+    await hasPermission('organization', 'view', {
+        organization: oid,
+    });
+
+    const rows = await query(
+        sqlClientPool,
+        `SELECT *
```
bomoko: More raw SQL that could be moved to knex.

services/api/src/clients/redisClient.ts
shreddedbacon: Not used, removed

services/api/src/resources/group/resolvers.ts
shreddedbacon: yeah, i have thought about something like this a bit too. need to think about it a bit more though

services/api/src/models/user.ts
shreddedbacon: i think the try/catch was copypasta, so i've removed it

tobybellwood (approved):
okay to merge into test infra and monitor

PR #3380, opened 2023-01-13
fix: only update routes if the value isnt empty

shreddedbacon:
# Checklist

If there are no routes, the actions-handler tries to set the routes to the empty value. This could result in the `routes` value of a project never being unset if the environment has no routes in kubernetes.

This is for testing at the moment to see if the problem can be solved properly in remote-controller too.


shreddedbacon:
closing this as not required. this was used temporarily for a specific case with migrating environments. it isn't required now

PR #3393, opened 2023-02-07
change: task visibility improvements and changes to admin tasks

### LINKED ISSUE #3392: Advanced task creation permission check error
# Describe the bug

When creating an advanced task as a non-platform admin, if `adminTask` is undefined, task creation fails with a permission error.

## To Reproduce

1. Create an advanced task:
```
mutation createMyTask {
  addAdvancedTaskDefinition(
    input:{
      name: "My Task"
      description: "My Task that does stuff"
      type: COMMAND
      service: "cli"
      command: "env"
      permission: MAINTAINER
      project: 18
    }
  ){
    ... on AdvancedTaskDefinitionCommand {
      id
      name
      description
      service
      command
    }
  }
}
```
2. See that it fails with a permission error
3. Update mutation to include in the input `adminTask: 0`
4. See that the task is created successfully

## Expected behavior

Task creation succeeds if I have permission to create task

## Additional context

The check is for `input.adminTask != 0`, but by default if not provided this value is `undefined` so this check fails.
https://github.com/uselagoon/lagoon/blob/3e1e22ba91a40c2ddb5b2a3983a057f6191f5f45/services/api/src/resources/task/task_definition_resolvers.ts#L742

shreddedbacon:
# Checklist

The current showUi implementation processes what can be seen in the UI, it only hides the options from the dropdown menu, but doesn't prevent them from being shown elsewhere. It also doesn't prevent the logs for these tasks being hidden.

This moves all the hiding functionality to behind RBAC in the API.

It removes `showUi` and creates `adminOnlyView` instead. These are used by resolvers to check if the user is admin or not, and will filter out tasks in resolvers if these are set to true.

Setting `adminOnlyView` to true on a task will hide this task from the drop down menu, and also hide the item in the task view thus preventing access to the logs for non-admin users.

`adminTask` is split into:
* `deployTokenInjection`, as this is used as an identifier to inject the lagoon-deployer token into tasks. Only admins can set this flag.
* `projectKeyInjection` as a second option, this can be used to control the injection of the projects ssh key used for tasks etc.

It leaves `showUi` and `adminTask`, but these reflect the value of the replacements to an extent.
* showUi is the inverted value of adminViewOnly
* adminTask is the same as deployTokenInjection

# Closing issues

closes #3392

### DIFF
--- a/services/api/src/resources/task/resolvers.ts
+++ b/services/api/src/resources/task/resolvers.ts
@@ -131,9 +131,6 @@ export const getTasksByEnvironmentId: ResolverFn = async (
   let adminTasks = false
   // do a check of all returned tasks for any that are admin tasks
   for (const row of rows) {
-    if (row.adminTask == true) {
-      adminTasks = true
-    }
     if (row.adminOnlyView == true) {
       adminTasks = true
     }
@@ -146,7 +143,7 @@ export const getTasksByEnvironmentId: ResolverFn = async (
     } catch (err) {
       for (const row in rows) {
         // then remove any admintasks from the result
-        if (rows.hasOwnProperty(row) && rows[row].adminTask == true) {
+        if (rows.hasOwnProperty(row) && rows[row].adminOnlyView == true) {
             delete rows[row];
         }
       }
@@ -172,7 +169,7 @@ export const getTaskByTaskName: ResolverFn = async (
   }
 
   const rowsPerms = await query(sqlClientPool, Sql.selectPermsForTask(task.id));
-  if (R.path(['0', 'admin_only_view'], rowsPerms) || R.path(['0', 'admin_task'], rowsPerms)) {
+  if (R.path(['0', 'admin_only_view'], rowsPerms)) {
     await hasPermission('project', 'viewAll');
   } else {
     await hasPermission('task', 'view', {
@@ -200,7 +197,7 @@ export const getTaskByRemoteId: ResolverFn = async (
   }
 
   const rowsPerms = await query(sqlClientPool, Sql.selectPermsForTask(task.id));
-  if (R.path(['0', 'admin_only_view'], rowsPerms) || R.path(['0', 'admin_task'], rowsPerms)) {
+  if (R.path(['0', 'admin_only_view'], rowsPerms)) {
     await hasPermission('project', 'viewAll');
   } else {
     await hasPermission('task', 'view', {
@@ -228,7 +225,7 @@ export const getTaskById: ResolverFn = async (
   }
 
   const rowsPerms = await query(sqlClientPool, Sql.selectPermsForTask(task.id));
-  if (R.path(['0', 'admin_only_view'], rowsPerms) || R.path(['0', 'admin_task'], rowsPerms)) {
+  if (R.path(['0', 'admin_only_view'], rowsPerms)) {
     await hasPermission('project', 'viewAll');
   } else {
     await hasPermission('task', 'view', {
@@ -253,7 +250,8 @@ export const addTask: ResolverFn = async (
       service,
       command,
       remoteId,
-      adminTask,
+      deployTokenInjection,
+      projectKeyInjection,
       adminOnlyView,
       execute: executeRequest
     }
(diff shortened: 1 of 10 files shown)


shreddedbacon:
Noting that as the current released UI has `showUi` support still, admin view only tasks will be hidden by the UI even for admins until the showUi stuff is reverted. I built in the showUi backwards compatability support only because of the UI

Once the UI is reverted, showUi will be ignored, and the admin view only tasks will be visible to admins only.

I've updated this PR with the [PR image for the UI](https://github.com/uselagoon/lagoon-ui/pull/102) in the docker-compose stack only for local checking.

services/api/src/resources/task/resolvers.ts
```diff
@@ -125,7 +126,30 @@ export const getTasksByEnvironmentId: ResolverFn = async (
...
   }
 
-  return query(sqlClientPool, queryBuilder.toString());
+  const rows = await query(sqlClientPool, queryBuilder.toString())
+
+  let adminTasks = false
+  // do a check of all returned tasks for any that are admin tasks
+  for (const row of rows) {
+    if (row.adminOnlyView == true) {
+      adminTasks = true
```
bomoko: Although this is in memory, so it'll be nice and fast, you could "break" here, since the first instance of an admin task tells us that there are admin tasks.

bomoko: I think that to be consistent with line 147, maybe add a semicolon (I'm not sure where we stand on ASI with ts, but I guess consistency in the same block of code is good.

services/api/src/resources/task/resolvers.ts
```diff
@@ -125,7 +126,30 @@ export const getTasksByEnvironmentId: ResolverFn = async (
     queryBuilder = queryBuilder.limit(limit);
   }
 
-  return query(sqlClientPool, queryBuilder.toString());
+  const rows = await query(sqlClientPool, queryBuilder.toString())
+
+  let adminTasks = false
```
bomoko: This is repeated three times in the code. I'm thinking that maybe we introduce something like a "filters.ts" and just refactor out to that?

Something like 

```
import * as R from 'ramda';

export const Filters = {
    filterAdminTasks: async (hasPermission, rows) => {
        let adminTasks = false
        // do a check of all returned tasks for any that are admin tasks
        for (const row of rows) {
          if (row.adminOnlyView == true) {
            adminTasks = true;
            break;
          }
        }

        // if any of the tasks are admin tasks, check the user has admin permission
        if (adminTasks) {
          // we do this so we only check admin permission once, instead of on all tasks
          try {
            await hasPermission('project', 'viewAll');
          } catch (err) {
            return rows.filter((row) => row.adminOnlyView != true);
          }
        }
        // else we simply return the rows untouched
        return rows;
    }
}
```

Which can then be called like
`rows = Filters.filterAdminTasks(hasPermissions, rows);`

services/api/src/resources/task/resolvers.ts
```diff
@@ -125,7 +126,30 @@ export const getTasksByEnvironmentId: ResolverFn = async (
...
+  }
+  // if any of the tasks are admin tasks, check the user has admin permission
+  if (adminTasks) {
+    // we do this so we only check admin permission once, instead of on all tasks
+    try {
+      await hasPermission('project', 'viewAll');
+    } catch (err) {
+      for (const row in rows) {
+        // then remove any admintasks from the result
+        if (rows.hasOwnProperty(row) && rows[row].adminOnlyView == true) {
```
bomoko: you could just use an array.filter here (see code example in the other comment)

services/api/src/resources/task/task_definition_resolvers.ts
```diff
@@ -249,6 +272,16 @@ export const addAdvancedTaskDefinition = async (
     project
   );
 
+  if (input.deployTokenInjection == undefined) {
```
bomoko: So for these I think you could do something like this
`input.deployTokenInjection = input.deployTokenInjection || false;`

bomoko (commented):
Very nice. I've just added some comments/suggestions for you to think about.

bomoko (approved):

PR #3399, opened 2023-02-20
Fix insecure registries running k3d/test on docker desktop for mac

rocketeerbkw:
# Checklist

Some minor fixes to the Makefile to fix running `k3d/test` on Docker Desktop for Mac. Fixes insecure registries and cleans up the k3d proxy.

# Closing issues

n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -527,7 +527,10 @@ k3d/cluster: local-dev/k3d
 		docker network create k3d || true \
 		&& export KUBECONFIG=$$(mktemp) \
 		K3DCONFIG=$$(mktemp ./k3dconfig.XXX) \
-		K3D_NODE_IP=$$(docker run --rm --network k3d alpine ip -o addr show eth0 | sed -nE 's/.* ([0-9.]{7,})\/.*/\1/p') \
+		K3D_NODE_IP=$$(docker run --rm --network k3d alpine ip -o addr show eth0 | sed -nE 's/.* ([0-9.]{7,})\/.*/\1/p') && \
+		if [[ $(ARCH) == darwin ]]; then \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -527,7 +527,7 @@ k3d/cluster: local-dev/k3d
 		docker network create k3d || true \
 		&& export KUBECONFIG=$$(mktemp) \
 		K3DCONFIG=$$(mktemp ./k3dconfig.XXX) \
-		K3D_NODE_IP=$$(docker run --rm --network k3d alpine ip -o addr show eth0 | sed -nE 's/.* ([0-9.]{7,})\/.*/\1/p') \
```
rocketeerbkw, line 530: This doesn't work on my machine. I think what this is trying to do is guess the IP address of the k3d node to set it as the [registry in k3d](https://github.com/uselagoon/lagoon/blob/main/k3d.config.yaml.tpl#L16-L22). It does that by getting the "next" ip address that docker will allocate.

On my machine this was returning `172.21.0.2`, but it looks like k3d starts a "tools" container first, which grabs that IP, and the k3d node gets the next one `172.21.0.3`.
```
INFO[0000] Starting new tools node...
INFO[0000] Starting Node 'k3d-lagoon-tools'
INFO[0001] Creating node 'k3d-lagoon-server-0'
```

Since the later steps in the make file use a different method for getting the _true_ node IP, there is a mismatch and the image pulls for deploying lagoon core/remote fail.

Not sure what happens on other architectures, maybe the tools container isn't needed or it gets created last? Not sure what the better method would be to solve this?

rocketeerbkw: Fixed by adding another darwin exception

PR #3414, opened 2023-03-21
fix: groups in opensearch client token so that group-tenant association works

shreddedbacon:
# Checklist

#3319 introduced a way to reduce the number of groups and project-default groups being added to user tokens in opensearch clients to reduce the impact of large cookie sizes preventing users from logging in to opensearch. Unfortunately, this also introduced a bug in which the groups the user was in were consolidated down too. This caused groups the user is in to not match up to their associated tenant in opensearch.

This PR fixes it so that groups are no longer consolidated down, only project-default groups are.


tobybellwood (approved):
Looks good to me and shown working in testing

PR #3415, opened 2023-03-21
fix: filter out groups user isn't in when working out groups for user

shreddedbacon:
# Checklist

#3397 introduced some refactoring to the API to make permissions checks faster, but there is a bug when a group has multiple rolesubgroups and it adds all of these to the users groups that are returned. This could result in a user getting a role they may not have being added to their roles.

This fixes it by filtering out all the subgroups that the user isn't in from the groups returned from keycloak before populating the users groups with attributes.


tobybellwood (approved):
This makes sense, a smarter filter

PR #3419, opened 2023-03-28
chore: remove kind and apiversion from restore

### LINKED ISSUE #3253: Stop hardcoding k8up references in core
We currently hardcode the restore CRD in the API https://github.com/uselagoon/lagoon/blob/v2.9.0/node-packages/commons/src/tasks.ts#L945-L968

This should be handled in the `remote-controller` and a more generic payload sent to remote

This is a phased approach, first we need to get support for doing this built into `remote-controller`

shreddedbacon:
# Checklist

To aid in Lagoon supporting a newer version of k8up, we need to strip references to the older CRDs.

This PR removes the `Kind` and `apiVersion` references from the restore, since k8up v1 and v2 restore specs are identical, and the `remote-controller` has supported checking the cluster for specific versions of k8up since `v0.8.0`, this will make it so that the remote-controller will now determine which version of the restore to create based on which version of k8up is detected in the remote clusters.

# Closing issues

closes #3253

### DIFF
--- a/node-packages/commons/src/tasks.ts
+++ b/node-packages/commons/src/tasks.ts
@@ -979,8 +979,6 @@ export const createRemoveTask = async function(removeData: any) {
 // creates the restore job configuration for use in the misc task
 const restoreConfig = (name, backupId, backupS3Config, restoreS3Config) => {
   let config = {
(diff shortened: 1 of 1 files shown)


tobybellwood (approved):
makes sense, tested in remote-controller already

PR #3421, opened 2023-03-28
Docs updates - spacing, lists, indents and install instructions

tobybellwood:
A couple of time-relevant document updates, addition of a new plugin to format lists better, and a number of style and format updates

closes #3420


smlx:
This still seems to have broken formatting on some pages, such as https://docs.lagoon.sh/administering-lagoon/using-harbor/harbor-settings/

Maybe the plugin is making things worse?

tobybellwood:
ah, no, the plugin dislikes double-indented lists - eg
```
* list item
      * list item
```

PR #3422, opened 2023-03-29
WIP: addProblems resolver

timclifford:
We are missing a bulk addProblems resolver, add this ready for upcoming insights changes

### DIFF
--- a/services/api/src/resources/problem/resolvers.ts
+++ b/services/api/src/resources/problem/resolvers.ts
@@ -98,17 +98,15 @@ export const getProblemSources: ResolverFn = async (
 export const getProblemsByEnvironmentId: ResolverFn = async (
   { id: environmentId },
   { severity, source },
-  { sqlClientPool, hasPermission, adminScopes }
+  { sqlClientPool, hasPermission }
 ) => {
   const environment = await environmentHelpers(
     sqlClientPool
   ).getEnvironmentById(environmentId);
 
-  if (!adminScopes.projectViewAll) {
-    await hasPermission('problem', 'view', {
-      project: environment.project
-    });
-  }
+  await hasPermission('problem', 'view', {
+    project: environment.project
+  });
 
   const rows = await query(
     sqlClientPool,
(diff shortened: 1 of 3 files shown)


services/api/src/typeDefs.js
```diff
@@ -295,6 +295,12 @@ const typeDefs = gql`
     created: String
   }
 
+  input AddProblemsInput {
+    project: String
```
bomoko, line 299: Does this project and environment that live at the root level of this input type do anything?
It looks like all the environment details are being determined by the environment ids that appear in the `AddProblemInput`?

It seems like it might be better to and a different input type, replacing `AddProblemInput` to something that had all the appropriate fields but _not_ the environment ID (or the source - more on that in a minute).

That way, you can rely solely on the `project` and `environment` that comes through at the top level - do a permissions check once AND have it resolve by name, rather than by ID.

services/api/src/resources/problem/resolvers.ts
```diff
@@ -200,6 +198,82 @@ export const addProblem: ResolverFn = async (
   return R.prop(0, rows);
 };
 
+export const addProblems: ResolverFn = async (
+  root,
+  { input: { problems } },
+  { sqlClientPool, hasPermission, userActivityLogger }
+) => {
+  const environments = problems.reduce((environmentList, problem) => {
```
bomoko, line 206: As mentioned in my other comment - if we rely on the top level project and environment details, we can get rid of this `reduce`

services/api/src/resources/problem/resolvers.ts
```diff
@@ -200,6 +198,82 @@ export const addProblem: ResolverFn = async (
...
+    }
+
+    let { environment } = problem;
+    if (!environmentList.includes(environment)) {
+      environmentList.push(environment);
+    }
+    return environmentList;
+  }, []);
+
+  for (let i = 0; i < environments.length; i++) {
```
bomoko, line 219: This check could actually be done in the reduce - if you just check whether or not the environment has been added to "environmentList" or not, you could run the check there.

services/api/src/resources/problem/resolvers.ts
```diff
@@ -200,6 +198,82 @@ export const addProblem: ResolverFn = async (
...
+    const env = await environmentHelpers(sqlClientPool).getEnvironmentById(
+      environments[i]
+    );
+    await hasPermission('problem', 'add', {
+      project: env.project
+    });
+  };
+
+  const problemsArr = [];
```
bomoko, line 228: So here's a question - what do we want to do if an entry already exists?

Initially I was thinking of adding the "addProblemsBySource" because what I was planning to do was just nuking the problems for the environment that were in the set of problems marked by that "source".
That way when a new Trivy scan came in, we wouldn't need to reconcile which problems were no longer there.

Actually, this might be something we want to chat a bit about.

bomoko (changes requested):

services/api/src/typeDefs.js
timclifford: Yeah - I wanted to talk through the whole schema for problems since its been a while when we last put this together. But yeah, project isn't needed.

services/api/src/resources/problem/resolvers.ts
timclifford: This is the approach I took in the insights-handler, I delete all problems from source prior to adding new ones. Again though, something to discuss - this PR is currently to just get something up so I can point to this image to be able to add bulk problems

tobybellwood:
@bomoko to ensure this is still good to go!

tobybellwood:
@bomoko do we still need this with the way insights works?

tobybellwood:
CLosing pending problems/facts rewrite

PR #3433, opened 2023-04-13
Fixes inter-resolver call from addFactsByName

bomoko:
Adding the `adminScopes` details in the downstream resolver `getEnvironmentsByProjectId` seems to have broken the `addFactsByName` resolver.

This PR resolves that, as well as cleans up some of the logic
* no intra-resolver calls
* no internal api calls
* Adds a `getEnvironmentsByProjectId` function to the environment helper

# Checklist

### DIFF
--- a/services/api/src/resources/fact/resolvers.ts
+++ b/services/api/src/resources/fact/resolvers.ts
@@ -312,43 +312,37 @@ export const addFactsByName: ResolverFn = async (
   { input: { project, environment, facts } },
   { sqlClientPool, hasPermission, userActivityLogger, keycloakGrant, models, adminScopes }
 ) => {
-  if (project && environment) {
 
-    if (!adminScopes.projectViewAll) {
-      await hasPermission('environment', 'view', {
-        project: project
-      });
-    }
-
-    let lagoonProject = await projectHelpers(sqlClientPool).getProjectIdByName(project);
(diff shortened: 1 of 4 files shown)


services/api/src/resources/fact/resolvers.ts
```diff
@@ -309,29 +310,39 @@ export const addFacts: ResolverFn = async (
...
 
-  const returnFacts = await processAddFacts(facts, sqlClientPool, hasPermission);
+  if (!adminScopes.projectViewAll) {
+    await hasPermission('environment', 'view', {
+      project: project
+    });
+  }
+
+  let lagoonProject = await projectHelpers(sqlClientPool).getProjectIdByName(project);
+  let environments = await environmentHelpers(sqlClientPool).getEnvironmentsByProjectId(lagoonProject);
```
rocketeerbkw, line 327: I think this is not the same thing as calling the `getEnvironmentsByProjectId` resolver because it will included deleted environments. Is it OK/correct to add facts to a deleted environment?

rocketeerbkw (changes requested):

services/api/src/resources/fact/resolvers.ts
bomoko: Good catch!

rocketeerbkw (approved):

PR #3437, opened 2023-04-24
change: add and update additional admin permissions to platform-owner

### LINKED ISSUE #1895: userBySshKey not usable with RBAC
**Describe the bug**
`userBySshKey` is not available for use via the usual RBAC permissions (e.g. by `platform-owner`)
It also is not documented in the RBAC matrix.

**To Reproduce**
1. Log in to Lagoon via SSH or otherwise.
2. Use the token to query `userBySshKey`
3. See error

**Expected behavior**
I would expect `platform-owner` (at least) to be able to query `userBySshKey`, and for this to be documented in the RBAC matrix.

**Screenshots**
n/a

**Additional context**

There are several APIs which have the same issue:
- `userBySshKey`
- `allOpenshifts`
- `allEnvironments`
- `deleteProject`

### LINKED ISSUE #3411: Additional permissions for platform-owner role
Currently the legacy token has more permissions than the platform-owner role does. To reduce the need for users to use a legacy token, we should bring more of the legacy permissions across to the platform-owner role.

There are some that probably don't need to be moved across, like the `x:deleteAll/x:removeAll` permissions.

Off the top of my head, there are these three that need to be added to platform-owner
* `openshift:viewAll`
* `environment:storage`
* `environment:viewAll`

There may be others that I have missed

shreddedbacon:
# Checklist

This grants additional permissions to `platform-owner` role to allow more people with this role to do things. Since we want people to stop having to use the legacy tokens for general usage (internal service usage only for now is fine).

# Closing issues

closes #3411 
closes #1895


rocketeerbkw (commented):
There is one more that would be useful for platform owners which is resource `user` scope `getBySshKey`. Adding that can close #1895 also.

rocketeerbkw (approved):

PR #3443, opened 2023-05-05
fix: create the advanced task permission as it never existed

shreddedbacon:
# Checklist

The Delete Advanced Task permission was added to a function after it had already applied, so never got created.
https://github.com/uselagoon/lagoon/blob/main/services/keycloak/startup-scripts/00-configure-lagoon.sh#L1656-L1666

This pulls out the creation to its own step now


tobybellwood (approved):
this looks more like it!

PR #3447, opened 2023-05-09
Refactor: Refactored raw sql to Knex.js 

CGoodwin90:
This PR refactors some of the raw sql queries to Knex.js - replacing prepared statements, and creating some new queries for reuse.

No new functionality added or removed, but should be a minor performance gain, owing to the removal of "blocking" prepared statements


shreddedbacon (approved):
Looks good to me

PR #3448, opened 2023-05-09
fix: permissions changes for subscriptions after keyloak optimisations

shreddedbacon:
# Checklist

When #3397 was introduced, changes to the context were made, but the subscriptions was missed. This broke subscriptions permissions checks, so they resulted in a weird error that wasn't a permission denied error, but an undefined id error because the current user and all the new information was not passed through to the permission function.

### DIFF
--- a/services/api/src/apolloServer.js
+++ b/services/api/src/apolloServer.js
@@ -120,16 +120,50 @@ const apolloServer = new ApolloServer({
         esClient,
       };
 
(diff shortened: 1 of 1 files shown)


tobybellwood:
woot!

We can now see the subscribers receiving the updated requests
image

tobybellwood (approved):
tested, subscriptions now flowing

PR #3449, opened 2023-05-25
fix: only update deployment fields if they are provided by remote-controller

shreddedbacon:
# Checklist

Minor fix that will only update the remoteid of a deployment if it is provided. In most cases the remote-controller will send this. But if an old build is cancelled, it can't determine the remoteid (because the pod or the build resource may have been deleted from the cluster a long time ago) so it doesn't provide the remoteid

This remoteid is used by the logs retrieval, and when a long time removed build is cancelled via the ui/api, it removes the remoteid which makes the logs inaccessible.

There is no reason to continually update the remoteid of a deployment once it has been set by the build.


bomoko (approved):
Shipit

PR #3453, opened 2023-05-31
fix: validate project key if it is changed, update default-user

shreddedbacon:
# Checklist

When updating a projects private key, there is no validation on it to make sure that the key being changed is valid. This adds validation to the key being updated, and also updates the default-user with the new public key associated to it.

### DIFF
--- a/services/api/src/resources/project/resolvers.ts
+++ b/services/api/src/resources/project/resolvers.ts
@@ -629,8 +629,48 @@ export const updateProject: ResolverFn = async (
 
   const oldProject = await Helpers(sqlClientPool).getProjectById(id);
 
-  // TODO If the privateKey changes, automatically remove the old one from the
+  // If the privateKey is changed, automatically remove the old one from the
   // default user and link the new one.
+  if (patch.privateKey) {
+    let keyPair: any = {};
+    try {
+      const privateKey = R.cond([
+        [R.isNil, generatePrivateKey],
+        [R.isEmpty, generatePrivateKey],
+        [R.T, sshpk.parsePrivateKey]
+      ])(R.prop('privateKey', patch));
+
+      const publicKey = privateKey.toPublic();
+
+      keyPair = {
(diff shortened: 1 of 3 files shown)


services/api/src/resources/project/resolvers.ts
```diff
@@ -629,8 +629,54 @@ export const updateProject: ResolverFn = async (
...
+      };
+
+      const keyParts = keyPair.public.split(' ');
+
+      try {
+        // since public keys can only be associated to a single user
+        // if a polysite tries for some reason uses the same key across all the polysites
+        // then only 1 of the default-users will be able to hold the public key
+        // ideally each project should have its own key and polysite repos should
+        // have the public key for each project added to it for pulls
```
rocketeerbkw: Not sure how valuable this comment is? Projects might share a default user due to past migrations, in addition to a possible method for polysites.

It can live in the commit message to give more context to the change instead?

services/api/src/resources/project/resolvers.ts
```diff
@@ -629,8 +629,54 @@ export const updateProject: ResolverFn = async (
 
   const oldProject = await Helpers(sqlClientPool).getProjectById(id);
 
-  // TODO If the privateKey changes, automatically remove the old one from the
+  // If the privateKey is changed, automatically remove the old one from the
```
rocketeerbkw: We're not removing the old private key in this PR. Is it still a TODO or are we deciding not to do that? Comment should reflect either option.

rocketeerbkw (changes requested):

services/api/src/resources/project/resolvers.ts
shreddedbacon: yeah im not sure. i wrote it because it was the motivation for this change. im happy for it to just live on in this PRs description too if you dont want it in the code base

rocketeerbkw (approved):

PR #3460, opened 2023-06-07
refactor: allow environments to be deleted if the deploytarget is disabled

### LINKED ISSUE #3324: Deleting environments on clusters that no longer exist
Sometimes when a cluster is decommissioned, any environments that are still marked as being on this cluster are unable to be deleted. This is because the API tries to send a message to the remote cluster (which obviously doesn't exist), and then nothing happens.

It would be nice if there was a way to flag a cluster in the API as deleted/decomissioned when they are removed. This issue may still occur even when the cluster is deleted from the API too, but it would probably error silently.

This would allow the API to just mark the environment as deleted, rather than try send a message to a non-existing cluster and hope for a response.

shreddedbacon:
# Checklist

When attempting to delete an environment on a cluster that no longer exists, or has been marked as disabled, Lagoon still tries to send a deletion request which goes unanswered.

Lagoon will now check if the target that the environment is deployed against is disabled, or doesn't exist anymore, and will perform the deletion as requested without sending the deletion request to the target cluster.

# Closing issues

closes #3324

### DIFF
--- a/services/api/src/resources/deployment/resolvers.ts
+++ b/services/api/src/resources/deployment/resolvers.ts
@@ -502,7 +502,7 @@ export const cancelDeployment: ResolverFn = async (
   });
 
   // check if the deploytarget for this environment is disabled
-  const deploytarget = await query(sqlClientPool, Sql.selectDeployTarget(environment.openshift));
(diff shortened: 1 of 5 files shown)


tobybellwood (approved):
tested and LGTM

PR #3462, opened 2023-06-20
refactor: when deleting a user, delete the users keys too

### LINKED ISSUE #2398: ssh key not removed when delete user
# Describe the bug

Deleting a lagoon user should also remove ssh keys.
Now the deletion of a user doesn't touch ssh keys that remain in left-over status.

## To Reproduce

1. Query user and retrieve info about ssh keys
2. Query user by `UserBySshKey` GraphQL query
3. Delete the user via GraphQL or via lagoon-cli
4. Query user by `UserBySshKey` GraphQL query and get the message `"message": "User not found: <id_of_the_user>"`
5. Query the table `ssh_key` of api-db database and check the ssh key is still there 

## Expected behavior

Deletion of ssh keys belonging to the deleted user

shreddedbacon:
# Checklist

Currently if a user is deleted, their SSH keys remain in the database.

For example, if we have John Smith and they have an existing account with `jsmith@example.com` and their single ssh key attached to the user, but Johns account is accidentally deleted. A new account is created, but when John attempts to re-add the key, it fails because the key is already in the database linked to the previous account.

# Closing issues

closes #2398


services/api/src/resources/user/helpers.ts
```diff
@@ -0,0 +1,16 @@
+import { Pool } from 'mariadb';
+import { query } from '../../util/db';
+import { Sql } from './sql';
+// import { logger } from '../../loggers/logger';
+
+export const Helpers = (sqlClientPool: Pool) => {
```
rocketeerbkw: The models were created to supersede the helper files pattern. All this should be moved into to the User model (part of, or called by, `UserModel.deleteUser`) which should be the single place for business logic.

services/api/src/resources/user/sql.ts
```diff
@@ -14,4 +14,10 @@ export default {
       .andWhere('sk.key_type', keyType)
       .select('user_ssh_key.usid')
       .toString(),
+  deleteUserSshKeys: (id: string) =>
```
rocketeerbkw: I think this leaves behind records in the `user_ssh_key` table?

rocketeerbkw (changes requested):

services/api/src/resources/user/sql.ts
shreddedbacon: ah yep

services/api/src/resources/user/helpers.ts
shreddedbacon: rgr

rocketeerbkw (approved):

PR #3466, opened 2023-06-22
Updated input validation to allow Azure Git Repositories to pass

### LINKED ISSUE #1743: Azure git repository names don't validate
**Describe the bug**
Azure Git Repositories don't pass the input validation

`git@ssh.dev.azure.com:v3/ORG/SUBPROJECT/REPONAME`

The regex https://github.com/amazeeio/lagoon/blob/027f9a80cc40699de99c70f43bac1453c478ad7d/services/api/src/resources/project/resolvers.js#L32

Seems not to let the Azure Repositories being added - the graphql mutations fail with `The provided gitUrl is invalid`

**To Reproduce**
1. Update a test project with the repository url `git@ssh.dev.azure.com:v3/ORG/SUBPROJECT/REPONAME`

**Expected behavior**
Regex is passed

CGoodwin90:
This PR updates the current artisanal regex to use a reputable git URL parser to verify that a URL parses.

Successfully created test projects using the Git URL patterns mentioned in #1743 

image

Closes #1743

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,5 @@
   },
   "scripts": {
     "glow": "glow check"
-  },
-  "dependencies": {
-    "git-url-parse": "^13.1.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -23,5 +23,8 @@
   },
   "scripts": {
     "glow": "glow check"
+  },
+  "dependencies": {
+    "git-url-parse": "^13.1.0"
   }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -23,5 +23,8 @@
   },
   "scripts": {
     "glow": "glow check"
+  },
+  "dependencies": {
```
tobybellwood: we could probably push this into the api package.json at `services/api/package.json` in case it's of use elsewhere (and avoid adding dependencies to too many packages if we can avoid it?

CGoodwin90: Good call, updated now.

rocketeerbkw (approved):

PR #3471, opened 2023-06-28
refactor: check if retries are required on some errors

shreddedbacon:
# Checklist

If the lagoon-api is unresponsive, sometimes the actions-handler is unable to update properly. This adds error checking that if a connection reset is hit (api down) then it will flag an error and the message will remain in the queue to be retried again.

An example of a connection reset is below. This message was dropped, which means the status of this deployment may not have been updated correctly because the message is dropped when this error is hit.
```
(messageid:pqt3csxj) solr-service-qa/lagoon-build-do90w: ERROR: unable to get deployment - Post "http://lagoon-core-api:80/graphql": read tcp 10.204.12.70:49118->10.208.4.225:80: read: connection reset by peer
```


bomoko (approved):
This makes sense and looks good.

Just one thing, I wonder if there  could be a case where this becomes a choke point?

shreddedbacon:
> Just one thing, I wonder if there could be a case where this becomes a choke point?

It is possible, but only if the errors that are returned are consistently reset by peer errors. Which from analysing logs only happens during scaling events where the API pod becomes unavailable mid request, that is basically the protection this is covering for.

bomoko:
sweet, happy happy - LGTM and all that jazz.

PR #3481, opened 2023-07-06
feat: support shared bucket for backups

### LINKED ISSUE #3444: Support Shared S3 Bucket for K8up Backups
## Context

AWS has hard limit on the number of S3 buckets one can create within an account (1000). Lagoon sets the backup bucket name on a project level (`baas-${project}`) and offloads the credentials and bucket creation to K8up on the Kubernetes cluster. With this approach a cluster can not contain over 1000 projects without configuring custom backup locations on a project level.

## Idea

K8up uses Restic under the hood which supports multiple Restic repositories within the same S3 bucket using different paths. This can be leveraged by introducing an (opt-in) option to configure a shared S3 bucket on a Lagoon Remote (~K8s cluster) level.

### Rough Example
Old:
```
# S3 bucket name supplied to K8up
baas-project-a
baas-project-c
```

New:
```
# Config for Lagoon Remote
baasSharedBucketName: cluter-x-shared-bucket

# S3 bucket name supplied to K8up
cluter-x-shared-bucket/baas-project-a
cluter-x-shared-bucket/baas-project-c
```

```
---
apiVersion: backup.appuio.ch/v1alpha1
kind: Schedule
spec:
  backend:
    s3:
      bucket: cluter-x-shared-bucket/baas-project-a
[...]
```

### Additional Information

AFAIK Lagoon Core needs to the know the full bucket name to start a restore. Otherwise it might be possible to implement this only within Lagoon Remote (https://github.com/uselagoon/build-deploy-tool)

shreddedbacon:
# Checklist

As described in #3444 there is a limit on how many buckets there can be in S3.

This changes the behaviour of `new` projects to use a bucket derived from the deploytarget name, or from a defined `sharedBaasBucketName` field on the deploytarget.

There is currently a `LAGOON_BAAS_BUCKET_NAME` override variable that can also be defined in a project, so this still overrides the shared bucket even if the project is `new`. To revert to the shared bucket in this case is to simply delete the `LAGOON_BAAS_BUCKET_NAME` variable.

The change will now create the k8up path like so.
```
${sharedBaasBucketName}/baas-${projectName}
```

To convert an existing project to use the shared bucket, you will need to check for the existence of the override variable and remove it, and then update the project in the Lagoon API to set the projects `sharedBaasBucket=true` (platform-owner role required). This has the negative effect that it basically makes all the existing backups unusable. So transferring the data somehow will need to be done if you wish to retain those backups. A deployment will also need to be performed after setting the `sharedBaasBucket` value.

This requires https://github.com/uselagoon/build-deploy-tool/pull/219 to work

# Closing issues

closes #3444

### DIFF
--- a/services/api/src/resources/project/resolvers.ts
+++ b/services/api/src/resources/project/resolvers.ts
@@ -785,7 +785,8 @@ export const updateProject: ResolverFn = async (
         openshift,
         openshiftProjectPattern,
         developmentEnvironmentsLimit,
-        buildImage
+        buildImage,
+        sharedBaasBucket
       }
     })
   );
@@ -887,7 +888,8 @@ export const updateProject: ResolverFn = async (
         deploymentsDisabled,
         pullrequests,
         developmentEnvironmentsLimit,
-        buildImage
+        buildImage,
+        sharedBaasBucket
       }
     }
   });
--- a/services/api/src/resources/project/resolvers.ts
+++ b/services/api/src/resources/project/resolvers.ts
@@ -328,6 +328,15 @@ export const addProject = async (
     }
   }
 
+  let sharedBaasBucket = null;
+  if (input.sharedBaasBucket === false || input.sharedBaasBucket === true) {
+    if (adminScopes.projectViewAll) {
+      sharedBaasBucket = input.sharedBaasBucket
+    } else {
+      throw new Error('Setting shared baas bucket is only available to administrators.');
+    }
+  }
+
   const osRows = await query(sqlClientPool, OS.Sql.selectOpenshift(openshift));
   if(osRows.length == 0) {
     throw Error(`Openshift ID: "${openshift}" does not exist"`);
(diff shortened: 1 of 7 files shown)


services/api/src/resources/project/resolvers.ts
```diff
@@ -611,6 +621,12 @@ export const updateProject: ResolverFn = async (
     }
   }
 
+  if(typeof sharedBaasBucket == "boolean") {
+    if (adminScopes.projectViewAll) {
+      throw new Error('Setting shared baas bucket is only available to administrators.');
+    }
+  }
+
```
rocketeerbkw, line 629: The PR description and this authz check suggest that updating `sharedBaasBucket` should be possible, but there are still some missing pieces to complete that: The update/patch types for project need to be updated in `typeDefs.js`; The `sql.updateProject` call needs to get passed the input later on in this function, and the argument needs to get added to the function definition in `project/sql.ts`.

rocketeerbkw (changes requested):

services/api/src/resources/project/resolvers.ts
shreddedbacon: Ah thanks

rocketeerbkw (approved):
LGTM

PR #3489, opened 2023-07-26
fix: nack messages that encounter error to resubmit

shreddedbacon:
# Checklist

This fixes the action-handler so that it will use `nack` on messages that need to resubmit to the queue for processing if there is an error in the Lagoon API


tobybellwood (approved):
Tested as working

PR #3497, opened 2023-08-01
feat: support changing more settings in keycloak

shreddedbacon:
# Checklist

This adds the ability to configure additional realm settings easily using json files, and also set the admin user email address which seems to be required for configuring email.

There does exist an existing way to apply realm settings using an envvar, but it is cumbersome. This new way allows helm to create a configmap to then volumemount to the required location and be consumed.

I've separated out the smtp configuration to its own step so that the email configuration can be provided separately to other realm settings easily, but they still use the same endpoint in the end.


tobybellwood (approved):
Tested working

PR #3507, opened 2023-08-07
feat: add delete organization and other removal operation functions

shreddedbacon:
# Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for inclusion in changelog

This adds new resolvers to support organizations:

* deleteOrganization - you could add one, now you can delete one
* removeDeployTargetFromOrganization - you could add one, now you can remove one
* removeProjectFromOrganization - you could add one, now you can remove one (it removes all groups and notifications from the project)

A bunch of new helpers were written to replace existing code and reduce some duplication.

PR #3509, opened 2023-08-07
Introducing removeUserFromOrganizationGroups resolver

timclifford:
For use in situations where an organisation admin wants to be able to remove a user from all groups in the org. Linked to https://github.com/uselagoon/lagoon-ui/pull/142


services/api/src/resources/organization/resolvers.ts
```diff
@@ -765,4 +766,69 @@ export const addGroupToOrganization: ResolverFn = async (
...
+      });
+    }
+
+    try {
+      await models.GroupModel.removeUserFromGroup(user, orgGroups[group]);
+      groupsRemoved.push(orgGroups[group].name);
+    } catch (error) {
+      throw new Error(`Unable to remove user from group: ${error}`)
+    }
+  }
```
shreddedbacon: I'd probably change this up a bit and do one permission check, then work out the groups the user is in that are in the organization, then remove them from those groups only.

```
  await hasPermission('organization', 'removeGroup', {
    organization: organizationInput,
  });
  const orgGroups = await models.GroupModel.loadGroupsByOrganizationId(organizationInput);

  // iterate through groups and work out which of this users groups are in the organization
  let groupsRemoved = []
  for (const group in orgGroups) {
    // if the groups organization is the one to remove from, push it to a new array
    if (R.prop('lagoon-organization',  orgGroups[group].attributes) == organizationInput) {
      groupsRemoved.push(orgGroups[group]);
    }
  }
  // iterage over the groups to remove now (or use a model helper that takes the `groupsRemoved` as an input instead of doing the for loop here one by one, see note below)
  for (const group in groupsRemoved) {
    try {
      await models.GroupModel.removeUserFromGroup(user, groupsRemoved[group]);
    } catch (error) {
      throw new Error(`Unable to remove user from group: ${error}`)
    }
  }
```

What could be even better though is using a group model helper similar to this https://github.com/uselagoon/lagoon/blob/delete-organization/services/api/src/models/group.ts#L744-L785 (which is from https://github.com/uselagoon/lagoon/pull/3507) that would remove the user from all provided groups, and then update the redis keycloak cache once complete

shreddedbacon (changes requested):

services/api/src/resources/organization/resolvers.ts
timclifford: https://github.com/uselagoon/lagoon/pull/3509/commits/b772bc0a71d598bc0eda8937b8b68e909ca92f21

services/api/src/models/group.ts
```diff
@@ -833,6 +834,42 @@ export const Group = (clients: {
...
+
+      const allGroups = await loadAllGroups();
+      const keycloakGroups = await transformKeycloakGroups(allGroups);
+      const data = Buffer.from(JSON.stringify(keycloakGroups)).toString('base64')
+      try {
+        // then attempt to save it to redis
+        await saveRedisKeycloakCache("allgroups", data);
+      } catch (err) {
+        logger.warn(`Couldn't save redis keycloak cache: ${err.message}`);
+      }
```
shreddedbacon: This will run on every group that the user is removed from. This query is quite intense.
Would be better to run it after the user has been removed from all groups

shreddedbacon (changes requested):

services/api/src/resources/organization/resolvers.ts
```diff
@@ -871,6 +872,64 @@ export const addGroupToOrganization: ResolverFn = async (
...
+  const organizationData = await Helpers(sqlClientPool).getOrganizationById(organizationInput);
+  if (organizationData === undefined) {
+    throw new Error(`Organization does not exist`)
+  }
+
+  // check permissions and get groups
+  await hasPermission('organization', 'removeGroup', {
+    organization: organizationInput,
+  });
+  const orgGroups = await models.GroupModel.loadGroupsByOrganizationId(organizationInput);
```
shreddedbacon: And change this to use this helper `models.GroupModel.loadGroupsByOrganizationIdFromGroups`, there is an example of how to use it here https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/organization/resolvers.ts#L325-L331

services/api/src/models/group.ts
timclifford: Good spot - updated

shreddedbacon (approved):
Yep, ran this locally and looks good

PR #3511, opened 2023-08-10
refactor: improve the query for deploymentsbyfilter for platform users

shreddedbacon:
# Checklist

This changes the query for `deploymentsByFilter` to only add the project join when projects are requested. This improves the response on the query considerably by not doing the project join for no reason.

```
> select deployment.* from deployment join environment on deployment.environment = environment.id join project on environment.project = project.id where environment.deleted = '0000-00-00 00:00:00' and deployment.status in ("NEW", "PENDING", "RUNNING", "QUEUED");
<SNIP>
8 rows in set (25.832 sec)
```
versus this when not including the project join for no reason
```
> select deployment.* from deployment join environment on deployment.environment = environment.id where environment.deleted = '0000-00-00 00:00:00' and deployment.status in ("NEW", "PENDING", "RUNNING", "QUEUED");
<snip>
8 rows in set (2.333 sec)
```

### DIFF
--- a/services/api/src/resources/deployment/resolvers.ts
+++ b/services/api/src/resources/deployment/resolvers.ts
@@ -193,21 +193,19 @@ export const getDeploymentsByFilter: ResolverFn = async (
   }
 
   let queryBuilder = knex.select("deployment.*").from('deployment').
-      join('environment', 'deployment.environment', '=', 'environment.id').
-      join('project', 'environment.project', '=', 'project.id').
-      where('environment.deleted', '=', '0000-00-00 00:00:00');
+      join('environment', 'deployment.environment', '=', 'environment.id');
+
+  if (userProjectIds) {
+      queryBuilder = queryBuilder.whereIn('environment.project', userProjectIds);
+  }
 
   if(openshifts) {
     queryBuilder = queryBuilder.whereIn('environment.openshift', openshifts);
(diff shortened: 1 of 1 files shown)


services/api/src/resources/deployment/resolvers.ts
```diff
@@ -193,21 +193,20 @@ export const getDeploymentsByFilter: ResolverFn = async (
   }
 
   let queryBuilder = knex.select("deployment.*").from('deployment').
-      join('environment', 'deployment.environment', '=', 'environment.id').
-      join('project', 'environment.project', '=', 'project.id').
-      where('environment.deleted', '=', '0000-00-00 00:00:00');
+      join('environment', 'deployment.environment', '=', 'environment.id');
+
+  if (userProjectIds) {
+      queryBuilder.join('project', 'environment.project', '=', 'project.id').
```
bomoko: Actually, do we even pull anything from `project`?

We might want to kill this join directly, don't you think? Then we just say (below) 
`queryBuilder = queryBuilder.whereIn('environment.project', userProjectIds);`

shreddedbacon: You're right, we don't need to join at all. Will update

bomoko (approved):

Issue #3519, opened 2023-08-14
Logging of every GraphQL Request


Schnitzel:
Currently the API is logging every mutation that a user is doing.
While this is great we don't have logging on the requests that are reading only.

this can be a quite easy log with these parameters:
- GraphQL Query
- GraphQL Parameters
- User that does the query

timclifford:
https://github.com/uselagoon/lagoon/pull/3520

PR #3520, opened 2023-08-14
User logs for all read queries

timclifford:
Re: https://github.com/uselagoon/lagoon/issues/3519

### DIFF
--- a/services/api/src/apolloServer.js
+++ b/services/api/src/apolloServer.js
@@ -284,7 +284,7 @@ const apolloServer = new ApolloServer({
         hasPermission,
         keycloakGrant,
         requestCache,
-        userActivityLogger: (message, meta) => {
+        userActivityLogger: (message, meta, level = 'user_action') => {
           let defaultMeta = {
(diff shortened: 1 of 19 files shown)


Schnitzel:
@timclifford thanks very much for this, this is super great, couple of questions:

1. in my understanding this logs each individual query of a full graphql query, which is great. But would it also be possible to log the original graphql query itself?
2. Is there any possibility to get some timing data for each of these individual queries? This might be too complex and happy to ditch this for the first version, but was wondering if this is maybe easy to add

Schnitzel:
Discussed with Lagoon team
- we would want a new `meta.level` field for these read queries: `user_query` to be able to easily filter for them in the logs.
- newrelic provides us timing information already
- @bomoko will work on logging the whole query

shreddedbacon:
It might be an idea to wrap these in a feature, so it can be enabled/disabled as required?

tobybellwood:
To be addressed in a Lagoon API audit logging overhaul

PR #3533, opened 2023-08-22
Adds bb sync custom redis env var support

### LINKED ISSUE #3532: Allow to define specific Redis to use for Bitbucket Sync (fallback to existing redis if not defined)
idea would be to have the possibility to define these env variables:

```
BITBUCKETSYNC_REDIS_PASSWORD
BITBUCKETSYNC_REDIS_PORT
BITBUCKETSYNC_REDIS_HOST
```

if they are defined that redis is used, if not we fall back to just the `REDIS_HOST` env variables.

bomoko:
This PR allows the setting of a custom redis for the bitbucket script as described in #3532 

closes #3532


tobybellwood:
No longer required

PR #3534, opened 2023-08-23
refactor: use keycloak cache

shreddedbacon:
# Checklist

The list all users groups function doesn't return all the attributes of the groups, so it was written to also consume all the groups and then iterate over them to get the attributes. This unfortunately had a cache bypass in there, this corrects it to consume from the cache instead of querying keycloak directly.


tobybellwood (approved):
tested working

PR #3542, opened 2023-09-07
refactor: wrap drush commands in error capture

### LINKED ISSUE #3541: Lagoon sql-dump task should inherit the exit code of the drush command
Spotted on a `drush sql-dump` task, if the command fails (in this case due to corrupted data)

```
================================================================================
Logs on pod lagoon-task-XXXXX, assigned to node ip-XX-X-XXX-XX.ap-southeast-2.compute.internal on cluster blah
================================================================================
Agent pid 18
Identity added: /home/.ssh/key (/home/.ssh/key)
> mysqldump: Error 2013: Lost connection to server during query when dumping table `queue` at row: 553

In SqlCommands.php line 242:
                                                                         
  Unable to dump database. Rerun with --debug to see any error message.  
```

The task status is still successful

image

This can lead to a false sense of confidence that everything worked (when in fact it did not).

## Ideal solution

If the sql-dump fails, then the task should report failure too.

From reading the [code](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/task/resolvers.ts#L548-L555), perhaps it is the chaining of commands together that causes the `drush` exit code to be lost?

shreddedbacon:
# Checklist

If the drush command fails, tasks still shows as complete. This wraps the drush commands in a statement to capture the errors so that tasks are marked as failed properly. It doesn't address the whole failure cases though as it could fail generating the token, and also during the file upload step. Lagoon CLI has a possible solution to handling the tokengeneration and file upload steps using https://github.com/uselagoon/lagoon-cli/pull/275, as it would capture errors in both the generation and upload steps

closes #3541

### DIFF
--- a/services/api/src/resources/task/resolvers.ts
+++ b/services/api/src/resources/task/resolvers.ts
@@ -497,7 +497,7 @@ export const taskDrushArchiveDump: ResolverFn = async (
     project: envPerm.project
   });
 
-  const command = String.raw`file="/tmp/$LAGOON_PROJECT-$LAGOON_GIT_SAFE_BRANCH-$(date --iso-8601=seconds).tar" && drush ard --destination=$file && \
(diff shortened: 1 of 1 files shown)


tobybellwood (approved):
makes sense

PR #3543, opened 2023-09-08
fix: reset exists check at the end of the loop

shreddedbacon:
# Checklist

There was a condition where the user already exists check resulted in the `true` statement being retained, preventing new users that aren't already counted from being counted.

This fixes it by resetting the exists check at the end of the user loop check


tobybellwood (approved):
nice fix

PR #3545, opened 2023-09-12
fix: return error if project not found

### LINKED ISSUE #3547: Better error message when trying to get a project that does not exist
e.g.

```
query GetProjectShort {
  projectByName(
    name: "does-not-exist"
  ) {
    id
  }
}
```

the result is 

```
{
  "errors": [
    {
      "message": "Cannot read properties of undefined (reading 'id')",
      "locations": [
        {
          "line": 6,
          "column": 3
        }
      ],
      "path": [
        "projectByName"
      ]
    }
  ],
  "data": {
    "projectByName": null
  }
}
```

This is not a very helpful error message, and it wastes precious time from a DX perspective, as you think your GraphQL query might be wrong.

## Ideal solution

Error message is something like

> "The Project `does-not-exist` does not exist in this Lagoon"

shreddedbacon:
# Checklist

If you query the API for a project by name and the project does not exist, you get a strange error that makes it difficult to know what went wrong. This fixes it to return a project not found error instead

Previous error, what does it mean?
```
{
  "errors": [
    {
      "message": "Cannot read properties of undefined (reading 'id')",
      "locations": [
        {
          "line": 100,
          "column": 3
        }
      ],
      "path": [
        "projectByName"
      ]
    }
  ],
  "data": {
    "projectByName": null
  }
}
```

closes #3547

### DIFF
--- a/services/api/src/resources/project/resolvers.ts
+++ b/services/api/src/resources/project/resolvers.ts
@@ -208,7 +208,9 @@ export const getProjectByName: ResolverFn = async (
   const project = withK8s[0];
 
   if (!project) {
-    return null;
+    throw new Error(
+      `No project matching ${project.name} found`
+    );
   }
 
   try {


shreddedbacon:
Closing this as the error is actually coming from something unrelated from a current hotfix applied to amazeeio infrastructure.

PR #3554, opened 2023-09-28
Feature/OrganizationByName resolver

CGoodwin90:
# General Checklist

Implements the `organizationByName` resolver


services/api/src/resources/organization/resolvers.ts
```diff
@@ -242,6 +242,30 @@ export const getOrganizationById: ResolverFn = async (
...
+  args,
+  { sqlClientPool, hasPermission }
+) => {
+  let orgName = args.organization;
+  if (organization) {
+    orgName = organization;
+  }
+
+  await hasPermission('organization', 'view', {
+    organization: orgName,
```
shreddedbacon: this won't work as the attribute has to be the organization id, not the name

you'll need to run the sql query prior to the haspermission check to retrieve the `orgResult.id` to pass to haspermissions

shreddedbacon (changes requested):

services/api/src/resources/organization/resolvers.ts
shreddedbacon: basically shift all of this above haspermissions
```
  const rows = await query(sqlClientPool, Sql.selectOrganizationByName(orgName));
  const orgResult = rows[0];

  if (!orgResult) {
    return null;
  }
```

CGoodwin90: Ah yeah of course, done!

services/api/src/typeDefs.js
```diff
@@ -1416,6 +1416,7 @@ const typeDefs = gql`
     Get an organization by its ID
     """
     organizationById(organization: Int!): Organization
+    organizationByName(organization: String!): Organization
```
shreddedbacon: Might be better to use `name: String!` instead here to reduce confusion.

I'm also seeing that `organizationById` should probably use `id` instead of `organization` but that's a different issue

shreddedbacon: The resolver will need to be updated too

CGoodwin90: Yep makes sense, updated now.

shreddedbacon (approved):

PR #3567, opened 2023-10-13
fix: check environment doesnt exist before checking env quotas

shreddedbacon:
# General Checklist

# Database Migrations

Fixes the environment quota check to make sure that an environment that already exists isn't evaluated to check if the quota is exceeded, and to allow it through as normal.

Also modifies the example organization deployed in local dev with a 4 env quota to validate


tobybellwood (approved):
Straightforward change. A++ for test consideration

PR #3568, opened 2023-10-15
fix: throw an error on deployment errors triggered via the api

shreddedbacon:
# General Checklist

# Database Migrations

# Closing issues

* https://github.com/uselagoon/lagoon-cli/issues/189

### DIFF
--- a/services/api/src/resources/deployment/resolvers.ts
+++ b/services/api/src/resources/deployment/resolvers.ts
@@ -741,7 +741,7 @@ export const deployEnvironmentLatest: ResolverFn = async (
       break;
 
     default:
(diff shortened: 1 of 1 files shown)


shreddedbacon:
This is just a place holder, probably not to be merged because unsure how far this possibly breaks

tobybellwood:
As odd as the logic may seem, we should flag this as a deprecation to change in a future release.

"We will be deprecating the returning of these as error strings, in favour of returning an actual error"

tobybellwood (approved):
Happy to move forwards with these. Will post a clear message in the release notes that errors are now actually errors. Will go through all internal tools to ensure they work as expected

PR #3579, opened 2023-10-25
update docker-compose-v2 docs

tobybellwood:
Updates the documentation around Docker Compose v2 and Buildkit to reflect the upstream changes and compatibility


rocketeerbkw (approved):
Keeping the error in the docs means users can still search for them to find the upgrade instructions :+1:

PR #3581, opened 2023-10-27
docs: NFS-provisioner updated docs

tobybellwood:
This PR updates the EFS provisioner docs to a more recent example, with more generic settings.


shreddedbacon (approved):
This is good to me

PR #3583, opened 2023-11-01
fix: purge the cached id of the group when group is deleted

shreddedbacon:
# General Checklist

# Database Migrations

When a group is deleted in Lagoon, there is a cached item that stores the group-id against the group-name. This allows us to recall information from keycloak quicker by using the ID, rather than a name based search and filtering method.

Unfortunately, if you delete a group, this cached ID remains, and subsequently re-creating the group before the cache expires, results in potential issues with trying to do things against the wrong group ID.

This fixes the issue by purging the cached ID when a group is deleted.


tobybellwood (approved):
Tested, able to reproduce fixed issue, change has minimal impact otherwise.

PR #3586, opened 2023-11-03
Update dockerfile builds, and consolidate broker images.

tobybellwood:
This tidies up a few services that are no longer needed:
* local-registry, local-dbaas-provider and local-mongodb-dbaas-provider aren't needed for a docker compose stack
* the broker and broker-single dockerfiles have been combined into one single image
* the configuration in the rabbitmq/broker image for clustering is no longer supported in rabbitmq (classic mirror) so has been removed from the image
* the git and data-watcher-pusher images are now built from commons instead of alpine for ease of update.


services/broker/start-rabbit.sh
```diff
@@ -11,18 +11,19 @@ if [ -e /etc/rabbitmq/definitions.json ]; then
...
 # Check if the container runs in Kubernetes/OpenShift
 if [ -z "$POD_NAMESPACE" ]; then
   # Single container runs in docker
-  echo "POD_NAMESPACE not set, spin up single node"
-  exec docker-entrypoint.sh rabbitmq-server
+  echo "running in Docker, spin up single node"
+else
+  # Single container runs in kubernetes
+  echo "running in Kubernetes, spin up single node"
 fi
```
shreddedbacon, line 26: is this even required? nothing changes whether it is in docker or kubernetes here

shreddedbacon (approved):
This looks fine to me. Running locally everything seems operational as normal.

Tests are passing in Jenkins too so :rocket: 

Would be good to figure out if we can benefit from using a different broker entirely in the future if we don't make the NATS transition happen anytime soon.

services/broker/start-rabbit.sh
tobybellwood: no, not required, but whilst we find a way to start rabbit generically, it can stay

PR #3591, opened 2023-11-12
Update the data in the data-watcher-pusher, and rejig the process slightly

tobybellwood:
This PR introduces a more complete set of data created by the data-watcher pusher, suitable for use in front-end testing.

It also converts the api-data-watcher-pusher from an always-watching process to a one-off job.

It also takes the opportunity to align versions of kubernetes and helm with lagoon-charts


shreddedbacon:
Would be good to review these changes before they are merged, and also squash commits :)

tobybellwood:
yah - my bad on the squash - there's a whole house of cards behind this PR (hence the early merge) - but happy to revisit withany updates

PR #3598, opened 2023-11-19
feat: add new project and group importer for organizations

shreddedbacon:
# General Checklist

# Database Migrations

This adds a new query and mutation for helping with importing a project, and all associated groups into an organization.

The new query is called `checkBulkImportProjectsAndGroupsToOrganization` and it will return lists of all the projects and groups that would be imported into the organization. The payload allows an administrator to verify that the correct projects and groups would be imported. Giving the administrator time to verify the imported projects and groups and either remove, or add anything that may be missing, then run the query again.

This query also returns a second list of projects and groups, but it will contain anything in there that indicates a project or group has already been moved into a completely different organization, which would result in the import failing.

The new mutation is called `bulkImportProjectsAndGroupsToOrganization` and will perform the changes that `checkBulkImportProjectsAndGroupsToOrganization` outputs. There is the option of also detaching all notifications from all projects within the import. Otherwise existing notifications will remain attached to the projects, but the organization owners will only ever be able to remove these from their projects, because the main notification object will be owned by the platform (the way notifications always were before). The platform defined notifcations themselves will NOT be imported, as there could be the possibility that this is shared across multiple organizations, which is not supported by the organization model.

It marks all the previous importing queries and mutations as deprecated, and their usage is discouraged.

### DIFF
--- a/services/api/src/typeDefs.js
+++ b/services/api/src/typeDefs.js
@@ -1417,9 +1417,17 @@ const typeDefs = gql`
     """
     organizationById(id: Int!): Organization
     organizationByName(name: String!): Organization
-    getGroupProjectOrganizationAssociation(input: AddGroupToOrganizationInput!): String
-    getProjectGroupOrganizationAssociation(input: ProjectOrgGroupsInput!): String
+    getGroupProjectOrganizationAssociation(input: AddGroupToOrganizationInput!): String  @deprecated(reason: "Use checkBulkImportProjectsAndGroupsToOrganization instead")
+    getProjectGroupOrganizationAssociation(input: ProjectOrgGroupsInput!): String  @deprecated(reason: "Use checkBulkImportProjectsAndGroupsToOrganization instead")
     getEnvVariablesByProjectEnvironmentName(input: EnvVariableByProjectEnvironmentNameInput!): [EnvKeyValue]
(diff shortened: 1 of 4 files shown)


services/api/src/typeDefs.js
```diff
@@ -2446,6 +2454,13 @@ const typeDefs = gql`
     Remove a deploytarget from an organization
     """
     removeDeployTargetFromOrganization(input: RemoveDeployTargetFromOrganizationInput): Organization
+    """
+    Run the query getProjectGroupOrganizationAssociation first to see the changes that would be made before executing this, as it may contain undesirable changes
```
tobybellwood:
```suggestion
    Run the query checkBulkImportProjectsAndGroupsToOrganization first to see the changes that would be made before executing this, as it may contain undesirable changes
```

tobybellwood:
Sample check query:
```
query checkBulkImportProjectsAndGroupsToOrganization {
	checkBulkImportProjectsAndGroupsToOrganization(
		input: { project: 1, organization: 1 }
	) {
		projects {id, name}
		groups { id, name }
		otherOrgProjects { id, name }
		otherOrgGroups { id, name }
	}
}
```

tobybellwood (approved):
Works as advertised, tested against a few edge cases

PR #3600, opened 2023-11-21
add CLEAR_API_DATA env to allow API data clearing

tobybellwood:
This PR adds an optional CLEAR_API_DATA into the data watcher pusher that triggers the delete_all_* mutations.

In normal usage this isn't required, but can come in handy when rerunning tests (make k3d/retest) after failures.

Longer term goal is to make the ansible tests more fault tolerant, but let's eat the elephant one bite at a time...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -629,6 +629,7 @@ k3d/retest:
 			LAGOON_FEATURE_FLAG_DEFAULT_ISOLATION_NETWORK_POLICY=enabled \
 			USE_CALICO_CNI=false \
 			LAGOON_FEATURE_FLAG_DEFAULT_ROOTLESS_WORKLOAD=enabled \
+			CLEAR_API_DATA=true \
 		&& docker run --rm --network host --name ct-$(CI_BUILD_TAG) \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -629,6 +629,7 @@ k3d/retest:
 			LAGOON_FEATURE_FLAG_DEFAULT_ISOLATION_NETWORK_POLICY=enabled \
 			USE_CALICO_CNI=false \
 			LAGOON_FEATURE_FLAG_DEFAULT_ROOTLESS_WORKLOAD=enabled \
+			CLEAR_API_DATA=true \
```
shreddedbacon: I'd like to see this changed to set `CLEAR_API_DATA` as a make var with the default `true` similar to other options (https://github.com/uselagoon/lagoon/blob/testing/fix_dwp_rerun/Makefile#L57)

Then update this to use the make var, this way its possible to override it in the command line to `false` if required.
```
CLEAR_API_DATA=$(CLEAR_API_DATA)
```

shreddedbacon (changes requested):

shreddedbacon (approved):
Looks good to me

PR #3604, opened 2023-11-28
refactor: replace only the updated group in the cache

shreddedbacon:
# General Checklist

# Database Migrations

# Closing issues

### DIFF
--- a/services/api/src/models/group.ts
+++ b/services/api/src/models/group.ts
@@ -9,7 +9,7 @@ import { logger } from '../loggers/logger';
 // @ts-ignore
 import GroupRepresentation from 'keycloak-admin/lib/defs/groupRepresentation';
 import { User } from './user';
-import { saveRedisKeycloakCache, get, redisClient } from '../clients/redisClient';
+import { getRedisKeycloakCache, saveRedisKeycloakCache, get, redisClient, redlock, initialLockDuration, extendLockDuration } from '../clients/redisClient';
 import { Helpers as projectHelpers } from '../resources/project/helpers';
 import { sqlClientPool } from '../clients/sqlClient';
 import { log } from 'winston';
@@ -547,6 +547,7 @@ export const Group = (clients: {
             name: group.name
           }
         );
+        await purgeGroupInCache(parentGroup.id)
       } catch (err) {
         if (err instanceof GroupNotFoundError) {
           throw new GroupNotFoundError(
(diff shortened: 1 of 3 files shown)


services/api/src/models/group.ts
```diff
@@ -922,25 +842,72 @@ export const Group = (clients: {
...
-    // once the users are removed from the group, update the cache
-    const allGroups = await loadAllGroups();
-    const keycloakGroups = await transformKeycloakGroups(allGroups);
-    const data = Buffer.from(JSON.stringify(keycloakGroups)).toString('base64')
+    return await loadGroupById(group.id);
+  };
+
+  const purgeGroupInCache =async (groupId: string) => {
+    // lock the redis interactions
+    let lock = await redlock.acquire(["purgelock"], 5000);
```
bomoko: I reckon the number 5000 on this line and on the lock extend should at least be a constant - like a `lockRequestSeconds` or whatever (absolutely _not_ suggesting that as the name :D ).

services/api/src/models/group.ts
```diff
@@ -922,25 +842,72 @@ export const Group = (clients: {
...
+      }
+      let fixedKeycloakGroups = keycloakGroups
+      try {
+        // try and replace or update the group within the cache
+        const group = await loadGroupById(groupId);
+        let exists = false;
+        for (const g of keycloakGroups) {
+          if (g.id == groupId){
+            exists = true
+            fixedKeycloakGroups = keycloakGroups.map(item => item.id === groupId ? group : item)
```
bomoko, line 893: There might be more idiomatic ways of doing this for loop stuff (I've got some suggestions if you want to go into it, but not necessary).

Minimally, I'd add here, though, would be a "break" in this `if` statement so that once the group is matched you don't need to spin through the rest of the items. (if `g.id == groupId` and the loop index is 1 out of like 10K, you're doing a lot of extra cycles).

services/api/src/models/group.ts
```diff
@@ -922,25 +842,72 @@ export const Group = (clients: {
...
-    } catch (err) {
-      logger.warn(`Couldn't save redis keycloak cache: ${err.message}`);
+      try {
+        await saveRedisKeycloakCache("allgroups", data);
+      } catch (err) {
+        logger.warn(`Couldn't save redis keycloak cache: ${err.message}`);
+      }
+    } finally {
+      // release the lock
+      await lock.release();
```
bomoko, line 914: Just a general question - are there any other places where we need to worry about locking the redis kc cache?

bomoko (changes requested):

services/api/src/models/group.ts
shreddedbacon: Yup, makes sense

services/api/src/models/group.ts
shreddedbacon: Happy if you want to refactor it out, either comment here or directly in the branch. I don't mind

services/api/src/models/group.ts
shreddedbacon: Most other reads/writes are to different keys, and they're not read or written to very often, I think this one makes the most sense mainly because of how special the data in it is

services/api/src/models/group.ts
shreddedbacon: Added the break in the loop if the groupid is found for now

bomoko (approved):

shreddedbacon:
closing for #3612

PR #3606, opened 2023-11-29
introduce DEPRECATIONS.md to track deprecations

tobybellwood:
This PR introduces a central file, DEPRECATIONS.md to track deprecations across Lagoon products.

I've opted to put it in the repository root, but we can discuss whether it should live in the docs?

Inspiration:
* https://platform.openai.com/docs/deprecations
* https://developers.google.com/maps/deprecations
* https://auth0.com/docs/troubleshoot/product-lifecycle/deprecations-and-migrations
* https://github.com/goreleaser/goreleaser/blob/main/www/docs/deprecations.md
* https://github.com/wix-incubator/wix-rest-docs/blob/master/guides/Deprecations.md

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -1,0 +1,20 @@
+
+
+# Lagoon Project Deprecation Announcements
+
+## Overview
+This file is a constantly updated central tracker for deprecations across the suite of Lagoon products. As Lagoon continues to evolve, we occasionally need to replace, rename or retire existing processes, tools or configuration. Where this may impact a user's processes or procedures, we intend to outline a timeframe to allow any necessary changes to be made.
+
+Deprecations will be tracked by the release they are announced in, and then updated when the actual deprecation occurs. All deprecations should provide a rough timeline (in months or releases). Releases will normally only be listed here if they include planned deprecations.
+
+## Deprecation History
+All deprecations are listed below, with the most recent announcements at the top.
+
+### Lagoon v2.17.0
+release link: https://github.com/uselagoon/lagoon/releases/tag/v2.17.0
+* (insert any planned deprecations here)
+
+### Lagoon v2.16.0
+release link: https://github.com/uselagoon/lagoon/releases/tag/v2.16.0
+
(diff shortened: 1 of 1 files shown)


DEPRECATIONS.md
```diff
@@ -0,0 +1,20 @@
+
```
shreddedbacon, line 1: Intentionally left blank lines?

DEPRECATIONS.md
```diff
@@ -0,0 +1,20 @@
...
+## Overview
+This file is a constantly updated central tracker for deprecations across the suite of Lagoon products. As Lagoon continues to evolve, we occasionally need to replace, rename or retire existing processes, tools or configuration. Where this may impact a user's processes or procedures, we intend to outline a timeframe to allow any necessary changes to be made.
+
+Deprecations will be tracked by the release they are announced in, and then updated when the actual deprecation occurs. All deprecations should provide a rough timeline (in months or releases). Releases will normally only be listed here if they include planned deprecations.
+
+## Deprecation History
+All deprecations are listed below, with the most recent announcements at the top.
+
+### Lagoon v2.17.0
```
shreddedbacon, line 13: I'm not sure how best to handle these linking to releases that may not exist yet?

shreddedbacon: Do we do something like this, and then when the release is about to be cut, the final cut release updates it to be correct?
```
### Lagoon next release
```

or this, and we move them into the release when its announced
```
### Upcoming deprecations
```

tobybellwood: I want to avoid having to made a file change as part of a release - I think at the point we're collecting the PRs for a release, we know what the release is going to be called?

This table just announces the deprecations - we can link back to the deprecation announcement once we do the deprecation

shreddedbacon: If you think its manageable otherwise, I'm happy to run with it

DEPRECATIONS.md
tobybellwood: nope - we can clip them in the first PR :wink:

PR #3608, opened 2023-11-30
Lagoon Build Errors pages

tobybellwood:
This PR adds a dedicated page for Lagoon Build errors, and two "short URLs" for use in Lagoon Builds that point to it

* https://docs.lagoon.sh/lagoon-build-errors
* https://docs.lagoon.sh/docker-compose-errors

We can add more errors that we trap in Lagoon Builds as needed.


shreddedbacon:
Maybe call it Errors and Warnings?

AlannaBurke:
I like this, I think Ben's suggestion is good. Also, maybe give some more detailed examples? You wrote "The error message should help you understand where." - would it be possible to put an example where that happens? 

I think we should add something at the top like "Here are some common errors and warnings you may encounter and how to resolve them. If you need help with your build, please reach out to your Lagoon admin or reach out on Discord for community help."

shreddedbacon:
>The error message should help you understand where.

For those ones in particular, the build itself will spew out the warning/error message and the appropriate remediation strategy and where it went wrong. This could be different depending on the way the user has done their docker-compose or `.lagoon.yml` file so it would be difficult to give concrete examples, but not impossible.

AlannaBurke (approved):

PR #3611, opened 2023-12-01
add local-dev-yarn command

tobybellwood:
This PR adds two makefile commands to facilitate easier yarn updating

`make local-dev-yarn` will mount the entire lagoon directory into a node-builder container, preinstall and build the yarn packages, and drop you into a shell to yarn away. When you `exit` the container, it clears itself up, leaving the updates to the package.json/yarn.lock in your repo.

`make local-dev-yarn-stop` is a convenience function that clears the named container away if a process exits badly (or at the beginning/end of a local-dev-yarn cycle.

This removes the need to set up node/yarn on your local machine, and avoids the need to remember how it all sets up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -347,6 +347,21 @@ down:
 kill:
 	docker ps --format "{{.Names}}" | grep lagoon | xargs -t -r -n1 docker rm -f -v
 
+.PHONY: local-dev-yarn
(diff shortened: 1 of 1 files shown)


shreddedbacon (approved):
This is a very handy feature for those without node installed locally

PR #3612, opened 2023-12-03
refactor: store project-group association in api instead of keycloak

shreddedbacon:
# General Checklist

# Database Migrations

This changes the way that project ids are stored against groups. It currently retains the function of adding the attributes in keycloak still for backwards compatibility with tools that use these attributes, but with a plan to remove this functionality once the other tools can be updated to use the project ids from the api-db.

The big changes are
* store the group id to project id association in a table
* store the group id to organization id association in a table
* cache the group name to group id reference  (purged when anything related to groups is performed)
* cache the group response payload (purged when anything related to groups is performed)
* cache the group members (purged when membership is modified)
* a migration script that will retrieve existing groups/projects from keycloak and add them to the database
* uses a custom mapper in keycloak to allow the opensearch token mapper to retrieve project ids from the api-db

## Considerations
* The biggest issue Lagoon has with using the attributes to store the project id assocation in keycloak is how we work out which groups a project has associated, as this requires us to retrieve all the groups from keycloak and iterate over them to check the attributes. This all groups query from keycloak is slow*number of groups, as more groups are added, slowness increases. 
	* Adding these associations in the api-db, we can now query the api-db and get the information we need so we only need to go to keycloak if we absolutely have to.
	* While one or two calls to the allgroups query is usually fine, the way that a large number of lagoon queries work results in this query needing to be performed potentially multiple times, or a large number of times when users interact with the API in busy periods. This results in keycloak CPU utilization spiking. https://github.com/uselagoon/lagoon/pull/3397 was a first attempt at reducing how often the allgroups query was called, and it purged the data when groups were modified, but as the number of groups increases, this query gets slower and slower resulting in slow API responses or timeouts during busy periods
* The `lagoon-projects` attribute on groups would need remain, at least for a period of time, as there are other systems that use these that would need to be refactored. While these other systems/tools use this attribute though, when a group is added/removed to a project, there is the potential due to the non-atomic nature of a lot of requests in Lagoon, that keycloak may not receive the attribute update, but the api-db does, or vice versa. This would result in incorrect access in either api or the other tools depending on the failure point. This not-atomic nature exists now, so this out of sync condition can and already does occur at times.
	* [opensearch token mapper uses these attributes](https://github.com/uselagoon/lagoon/blob/18977b4837c444adc0900c4631b45b4e017b0118/services/keycloak/javascript/mappers/groups-and-roles.js#L17) - This PR contains a custom java token mapper that will connect to the api-db to retrieve project IDs. This eliminates the need here for storing the project ids in keycloak.
	* [opensearch-sync uses the attributes still](https://github.com/uselagoon/lagoon-opensearch-sync/blob/3312841f27353c2c6b4a7aad0296d35d5193e2e3/internal/sync/indexpatterns.go#L126-L152) - should be easy to refactor to extract the ids from the api/api-db instead
	* [ssh-portal uses the attributes still](https://github.com/uselagoon/lagoon-ssh-portal/blob/3724a6210db9b35af7a1d9b6cb70379bc71c9801/internal/sshtoken/sessionhandler.go#L194) - should be easy to refactor to extract the ids from the api/api-db instead
* More data is cached from keycloak to speed up performance of group and member based queries. As organizations exposes a part of Lagoon that has typically been hidden away from users for so long, and thus relatively unused. Organizations is very heavy on group and group member queries, it made sense to cache a lot of this where possible.
	* an entire group response is cached, this query when performed against keycloak is not that bad, but if a large number of groups are requested against keycloak directly, the response time can be long.
	* a groups membership is also cached, this sort of query can also be quite heavy as it can request a considerable amount of information from keycloak if multiple groups and their members are requested, pulling this from the cache is ideal.
	* any changes to a groups project ids, or a member/role change within a group results in that groups entire cache data being purged, and will be re-populated when it is requested next.


services/api/src/models/group.ts
```diff
@@ -202,28 +219,30 @@ export const Group = (clients: {
...
+      group = R.pipe(
         R.reduce(flattenGroups, []),
-        R.filter(R.propEq('name', name)),
-        R.path(['0', 'id'])
-      )(keycloakGroups) as string;
+        R.filter(R.propEq('name', name))
+      )(keycloakGroups);
 
-      if (R.isNil(groupId)) {
+      if (R.isNil(group)) {
```
rocketeerbkw:
```suggestion
      if (R.isEmpty(group)) {
```
`group` will always be at least an empty array `[]` which means `R.isNil` is always false

services/api/src/models/user.ts
```diff
@@ -45,7 +46,7 @@ interface UserModel {
   loadUsersByOrganizationId: (organizationId: number) => Promise<User[]>;
   getAllOrganizationIdsForUser: (userInput: User) => Promise<number[]>;
   getAllGroupsForUser: (userId: string) => Promise<Group[]>;
-  getAllProjectsIdsForUser: (userInput: User, groups?: Group[]) => Promise<{}>;
+  getAllProjectsIdsForUser: (userId: string, groups?: Group[]) => Promise<{}>;
```
rocketeerbkw, line 49: There are still places where this parameter change hasn't been accounted for:

apolloServer.js:137 `groupRoleProjectIds = await User.User(modelClients).getAllProjectsIdsForUser(currentUser, keycloakUsersGroups);`
apolloServer.js:201 `groupRoleProjectIds = await User.User(modelClients).getAllProjectsIdsForUser(currentUser, keycloakUsersGroups);`
util/auth.ts:247 `groupRoleProjectIds = await UserModel.getAllProjectsIdsForUser(currentUser, keycloakUsersGroups);`

services/api/src/models/group.ts
```diff
@@ -530,6 +572,12 @@ export const Group = (clients: {
     }
 
     const group = await loadGroupById(response.id);
+    if (projectId) {
+      await groupHelpers(sqlClientPool).addProjectToGroup(projectId, group.id)
+    }
+    if (organizationId) {
+      await groupHelpers(sqlClientPool).addOrganizationToGroup(organizationId, group.id)
+    }
```
rocketeerbkw, line 598: This only adds the project/org association in the api-db for this function. Is it expected that `addProjectToGroup` is also called at some point? Otherwise it looks to me like this would cause a data sync issue between keycloak/api-db.

services/api/database/migrations/20240117000000_group_projects.js
```diff
@@ -0,0 +1,35 @@
...
+        return knex.schema
+    }
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function(knex) {
+    return knex.schema
```
rocketeerbkw: Why is there no rollback provision? Can we add a comment explaining why this migration doesn't follow conventions?

services/api/database/migrations/20240117000000_group_projects.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function(knex) {
+    group_projects = await knex.schema.hasTable('group_projects');
+    if (!group_projects) {
+        return knex.schema
+        // this table holds the main group to organization id association
+        .createTable('group_organization', function (table) {
```
rocketeerbkw: We talked about this briefly elsewhere. The long-term desire is probably to also move group data out of keycloak. In that scenario, we aren't necessarily limited to the data structure that keycloak has used and I'd prefer not to take the generic "group" prefix this early if there are bigger changes coming. It's a lot easier to move data from a `kc_group_*` table to a `group_*` table than to try and mutate `group_*` in-place.

Could be `kc_group_organization` or `keycloakgroup_organization`? Same for `group_projects`.

services/api/src/migrations/lagoon/migrations/20231130000000_group_projects.ts
```diff
@@ -0,0 +1,45 @@
...
+    redisClient
+  });
+
+  // load all groups from keycloak
+  const allGroups = await GroupModel.loadAllGroups();
+  // transform them into lagoon expected format
+  const keycloakGroups = await GroupModel.transformKeycloakGroups(allGroups);
+  for (const kg of keycloakGroups) {
+    // extract the project ids from the group
+    const groupProjects = await GroupModel.getProjectsFromGroupAndSubgroups(kg);
```
rocketeerbkw: I don't think `getProjectsFromGroupAndSubgroups` returns the desired result here? That is designed to calculate the projects a user has access to, taking into account group hierarchy. It will return projects that are not tied directly to the group being iterated (`kg`). But this migration is to copy 1-1 the data in keycloak, which should be tying projects only to the groups they are explicitly added to.

I think instead you would [flatten](https://github.com/uselagoon/lagoon/blob/main/services/api/src/models/group.ts#L198-L206) the `loadAllGroups` result, and [filter](https://github.com/uselagoon/lagoon/blob/main/services/api/src/models/group.ts#L147) out role groups. The remaining should be regular groups that you can iterate over and [getProjectIdsFromGroup](https://github.com/uselagoon/lagoon/blob/main/services/api/src/models/group.ts#L94-L101).

rocketeerbkw (changes requested):

services/api/src/models/group.ts
shreddedbacon: In `models/groups.ts/addProjectToGroup L744`  it is also called
https://github.com/uselagoon/lagoon/pull/3612/files/b5d533d7f7e39ea3f4098359fe500a2bfa255dcb#diff-4328fb3b581744d3027b0e4f5bf542287279bfca717b492571d5ab6fd336ee6bR761

Or am I misunderstanding this one

services/api/database/migrations/20240117000000_group_projects.js
shreddedbacon: We can do a rollback by dropping the tables. At some stage though when we stop storing the project attributes in keycloak, it will result in a data loss scenario for any groups added after that point.

I guess this is one of those issues with doing rollbacks generally though.

services/api/src/models/group.ts
rocketeerbkw: I think the only place this is actually called this way is during `addProject` and that also sets the project ids as part of the attributes` which means there's no data sync problem in actual usage, so no action required.

https://github.com/uselagoon/lagoon/blob/project-groups-db-cache/services/api/src/resources/project/resolvers.ts#L388

rocketeerbkw (approved):
Will work as a middle ground while we continue to explore options for all group data.

PR #3618, opened 2023-12-13
feat: allow registration of systemwide tasks

bomoko:
This PR introduces the ability to register system wide tasks.

These tasks will show up for particular environments just in case of the user having the appropriate permissions.

As part of the registration of a task, you will now need to set the 'systemWide' field to be equal to `true` in your API call, and specify no group, project, or environment.

# General Checklist

# Database Migrations


services/api/src/resources/task/task_definition_resolvers.ts
```diff
@@ -142,18 +142,31 @@ export const resolveTasksForEnvironment = async (
     Sql.selectAdvancedTaskDefinitionsForGroups(projectGroupsFiltered)
   );
 
+  const systemWideRows = await query(
+    sqlClientPool,
+    Sql.selectAdvancedTaskDefinitionsForSystem()
+  );
+
+
+
```
shreddedbacon: just small nits, fair amount of additional lines throughout that bloat the diff a bit

shreddedbacon (changes requested):
We should probably just flag [these existing tasks](https://github.com/uselagoon/lagoon/blob/feature/systemwide-tasks/services/api/src/typeDefs.js#L2384-L2396) as deprecated in this pull request with something like
```
@deprecated(reason: "This task will be removed in a future release. See https://github.com/uselagoon/lagoon/blob/main/DEPRECATIONS.md for alternatives if you use it")
```
Then update DEPRECATIONS.md in this PR with a note that version X announced the deprecation of the bundled tasks, and point to the documentation regarding implementation steps to replace them if you need them. And we should probably include graphql snippets of how to create them the same way that the existing tasks did.

Then once a few releases go by, we can fully remove them from the code base.

services/api/database/migrations/20231213_add_advanced_task_systemwide.js
```diff
@@ -0,0 +1,21 @@
+/**
```
shreddedbacon: This filename should probably be adjusted, the timestamp is shorter than others, and to be safe & ensure that the migrations run in the correct order, it would be good to update the filename to be in line with other migrations :safety_vest:

services/api/src/typeDefs.js
```diff
@@ -2378,19 +2381,19 @@ const typeDefs = gql`
     addWorkflow(input: AddWorkflowInput!): Workflow
     updateWorkflow(input: UpdateWorkflowInput): Workflow
     deleteWorkflow(input: DeleteWorkflowInput!): String
-    taskDrushArchiveDump(environment: Int!): Task
-    taskDrushSqlDump(environment: Int!): Task
-    taskDrushCacheClear(environment: Int!): Task
-    taskDrushCron(environment: Int!): Task
+    taskDrushArchiveDump(environment: Int!): Task @deprecated(reason: "This task will be removed in a future release. See https://github.com/uselagoon/lagoon/blob/main/DEPRECATIONS.md for alternatives if you use it")
```
shreddedbacon, line 2390: also need to update `DEPRECATIONS.md` now that 2.17 has released. Maybe something like this
```
### Lagoon v2.18.0
release link: https://github.com/uselagoon/lagoon/releases/tag/v2.18.0
* The standard drupal based tasks that Lagoon ships with (drush ....) have been flagged as deprecated and should not be used anymore. If you require these tasks, see [link to some notes on how to replace them]
```

We should probably write some documentation on how to replace the standard tasks with more Lagoon specific replacements, rather than direct drush replacements?

DEPRECATIONS.md
```diff
@@ -10,6 +10,10 @@ Deprecations will be tracked by the release they are announced in, and then upda
 ## Deprecation History
 All deprecations are listed below, with the most recent announcements at the top.
 
+### Lagoon v2.18.0
+release link: https://github.com/uselagoon/lagoon/releases/tag/v2.18.0
+* The standard drupal based tasks that Lagoon ships with (drush ....) have been flagged as deprecated and should not be used anymore. These will need to be replaced with [custom tasks](https://docs.lagoon.sh/using-lagoon-advanced/custom-tasks/). Example replacement tasks will be provided when these are completely deprecated.
```
shreddedbacon: Maybe this?
```
Example replacement tasks will be provided prior to their removal
```

shreddedbacon: This way the wording makes it sound like we aren't going to wait til the end to tell people how to replace them

shreddedbacon (approved):

shreddedbacon:
Once the tests pass :jenkins_dead:

PR #3624, opened 2023-12-14
refactor: upgrade keycloak to 21 and support initial realm import

shreddedbacon:
# General Checklist

# Database Migrations

This updates Keycloak to version 21, which is based on quarkus. This gets us into a better position to move to Keycloak version 23.

It also changes the startup of Keycloak to use a realm import script that gets Keycloak to a point of the latest changes. This is only called if the `lagoon` realm does not currently exist. This way a fresh install of Lagoon will be up and running much quicker than having to perform all the migrations from the dawn of time.

This significantly speeds up Keycloak startup time. 

This currently uses a realm export of the changes done in what would be Lagoon 2.17.0, but may need to be updated periodically if there are any changes. As such, whatever version this would be merged into of Lagoon would need a release note to say that users cannot upgrade to this version, and they must upgrade to the version prior before they can update to the version this is released in. As the whole point of this is to eliminate the dawn of time migrations that take place, as these run even on existing installations just in a no-op form, but still takes time before it gets to the more current migrations.

Future migrations can be added to this though that perform migrations the same way older migrations took place, but it would mean that it could get to the same position again where the migrations take too long.

I propose that every migration also gets added/updated into to the realm import. This way the initial realm import on brand new installs will always be up to date, and the migrations can periodically be cleaned up using the same method of `you need to update to version X before you update to version Y`. We should also track these upgrade paths better.


rocketeerbkw:
Discussed in meetings, good to merge and let us test in our clusters to ensure we got the right env var changes for quarkus :+1:

rocketeerbkw (approved):

PR #3627, opened 2023-12-17
Adds service to deleteFactsFromSource

bomoko:
This PR introduces the same functionality as an earlier PR https://github.com/uselagoon/lagoon/pull/2611

The rationale for its existence mirrors that of the earlier PR, but _mutatis mutandis_ wrt facts.

Essentially, it allows us to better target facts for clearing, since multiple facts could potentially exist with the same name, same source, but on different services. 

Think, for instance, a `php-version` collected by some process that runs on both the `php-nginx` service and a `cli` service, we want to be able to clear the facts from one of the services, without having it be cleared from both.

# General Checklist

### DIFF
--- a/services/api/src/resources/fact/resolvers.ts
+++ b/services/api/src/resources/fact/resolvers.ts
@@ -382,18 +382,22 @@ export const deleteFact: ResolverFn = async (
 
 export const deleteFactsFromSource: ResolverFn = async (
   root,
-  { input: { environment: environmentId, source } },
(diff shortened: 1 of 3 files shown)


CGoodwin90 (approved):

shreddedbacon:
Good to merge from my perspective, this just extends an existing mutations input without any breaking changes, and the logic contained within appears sound for the description provided.

PR #3628, opened 2023-12-17
feat: support idling and unidling from api with idled status

shreddedbacon:
# General Checklist

# Database Migrations

Just adding support to the API to be able to control idling from the API and view the state of an environment.

Also allows the ability to stop, start, or restart, services within an environment.

# Requires

The following pullrequests are requirements to support this. The changes in `remote-controller` can be released prior to this and the functionality will be basically dormant until the API supports it.

* https://github.com/uselagoon/remote-controller/pull/245
* https://github.com/uselagoon/machinery/pull/77


services/keycloak/lagoon-realm-base-import.json
```diff
@@ -2884,6 +2917,72 @@
...
+            "logic": "POSITIVE",
+            "decisionStrategy": "UNANIMOUS",
+            "config": {
+              "resources": "[\"environment_state\"]",
+              "scopes": "[\"restart:development\"]",
+              "applyPolicies": "[\"[Lagoon] User has access to project\",\"[Lagoon] Users role for project is Developer\"]"
+            }
+          },
+          {
+            "name": "Restart production environment services",
```
CGoodwin90: Duplicated permission `Restart production environment services`

services/api/src/resources/environment/resolvers.ts
```diff
@@ -1128,3 +1144,254 @@ export const getServiceContainersByServiceId: ResolverFn = async (
...
+    await createMiscTask({ key: 'environment:service', data });
+    return 'success';
+  } catch (error) {
+    sendToLagoonLogs(
+      'error',
+      '',
+      '',
+      'api:stopOrStartEnvironmentService',
+      { environment: environmentData.id },
+      `Environment idle attempt possibly failed, reason: ${error}`
```
CGoodwin90: Copy pasta from `idleOrUnidleEnvironment`

services/api/src/resources/environment/resolvers.ts
```diff
@@ -1128,3 +1144,254 @@ export const getServiceContainersByServiceId: ResolverFn = async (
...
+    );
+  }
+  if (environmentData.idleState != 'active') {
+    throw new Error(
+      `Can't perform action because the environment is idled`
+    );
+  }
+  await hasPermission(
+    'environment_state',
+    `restart:${environmentData.environmentType}`,
```
CGoodwin90, line 1306: Looks like all actions are checked against the `restart` permission, are we just going to use that as a catch-all rather than separate permissions for `start` & `stop`? Not an issue, just getting an understanding.

services/actions-handler/handler/controller_idling.go
```diff
@@ -0,0 +1,169 @@
...
+				log.Printf("%sERROR: unable to get environment - %v", prefix, err)
+			}
+			return err
+		}
+		environmentID = environment.ID
+	} else {
+		// pull the id from the message
+		environmentID = *message.Meta.EnvironmentID
+	}
+	decodeData, _ := base64.StdEncoding.DecodeString(message.Meta.AdvancedData)
```
CGoodwin90: With no error handling, if this or the `unmarshall` fails it could potentially pass through an empty `byteSlice`/`struct`.

CGoodwin90 (changes requested):

services/keycloak/lagoon-realm-base-import.json
shreddedbacon: :see_no_evil:

services/api/src/resources/environment/resolvers.ts
shreddedbacon: I'm open to changing the permissions, but I didn't see a case where someone would be able to stop a service in isolation without being able to start it again.

CGoodwin90 (approved):

PR #3630, opened 2023-12-21
Clean up insights data on environment deletion

bomoko:
Right now, when deleting an environment, there is nothing to clean up facts and problems that have been registered against that environment.

The really big problem here (beyond the fact that we're not cleaning up data) is with multiple PR/ephemeral environments, if they're running insights, there is potentially a _lot_ of insights information that is not even of any historical value.

This PR adds two things
1. Calls to queries that will clean up the appropriate facts and problems entries.
2. Adds a migration to remove any existing data that should be removed from the DB.

# General Checklist

# Database Migrations

### DIFF
--- a/services/api/database/migrations/20231222000000_delete_orphaned_insights.js
+++ b/services/api/database/migrations/20231222000000_delete_orphaned_insights.js
@@ -3,12 +3,20 @@
  * @returns { Promise<void> }
  */
 exports.up = function(knex) {
-    return knex.schema.raw(`
-    DELETE ef
-    FROM environment_fact ef
-    LEFT JOIN environment e ON ef.environment = e.id
-    WHERE e.id IS NULL OR e.deleted != '0000-00-00 00:00:00'
-  `);
+    return Promise.all([
+      knex.schema.raw(`
+       DELETE ef
+       FROM environment_fact ef
+       LEFT JOIN environment e ON ef.environment = e.id
+       WHERE e.id IS NULL OR e.deleted != '0000-00-00 00:00:00'
+      `),
+      knex.schema.raw(`
+       DELETE ep
+       FROM environment_problem ep
+       LEFT JOIN environment e ON ep.environment = e.id
+       WHERE e.id IS NULL OR e.deleted != '0000-00-00 00:00:00'
+      `),
+]);
   };
 
   /**
--- a/services/api/database/migrations/20231222000000_delete_orphaned_insights.js
+++ b/services/api/database/migrations/20231222000000_delete_orphaned_insights.js
@@ -8,15 +8,15 @@ exports.up = function(knex) {
        DELETE ef
        FROM environment_fact ef
        LEFT JOIN environment e ON ef.environment = e.id
-       WHERE e.id IS NULL OR e.deleted != '0000-00-00 00:00:00'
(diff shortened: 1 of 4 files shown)


services/api/database/migrations/20231222000000_delete_orphaned_insights.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function(knex) {
+    return knex.schema.raw(`
+    DELETE ef
```
shreddedbacon: You may also want to run a separate one that prunes anything from environments that may not be marked as "deleted", but where the associated project no longer exists

There was a time where you could delete a project, but the environments would remain "undeleted" in the API. We have a check in place on project deletions that should prevent this, but there could still be lingering environments without an attached project that show as active.

Example as a select that would could be turned into a delete
```
SELECT e.name, e.id, e.project, e.deleted, ef.id, ef.name
    FROM environment_fact ef
    LEFT JOIN environment e ON ef.environment = e.id
    WHERE e.id IN (
        SELECT e2.id
        FROM (SELECT * FROM environment) AS e2
        LEFT JOIN project p on p.id = e2.project
        WHERE p.name IS NULL
    );
```

---
Might be worth a separate PR with a migration that could detect any environments that don't have an associated project, and mark them as "deleted", this is a query that can show them if they exist (our test cluster has a couple)
```
SELECT e.name, e.id, e.project, e.deleted
    FROM environment e
    WHERE e.deleted = '0000-00-00 00:00:00' AND e.id IN (
        SELECT e2.id
        FROM (SELECT * FROM environment) AS e2
        LEFT JOIN project p on p.id = e2.project
        WHERE p.name IS NULL
    );
```

bomoko: Ah, of course, slipped my mind - thanks for reminding me.

I believe we can just change up the current migrations to do it? Wanna take a quick look @shreddedbacon 

Agreed about the separate PR.

shreddedbacon: Well the two queries I proposed are for different issues. This PR should probably include the first migration cleanup at a minimum as it will remove data for environments with no project associated, where the environment may not be marked as deleted, but effectively is because the project is deleted. So the data lingering should go too.

The second migration would be something that would mark environments with no project associated as deleted, the second one could be in a separate PR though.

bomoko: Hey @shreddedbacon - sorry, I think my response was too terse :D

1 - for the first query, that's what I meant when I said "I think we can do it in the existing migration" - I actually meant to just update the existing cleanup queries - you can see that in https://github.com/uselagoon/lagoon/pull/3630/commits/915f3818290ca790fa541ffd4e6576ca0b227acc

2 -100, I agree(d) about a separate PR

shreddedbacon (approved):
Yep, this makes sense

services/api/database/migrations/20231222000000_delete_orphaned_insights.js
shreddedbacon: Ah yep, sorry misread :shipit:

Issue #1980, opened 2020-06-22
Lagoon dashboard audit trail to include user


proeinfo:
It would be nice to have an audit trail with username for the various functions on the Lagoon dashboard.

This would optimally be for any action (deployments, tasks, backup restores, etc.) that users can perform.

As a developer I would like to know who did what and when, as it makes it easier to understand and investigate changes and outages in a given environment.

It also makes sense for security reasons.

tobybellwood:
also partly covered by #911

PR #3631, opened 2023-12-27
Improvement: add source user and type to tasks and deployments

### LINKED ISSUE #3144: Add author name to pushes
At least for github pushes I believe https://github.com/uselagoon/lagoon/compare/main...chx:patch-1 is all you need.

I could expand on this of course if there's interest.

shreddedbacon:
# General Checklist

# Database Migrations

Track the user or system that triggered a task or deployment.
It adds `sourceType` and `sourceUser` to Task and Deployments which can only be set when the tasks are created initially. These fields cannot be modified afterwards (beyond direct database interactions)

The two trigger types are `API` and `WEBHOOK`. Deployments currently support both source types.
Tasks are not triggered by webhooks, so only the `API` type is available to it.

The webhook source type will also try and extract information about the user that initiated the webhook call if possible, otherwise the trigger user will just be `webhook`.

API source types will try and retrieve the user email address from the token, if the user triggering the task is a system token, it will extract the `iss` field from the system token, this could result in strange names if a token is really old.

This does not backport the data to existing tasks/deployments as this information is unknown, so the data will be `null` for old tasks and deployments. The thought of adding an `UNKNOWN` type with it being the default, so historical fields would be `UNKNOWN` type. The source user could also default to `unknown` if not provided. Rather than being a `null` field.

# Closing issues

closes #3144 
partially #1980 (better audit capability in lagoon needed generally)

### DIFF
--- a/services/api/database/migrations/20240115000000_taskdeployment_source.js
+++ b/services/api/database/migrations/20240115000000_taskdeployment_source.js
@@ -1,0 +1,31 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function(knex) {
+    return knex.schema
+    .alterTable('task', function (table) {
+        table.enu('source_type', ['api']);
+        table.string('source_user', 300);
+    })
+    .alterTable('deployment', function (table) {
+        table.enu('source_type', ['api', 'webhook']);
+        table.string('source_user', 300);
+    })
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function(knex) {
+    return knex.schema
+    .alterTable('task', (table) => {
+        table.dropColumn('source_type');
+        table.dropColumn('source_user');
+    })
+    .alterTable('deployment', (table) => {
+        table.dropColumn('source_type');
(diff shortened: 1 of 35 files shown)


services/api/database/migrations/20240115000000_taskdeployment_source.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function(knex) {
+    return knex.schema
+    .alterTable('task', function (table) {
+        table.enu('source_type', ['api']);
```
tobybellwood, line 8: acknowledging that there may be future different ways to trigger a task, to be handled in future ENUM migrations

shreddedbacon:
I've added some basic tests to the test suite to check that the fields are populated when deployments are triggered via webhook and api calls both git push and pullrequest, deployenvironmentlatest, and deployenvironment branch.

shreddedbacon:
the `promote` test still seems to use a `test-suite` legacy token, when other tests uses a `ci` user token, this can be resolved in a separate PR though

tobybellwood (approved):
This looks good to me, and +100 for the tests.

Happy to follow up the errant test token in a follow up.

PR #3632, opened 2023-12-27
feat: introduce /.well-known endpoint for service discovery information

shreddedbacon:
# General Checklist

# Database Migrations

Just a simple way to expose some discovery data via a `/.well-known/appspecific/sh.lagoon.discovery.json` endpoint.

This could expose other information too, and for tools like `lagoon-cli` and `lagoon-sync`, users only need to know the API hostname (and maybe port if not standard http/s) when configuring their tools, rather than needing to know all the other endpoint information.

This contains some basic information like so, as long as the variables are provided to populate it correctly. Some are already provided by the chart, others will need to be created in the chart
```
{
    "lagoon_version": "v2.17.0",
    "authorization_endpoint": "https://keycloak.example.com",
    "ssh_token_exchange": {
        "token_endpoint_host": "token.example.com",
        "token_endpoint_port": 22
    },
    "webhook_endpoint": "https://webhook.example.com",
    "ui_url": "https://ui.example.com"
}
```

This requires updates to the charts to inject `SSH_TOKEN_ENDPOINT` and `SSH_TOKEN_ENDPOINT_PORT` variables that require a user to define, the `port` may be able to be consumed via the chart though, so only the hostname would need to be provided.

Conditions for the endpoint DNS record would be if this is a fresh install of the token or ssh service, the cluster loadbalancer/service IP or name for the token endpoint may not be known immediately, so could be set to the DNS record that will eventually be populated once the loadbalancer/service IP is know after deployment.

### DIFF
--- a/services/api/src/app.ts
+++ b/services/api/src/app.ts
@@ -22,7 +22,15 @@ app.use(json());
 app.use(
   morgan('combined', {
     skip: (req, res) => {
-      return req.originalUrl.startsWith('/status');
+      if (req.originalUrl.startsWith('/status')) {
+        return req.originalUrl.startsWith('/status')
+      }
+      if (req.originalUrl.startsWith('/favicon.ico')) {
+        return req.originalUrl.startsWith('/favicon.ico')
+      }
+      if (req.originalUrl.startsWith('/.well-known')) {
+        return req.originalUrl.startsWith('/.well-known')
+      }
     },
     stream: {
       write: message => logger.info(message.trim())
--- a/services/api/src/authMiddleware.ts
+++ b/services/api/src/authMiddleware.ts
@@ -44,6 +44,10 @@ const authenticateJWT = async (
   if (req.url === '/status') {
     return next();
   }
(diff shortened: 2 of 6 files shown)


smlx:
[IANA](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml) has an `appspecific` URI suffix for this purpose that you might be able to use. I guess it would be something like `/.well-known/appspecific/sh.lagoon.discovery.json`?

shreddedbacon:
> [IANA](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml) has an `appspecific` URI suffix for this purpose that you might be able to use. I guess it would be something like `/.well-known/appspecific/sh.lagoon.discovery.json`?

Yeah, that can work too

Just would have been nicer to use `/.well-known/lagoon` from a vanity perspective :sparkles:

rocketeerbkw (approved):
Also discussed in meeting, agreed this URL scheme shouldn't cause future headaches.

PR #3633, opened 2023-12-28
fix: use Float for restore size

### LINKED ISSUE #3629: API uses `Int` for backup restore sizes which can be larger than `Int` size
# Describe the bug

The GraphQL type for the `restoreSize` field is set as `Int` which [only allows signed 32-bit integers](https://graphql.org/learn/schema/#scalar-types). Since the size is recorded in bytes, this can be larger than the `Int` type. In this case the api will error with: `Int cannot represent non 32-bit signed integer value`

## To Reproduce

1. Restore a backup larger than 2.2GB
2. Attempt to retrieve restore info from the api (or view the backups page in the UI)

## Expected behavior

The api stores and retrieves restore size without errors.

## Additional context

Introduced in https://github.com/uselagoon/lagoon/pull/3599.

shreddedbacon:
# General Checklist

# Database Migrations

Changes the return type from `Int` to `Float`, based on discussion in #3629 

Other services in the API that store information related to storage size use Float, and these values would very likely see sizes into the GB if not TB of sizes that a restore would also see.
https://github.com/uselagoon/lagoon/blob/v2.17.0/services/api/src/typeDefs.js#L941
https://github.com/uselagoon/lagoon/blob/v2.17.0/services/api/src/typeDefs.js#L947

# Closing issues

closes #3629

### DIFF
--- a/services/api/src/typeDefs.js
+++ b/services/api/src/typeDefs.js
@@ -975,7 +975,7 @@ const typeDefs = gql`
     """
     The size of the restored file in bytes
     """
-    restoreSize: Int
+    restoreSize: Float
     created: String
   }
 


rocketeerbkw (approved):
Tested with a large value `1000000000000000000000000000000000000000000000000000000000000000` and the api returns `"restoreSize": 1e+63`. This is a number format that javascript can natively use, so it won't error out our current UI implementation. It also corresponds to a file size of 1e+39 YB (yottabytes), which is so large that I couldn't find an example online to compare it to, so it should be fine for file sizes :rofl:

PR #3637, opened 2024-01-11
fix: return error when no matching event to prevent empty email notifications

shreddedbacon:
# General Checklist

# Database Migrations

Just a fix for the email notification handler to prevent empty email notifications from being sent when no matching event type is found.


bomoko (approved):
Makes sense - in these fall through cases the client code will now not send an email - which is the desired behavior.

LGTM

PR #3644, opened 2024-01-29
Rejig test order

tobybellwood:
This PR just balances out the test suite a little and moves some of the tests around. It should complete a little faster now, and with hopefully less contention!


shreddedbacon (approved):
This shaves a couple of minutes off the overall run time for sure. Definitely an improvement as an initial run through of reducing the overall test speed for minimal effort.

PR #3645, opened 2024-01-31
Change: check expiration of legacy tokens, reject if necessary

shreddedbacon:
# General Checklist

# Database Migrations

As we look to phase out the usage of legacy tokens, we need to be able to reject long lived or non-expiring legacy tokens initially.
This introduces two new environment variables that can be set in the API.

* `LEGACY_EXPIRY_MAX` (default 3600s) - This is the maximum expiry that can be allowed to interact with the API
* `LEGACY_EXPIRY_REJECT` (default false) - This is what controls if a legacy token is rejected based on the checks

If a legacy token has the `exp` field, the remaining duration of the token is calculated against the `iss` timestamp. If this duration is greater than the `LEGACY_EXPIRY_MAX`, then depending on the `LEGACY_EXPIRY_REJECT` setting, it will log, or log and reject the request.

If there is no `exp` field on the token, then depending on the `LEGACY_EXPIRY_REJECT` setting, it will log, or log and reject the request.

### DIFF
--- a/services/api/src/util/auth.ts
+++ b/services/api/src/util/auth.ts
@@ -13,6 +13,7 @@ import { saveRedisKeycloakCache } from '../clients/redisClient';
 
 interface ILegacyToken {
   iat: string;
+  exp: string;
   iss: string;
   sub: string;
   aud: string;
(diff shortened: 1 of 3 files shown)


services/api/src/util/auth.ts
```diff
@@ -105,7 +106,26 @@ export const getCredentialsForLegacyToken = async token => {
...
+  const rejectLegacyExpiry = getConfigFromEnv('LEGACY_EXPIRY_REJECT', 'false') // don't reject intially, just log
+  if (exp) {
+    if ((parseInt(exp)-parseInt(iat)) > parseInt(maxExpiry)) {
+      const msg = `Legacy token (sub:${sub}; iss:${iss}) expiry ${(parseInt(exp)-parseInt(iat))} is greater than ${parseInt(maxExpiry)}`
+      logger.warn(msg);
+      if (rejectLegacyExpiry == "true") {
+        throw new Error(msg);
+      }
+    }
+  } else {
```
rocketeerbkw, line 122: This is checking that the token has a maximum expiration window (e.g., only valid for 60 seconds), but it doesn't check if the token has actually expired. This would still allow a token that has expired (`exp` is a date in the past) to be used as long as the max expiration time is within limits.

I think there would need to be three checks (along with the `rejectLegacyExpirary` flag): If the token is expired, deny. If the token is not expired and has a too long expiration window, deny. If the token does not have an expiration date, deny.

rocketeerbkw (changes requested):

services/api/src/util/auth.ts
shreddedbacon: Isn't that what happens here https://github.com/uselagoon/lagoon/pull/3645/files/ec02b5fe2c6835ffcd6cc1ecfa67c62f0b0d3d55#diff-2bc3f88bb67720bab520a61d6c56739b9d1a5239acbe0c290855e9a3d287ecdaR103 ?

rocketeerbkw: Nevermind then :)

rocketeerbkw (approved):
Discussed in meeting, looks good :+1:

tobybellwood (approved):
Approving additional logic to set internal tokens to 60sec

PR #3647, opened 2024-02-05
Deprecate activeSystems* in API

rocketeerbkw:
# General Checklist

# Database Migrations

Lagoon has not supported multiple "activeSystems*" for quite some time. This PR removes all logic tied to their data and marks the fields deprecated in the API.

# Closing issues

n/a


shreddedbacon:
Probably just add a deprecation notice in the `DEPRECATIONS.md`, use `2.18.0` for now similarly to #3618 and we can work out merge conflicts when it comes time to merge

rocketeerbkw:
The tests are failing at the same spot, so not a transient testing issue.

node-packages/commons/src/tasks.ts
```diff
@@ -776,7 +760,7 @@ export const createDeployTask = async function(deployData: any) {
     pullrequestTitle
   } = deployData;
 
-  const project = await getActiveSystemForProject(projectName, 'Deploy');
+  const project = await getOpenShiftInfoForProject(projectName);
```
shreddedbacon: `getOpenShiftInfoForProject` return payload returns `{project: {project: { id, name, etc... }}}`

Which means you need to do [something like this](https://github.com/uselagoon/lagoon/blob/deprecate-activesystems/node-packages/commons/src/tasks.ts#L364-L365) to be able to then access the actual project data properly.

Maybe its worth fixing `getOpenShiftInfoForProject` so that it only returns the project data, rather than needing to extract the bit thats actually needed afterwards. I don't care either way, but if you change the response, you'll need to update other places that consume it then

node-packages/commons/src/tasks.ts
```diff
@@ -907,27 +877,14 @@ export const createPromoteTask = async function(promoteData: any) {
...
-          promoteData
-        }
-        return deployTargetPromote(lagoonData)
-    default:
-      throw new UnknownActiveSystem(
-        `Unknown active system '${project.activeSystemsPromote}' for task 'deploy' in for project ${projectName}`
-      );
+  // use deployTargetPromote function to handle
+  let lagoonData = {
+    projectId: project.id,
```
shreddedbacon, line 966: This needs to be `projectId: project.project.id` because `getOpenShiftInfoForProject`. See other comment on possible resolution

shreddedbacon (changes requested):

node-packages/commons/src/tasks.ts
shreddedbacon: It is used in quite a few places though

rocketeerbkw:
:pray: to the test Gods

shreddedbacon (approved):
Looks good. Especially like the cleanup done in `api.ts` to remove all the unused functions :rocket: 

Not fully wrapped my head around typescript `Pick<>`, but reading about how it works makes sense.

PR #3652, opened 2024-02-12
fix: add DropUnique to rollback ES migration

tobybellwood:
In #3641 additional tables, fields and indexes were added to show environment<>service connections. In the rollback migration, an index was missed for removal. This PR resolves that. It doesn't need a seperate migration, as it hasn't been released yet, and would only be encountered on a rollback.

This PR also adds a github action test designed to try and catch any rollback/migration errors at the point of initial commit. It shouldn't fire without a migration present.

In the development of this PR, it was discovered that migrations could fail if there were multiple services in the same environment with the same "name". As this name is not descriptive (nor unique), the migration now adds a functionality to remove any duplicate service names. With the advent of the build deploy and remote controller changes, it will become impossible for an environment to have multiple identically named services (as it is a docker compose constraint).

### DIFF
--- a/services/api/database/migrations/20240114000000_environment_services.js
+++ b/services/api/database/migrations/20240114000000_environment_services.js
@@ -4,6 +4,7 @@
  */
 exports.up = async function(knex) {
     return knex.schema
+    .raw(`DELETE es1 FROM environment_service es1 INNER JOIN environment_service es2  WHERE  es1.id < es2.id AND es1.name = es2.name;`)
     .alterTable('environment_service', function (table) {
         table.string('type', 300);
         table.timestamp('updated').notNullable().defaultTo(knex.fn.now());
--- a/services/api/database/migrations/20240114000000_environment_services.js
+++ b/services/api/database/migrations/20240114000000_environment_services.js
@@ -27,6 +27,7 @@ exports.down = async function(knex) {
         table.dropColumn('type');
         table.dropColumn('updated');
         table.dropColumn('created');
+        table.dropUnique(['name', 'environment'], 'service_environment');
     })
     .dropTable('environment_service_container')
 };
--- a/.github/workflows/test-db-migrations.yml
+++ b/.github/workflows/test-db-migrations.yml
@@ -1,0 +1,62 @@
+name: Test database migration and rollback
+
+on:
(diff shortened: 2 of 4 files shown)


services/api/database/migrations/20240114000000_environment_services.js
```diff
@@ -8,8 +8,11 @@ exports.up = async function(knex) {
         table.string('type', 300);
         table.timestamp('updated').notNullable().defaultTo(knex.fn.now());
         table.timestamp('created').notNullable().defaultTo(knex.fn.now());
-        table.unique(['name', 'environment'], {indexName: 'service_environment'});
     })
+    // create a unique index but drop any duplicates when the unique is added
+    // the deprecated setEnvironmentServices resolver will remove duplicates before adding them now too
+    .raw(`DELETE es1 FROM environment_service es1 INNER JOIN environment_service es2  WHERE  es1.id < es2.id AND es1.name = es2.name;`)
```
tobybellwood: I'd suggest moving this above the first up stage, and then bundling the index into the knex alter table stage.

services/api/database/migrations/20240114000000_environment_services.js
```diff
@@ -8,8 +8,11 @@ exports.up = async function(knex) {
...
+    .raw(`ALTER TABLE environment_service ADD UNIQUE INDEX service_environment (name, environment);`)
```
tobybellwood: I'd convert this back to the native knex statement and move it into the knex alter table, just to avoid reliance on raw sql where possible?

services/api/database/migrations/20240114000000_environment_services.js
```diff
@@ -28,5 +31,6 @@ exports.down = async function(knex) {
         table.dropColumn('updated');
         table.dropColumn('created');
     })
+    .raw(`ALTER TABLE environment_service DROP INDEX service_environment;`)
```
tobybellwood: I'd convert this back to the native knex statement, and put in inside the knex alter table, just to avoid reliance on raw sql where possible?

tobybellwood:
Merging as reviewed and functional tested by @tobybellwood and @shreddedbacon

PR #3655, opened 2024-02-21
fix: add correct roles and scopes to keycloak account client realm import

shreddedbacon:
# General Checklist

# Database Migrations

Just fixes the realm import for initial installs to import the roles and scopes for the account client to allow users to access their profile page in the keycloak UI.

Prior to this, if a user went to `Your account` from the Lagoon UI and clicked on anything to edit their account it would give a permission denied. This fixes that by importing a known export of the roles and scopes.


tobybellwood (approved):
Gone through this and all makes sense. Nice pickup on the random s/maintainer/guest mismatch.

PR #3657, opened 2024-02-21
fix: use email address instead of username when looking up users

### LINKED ISSUE #3656: Changing email address has unintended side effects
If I log in to the Lagoon UI then go to `Your account` which takes me to the keycloak account settings page and change my email address, my account name in Lagoon UI remains the same as my old email address, and I am unable to change my username to match my email address.

This has another issue in that most user interactions require the email address. Unfortunately, Lagoon seems to pass the email address in to a username look up when interacting with Keycloak, so while I can change my email address, I still need to do any interactions using my old email address because that is what my username is.

For example, if I change the `maintainer@example.com` email address to `not-maintainer@example.com`, I still need to use `maintainer@example.com` to add the user to a group or remove this user from a group.

shreddedbacon:
# General Checklist

# Database Migrations

Just fixes the underlying resolvers and models to use the email address correctly.

# Closing issues

closes #3656

### DIFF
--- a/services/api/src/models/user.ts
+++ b/services/api/src/models/user.ts
@@ -40,7 +40,7 @@ interface UserEdit {
 interface UserModel {
   loadAllUsers: () => Promise<User[]>;
   loadUserById: (id: string) => Promise<User>;
-  loadUserByUsername: (username: string) => Promise<User>;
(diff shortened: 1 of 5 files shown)


rocketeerbkw (approved):
There shouldn't be any other issues with this.

PR #3658, opened 2024-02-23
Change: deprecate environment storage bytesUsed for kibUsed

### LINKED ISSUE #2245: Lagoon API lists `bytesUsed` when it is actually `kiloBytesUsed`

shreddedbacon:
# General Checklist

# Database Migrations

This introduces a deprecation on `bytesUsed` when interacting with environment storage queries and mutations.

When retrieving environment storage sizes, `bytesUsed` is provided as the name when the value of this data is actually kibibytes. This PR is the first step in the removal of `bytesUsed` and renaming to `kibUsed`. A follow up PR will introduce the actual breaking change to remove it in a future release.

Additionally, `bytesUsed` is an `Int` type, which has a 32bit integer limitation. The new `kibUsed` input is a `Float` type to allow greater input.

For now, `bytesUsed` is still the required input for `addOrUpdateEnvironmentStorage` with `kibUsed` as an optional argument. `kibUsed` will be preferred though if it is provided. Unfortunately, as `bytesUsed` is still an `Int` large numbers will still cause an error if they are provided, until `bytesUsed` is completely removed from the API in a future version

# Closing issues

closes #2245


rocketeerbkw (approved):
There could be nicer ways to deprecate this schema but it's primary use is for integrations with other systems (not end-user facing) and it requires platform-owner privileges, so I don't think it's worth any further effort.

PR #3659, opened 2024-02-23
chore: remove unneeded reference to legacy registry setting

### LINKED ISSUE #2238: improve usage of `REGISTRY` env variable
during working on the lagoon-core charts (https://github.com/uselagoon/lagoon-charts/issues/97) we saw a couple of weird things with the `REGISTRY` variable.

It's usage is for these cases:
- In OpenShift/MiniShift cases it's actually not used at all as we use the internal OpenShift Registry all the time that is always at a known location `docker-registry.default.svc:5000`
- in k8s production deployments we overwrite the registry to use with the injected internal registry (harbor) see https://github.com/amazeeio/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy.sh#L62, this only happens only if there is a harbor connected to lagoon, there are cases where this is not the case:
- in k8s local development and CI we use the env variable to tell Lagoon to use an unauthenticated registry that runs in docker-compose (we do this because running a full harbor during local development is quite a heavy task)

While this itself is an okay system, the way we inject the env variable is super confusing:
- we inject the registry into the build pods and as the build pods are created via lagoon-kbd which itself is created by the `controller-handler`
- the `controller-handler` does not know much about the deployment requirements itself and instead the payloads for the deployments (which include the registry) are managed by the commons/tasks.ts here: https://github.com/amazeeio/lagoon/blob/main/node-packages/commons/src/tasks.ts#L92 which is used when `createDeployTask()` is called
- and `createDeployTask()` is used by two services: https://github.com/amazeeio/lagoon/search?q=createDeployTask:
  - webhooks2tasks
  - api

so in the end if we want to change the registry we need to change it in two services which then pass the env variable to other services just that it will be overwritten again at the very end (in the case of lagoon production).
Not very efficient or a nice DX.

therefore I would suggest:
- remove `REGISTRY` from lagoon-core all together
- have lagoon-kbd inject a registry if we have deployed lagoon-kbd with a specific registry (https://github.com/amazeeio/lagoon-kbd/blob/28950495d575bca45d8095f2dac3e2e6613bb855/controllers/lagoonbuild_controller.go#L298-L301), in the case of OpenShift we tell lagoon-kbd to use the openshift registry. For Local development we tell it to use the non-auth registry running in docker-compose
- for CI we start to use harbor as the registry so we actually test the interaction between lagoon and harbor during CI

shreddedbacon:
# General Checklist

# Database Migrations

Just removes an unneeded reference to a registry. This variable is not consumed by anything in core beyond passing it through to the remote. The usage of this is from a historical perspective, it was used when Lagoon didn't have support for harbor in core, or harbor in remote.

It isn't related to the legacy harbor implementation in core either as these provide variables to the project in the form of `INTERNAL_REGISTRY_X` which get added to the build payload when it sent to a remote. The legacy harbor implementation in core is planning to be removed (#2907 , #3259, and others) and was [announced as deprecated in v2.17.0](https://github.com/uselagoon/lagoon/blob/v2.17.0/DEPRECATIONS.md?plain=1#L16-L17)

# Closing issues

closes #2238


tobybellwood (approved):
straightforward enough

PR #3664, opened 2024-03-06
Documentation: Updates promote documentation to use lagoon-cli

bomoko:
The documentation about the "promote" workflow is slightly out of date.

This replaces the `curl` call in the documentation with the lagoon cli.


tobybellwood (approved):
oof - that was out of date, thanks!

PR #3669, opened 2024-03-07
Feature: Implements `getNotificationByNameAndType`

CGoodwin90:
Implements `getNotificationByNameAndType` allowing for a `notification` to be retreived by its `name` and `type`. Retreives both unassigned & assigned notifications.


services/api/src/typeDefs.js
```diff
@@ -579,6 +579,7 @@ const typeDefs = gql`
     contentType: String
     notificationSeverityThreshold: ProblemSeverityRating
     organization: Int
+    pid: Int
```
shreddedbacon: Are these `pid` used? It looks like they aren't, so can this be removed?

CGoodwin90: Added the `pid` initially to be able to show assigned vs unassigned notifications, though this setup doesn't account for a notification assigned to multiple projects. Do you think it's worth adding that functionality? Being able to retreive a notification and all of its assigned projects? The response looking something like this:

image

shreddedbacon: Yes, but not like that.

It'd be better to use the `project: [Project]` instread of `pid: Int` and return the projects objects, but this is a more involved change.

shreddedbacon: I think for now, remove `pid` and leave it the way it is.

Notifications are kinda painful association wise as they are

shreddedbacon: I'd see something like `getProjectsByNotification` and the input would be the notification, and the response would be just the projects.

This way it wouldn't require any schema changes beyond a new query

shreddedbacon:
I'm wondering if this is required or not now. Was there a reason for its need?

I'm wondering if `allNotifications` in #3671 could be extended to have an optional `name` input and omit the `type` input entirely? (or leave type too, and handle the permissions the same way this does)

CGoodwin90:
> I'm wondering if this is required or not now. Was there a reason for its need?
> 
> I'm wondering if `allNotifications` in #3671 could be extended to have an optional `name` input and omit the `type` input entirely? (or leave type too, and handle the permissions the same way this does)

No specific usecase, it was purely so we had some way to get the unassigned notifcations lost to the void. I like the idea of extending `allNotifications` to support the optional name, I can pull something together.

CGoodwin90:
Closing, replaced by https://github.com/uselagoon/lagoon/pull/3740

PR #3670, opened 2024-03-07
chore: update broker to rabbitmq:3.11.28

tobybellwood:
You need to upgrade to 3.11 to activate the feature flags required in 3.12, so will split this update over a couple of releases.

There is a manual step to be performed once on this version of RabbitMQ to enable all the feature flags

```
rabbitmqctl list_feature_flags
rabbitmqctl enable_feature_flag all
```

https://www.rabbitmq.com/docs/feature-flags#key-cli-tool-commands


shreddedbacon:
This will also need accompanying notes in the release pages in the docs to indicate the requirement (even though we do recommend people update to every minor version, you never know :see_no_evil: )

shreddedbacon:
Is there going to be a manual step that users will need to perform on the release that this would appear in?

tobybellwood:
> Is there going to be a manual step that users will need to perform on the release that this would appear in?

Yes - after this release has been deployed,, all the feature flags will need to be enabled via the command line before progressing to the 3.12 release out next.

shreddedbacon:
> Yes - after this release has been deployed,, all the feature flags will need to be enabled via the command line before progressing to the 3.12 release out next.

Hmm, it would be nice if we can figure out a way to avoid it being a manual step

tobybellwood:
Example output on a vanilla 3.11 install enables all feature flags not enabled by default 
```
check broker is running
enabling any disabled feature flags
 - enabling direct_exchange_routing_v2
 - enabling feature_flags_v2
 - enabling listener_records_in_ets
 - enabling stream_sac_coordinator_unblock_group
 - enabling stream_single_active_consumer
 - enabling tracking_records_in_ets
all feature flags enabled
```

shreddedbacon (approved):
Based on the use within the charts repository https://github.com/uselagoon/lagoon-charts/pull/666. This looks good

PR #3671, opened 2024-03-08
Feature: Add `getAllNotifications` query

CGoodwin90:
Implements `getAllNotifications` allowing for all `notifications` to be retreived.


services/api/src/resources/notification/resolvers.ts
```diff
@@ -705,3 +705,23 @@ export const removeAllNotificationsFromAllProjects: ResolverFn = async (
   // TODO: Check rows for success
   return 'success';
 };
+
+export const getAllNotifications: ResolverFn = async (
+  root,
+  args,
+  { sqlClientPool, hasPermission }
+) => {
+  await hasPermission('notification', 'view');
```
shreddedbacon: We likely need a `notification:viewAll` permission created for this that is restricted to platform-owner.

shreddedbacon:
Can you merge/rebase main into this

services/api/src/typeDefs.js
```diff
@@ -579,6 +579,7 @@ const typeDefs = gql`
     contentType: String
     notificationSeverityThreshold: ProblemSeverityRating
     organization: Int
+    pid: Int
```
shreddedbacon: What is `pid` used for here? This query should only return top level notifications, not anything related to projects?

services/api/src/resources/notification/resolvers.ts
```diff
@@ -705,3 +705,23 @@ export const removeAllNotificationsFromAllProjects: ResolverFn = async (
...
+export const getAllNotifications: ResolverFn = async (
+  root,
+  args,
+  { sqlClientPool, hasPermission }
+) => {
+  await hasPermission('notification', 'viewAll');
+
+  const rows = await Helpers(sqlClientPool).selectAllNotifications();
+
+  const nids = await Helpers(sqlClientPool).selectAllProjectNotifications(rows);
```
shreddedbacon: I'm not sure if it is a good idea to also return the project notifications here, as the return data will be overly complicated

shreddedbacon: There would be a lot of duplicates and the only difference would be the `pid` field

services/api/src/typeDefs.js
CGoodwin90: Added that so we can show both project assigned and unassigned notifications. Happy to remove it though if we don't want to go that deep.

services/api/src/resources/notification/resolvers.ts
CGoodwin90: The project notifications shouldn't be returned. Only iterating over them to assign the project id to what were getting in `rows`, which is then returned.

services/api/src/typeDefs.js
shreddedbacon: Yeah, I think for this particular one it should be omitted.

We can revisit it later if required. But a `projectsbynotification` resolver would probably be a good fit here instead.

services/api/src/resources/notification/resolvers.ts
shreddedbacon: Yeah, I think strip it out, because a notification can be attached to 1 or more projects, so the `pid` would only ever be 1 of those projects

services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -239,6 +267,7 @@ function configure_keycloak {
 
     check_migrations_version
     migrate_to_custom_group_mapper
+    add_notification_viewAll
```
shreddedbacon: Minor nit, all lowercase

shreddedbacon (approved):
Handy addition to find all notifications

PR #3672, opened 2024-03-08
Chore: Remove Harbor Trivy integration from Lagoon Core.

### LINKED ISSUE #3082: Insights version 2 work - remove Lagoon Core Harbor/Trivy/Drutiny integration

bomoko:
This PR removes Trivy integration in core - since Harbor is no longer part of Lagoon core, and Insights has now taken the place of generating SBOM scan results.

It does the following:
* Removes the tables and code for harbor scan matches
* Deprecates calls in the GraphQL schema for the above
* Adds exceptions when these GraphQL endpoints are called
* Removes all webhook integrations and tests for Trivy payloads

Closes #3082


tobybellwood (approved):
This looks good to me, a neat and tidy excision of code that's no longer functional

PR #3675, opened 2024-03-12
feat: store user last accessed, and sshkey last used timestamps

shreddedbacon:
# General Checklist

# Database Migrations

Whenever a user accesses the API update the last_accessed attribute on the user.

Alternative approach could be to store this in the API DB to reduce calls to keycloak though.

```
query allUsers {
  allUsers{
    email
    lastAccessed
  }
}
# result
{
  "data": {
    "allUsers": [
      {
        "email": "guest@example.com",
        "lastAccessed": "2024-03-12T00:14:32.000Z"
      }
    ]
  }
}
```
Also when the `api/keys` endpoint is hit, and a valid sshkey is detected, a new `last_used` field is updated to indicate when the key was last used.
```
      {
        "email": "maintainer@example.com",
        "lastAccessed": "2024-03-12T09:07:13.000",
        "sshKeys": [
          {
            "name": "maintainer-key",
            "lastUsed": "2024-03-12 09:06:07"
          }
        ]
      }
```


shreddedbacon:
ssh-portal-api will either need to use the API, or be updated to include the function to update the last used timestamp when an ssh key is used

smlx:
Should the SSH key last used also include the last time it was used to generate a token? From what I can tell this PR doesn't do that but I could be wrong.

shreddedbacon:
> Should the SSH key last used also include the last time it was used to generate a token? From what I can tell this PR doesn't do that but I could be wrong.

In the legacy SSH service, it is called whenever the [authorize.sh](https://github.com/uselagoon/lagoon/blob/085cd831e8116a4ab85ad12ea75a36e563047813/services/ssh/authorize.sh#L24) script is called and hits the [/keys](https://github.com/uselagoon/lagoon/blob/user-last-accessed/services/api/src/routes/keys.ts#L21) endpoint in the API. So whenever a key is used and matches, it will get updated. Token, SSH, whatever.

tobybellwood (approved):
This looks good - tested ok - picked up and closed any minor nits already :-)

PR #3679, opened 2024-03-14
deprecation: delete backups when they are deleted

shreddedbacon:
# General Checklist

# Database Migrations

Deleted backups remain in the API DB, though there is no method of recovering from a backup that is deleted. It makes little sense to retain backups that are deleted in this case.

This announces the deprecation of deleted backups, and includes a migration that will remove deleted backups from the database.

Additionally, the `includeDeleted` option when querying backups will now do nothing, this will be removed in a future release.

### DIFF
--- a/docs/releases/2.20.0.md
+++ b/docs/releases/2.20.0.md
@@ -1,0 +1,12 @@
+#### Release Links
+
+!!! Info
+    This version has not been released yet
+
+## Upgrades
+There are no required actions or considerations with this release. As always, we suggest upgrading all minor versions.
+
+## Deprecations
+
+### Deleted Backups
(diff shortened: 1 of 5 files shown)


docs/releases/2.20.0.md
shreddedbacon, line 1: Since this file exists, it would then be in the docs once it is merged?

Would it make sense to put a section in the PR called `Deprecation Notices` and then when the release is ready, a single PR with deprecation notices gets done in the docs?

shreddedbacon: As discussed, we will just make a note that the release isn't released yet

tobybellwood (approved):
This works as expected, cleans up the db nicely.

PR #3682, opened 2024-03-21
Improvement: validate migrating labels for active standby ingress migrations

shreddedbacon:
# General Checklist

# Database Migrations

This refactors the activestandby task to perform validations of the migrating status label before migrations take place. It ensures that the label exists on the resources before they are migrated, and then post migration ensures that the label is reset.

This is done to help prevent the potential for race condition between the label being added to the ingress right before the migration of that ingress takes place.


tobybellwood (approved):
Testing has shown a minimization of race condition occurrence with the additional validation steps. We should still consider configuring external tooling to mitigate any risk of future race conditions

(edited to reflect minimization instead of elimination)

shreddedbacon:
Just to clarify, it doesn't eliminate the potential for a race

PR #3684, opened 2024-03-22
fix: only set the subgroups of a parentgroup to the users group

shreddedbacon:
# General Checklist

# Database Migrations

Just a quick fix that makes sure that only the users group is loaded as the subgroups of their parent group.


tobybellwood (approved):
tested working

PR #3705, opened 2024-04-04
FIX: fixes migration deletion logic for duplicate environment_services

### LINKED ISSUE #3704: Database migration for environment_services table in 2.18.0 is too agressive.
# Describe the bug

In #3652, the [SQL query to delete duplicate services](https://github.com/uselagoon/lagoon/pull/3652/files#diff-5e5ca086a87929246fd0a98184a632ebe84d2543ce53857e6cfc0598c1e73ec7R7) doesn't take into account the environment the services are in. This results in ALL duplicate service names being deleted instead of duplicates per environment. The result is that tasks that require a `cli` service fail with `Environment <id> has no service cli`.

bomoko:
This PR changes the logic of the existing migration (described in #3704 )
This fix is done in the actual migration because
1. If the current migration has already been run, there is nothing for a brand new migration to do
2. If the current migration hasn't been run, running it will cause data loss. We could no-op the existing migration, but that would just mean a second migration with complex logic to determine which version of the current migration (broken or no-op) had been run.

This closes #3704

### DIFF
--- a/services/api/database/migrations/20240114000000_environment_services.js
+++ b/services/api/database/migrations/20240114000000_environment_services.js
@@ -4,7 +4,7 @@
  */
 exports.up = async function(knex) {
     return knex.schema
-    .raw(`DELETE es1 FROM environment_service es1 INNER JOIN environment_service es2  WHERE  es1.id < es2.id AND es1.name = es2.name;`)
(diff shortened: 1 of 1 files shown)


rocketeerbkw (approved):
Tested the new migration with the same data set that surfaced the issue and it's working as expected now.

PR #3720, opened 2024-05-07
refactor: change default quotas to unlimited for organizations

shreddedbacon:
# General Checklist

# Database Migrations

Just changes the route quota default to be unlimited by default. 

This is mainly because currently the route quota can only be enforced per environment by the build process. With the default being 5, any builds would fail regularly if an environment has more than 5 routes, so when an organization is created, this consideration needs to be made.

Since most people probably wouldn't be doing any route enforcement, it makes more sense for now to make it unlimited and it can be used as required until a better solution can be found (routes in the API, other...)

Updated to include all quotas default to unlimited, this way when an organization is created there are no default limitations, but the creator of the organization has some ability to enforce things as required.

Also updated to include a created timestamp on organizations


tobybellwood:
This makes sense.

 I was also wondering if we should or could make quota_environment a multiple of quota_project - you could end up creating an org with less envs than project limits.

shreddedbacon:
> This makes sense.
> 
> I was also wondering if we should or could make quota_environment a multiple of quota_project - you could end up creating an org with less envs than project limits.

Yeah, I've thought about this too. But then I get into the "what if one project needs a higher env quota than another project?" and it just becomes a more complex quota solution.

Something like default is `quota_environment` is per project, rather than overall. But then the ability to define quota overrides for specific things, similar to how deploytargetconfigs works. This way we could implement in a future step without breaking the current quota mechanisms that may suit majority of organizations.

shreddedbacon:
For now, this PR is just to reset the route quota. We can discuss quota extensions/refactoring in a separate issue

shreddedbacon:
Updated this to reset all quotas for organizations. We will leave these at unlimited for now, but we will look to explore quota rulesets as a feature in the future for finer control over quotas.

The default quotas will remain as they are for now as limits on overall resources (except routes)

shreddedbacon:
Another to add a created timestamp on the organization

bomoko (approved):
Changes make sense.

PR #3721, opened 2024-05-10
feat: add query-group permission to the service-api client

smlx:
With the recent change in Lagoon to move project membership of groups
from Keycloak group annotations into the Lagoon API DB, the
ssh-portal-api can no longer rely on group annotations embedded in user
tokens to extract project-group membership.

Since v0.35.0, ssh-portal-api gets the project membership information
from the Lagoon API DB. But that is stored as group IDs, not group
names. So to map group IDs back to group names the ssh-portal-api now
queries Keycloak for a list of groups (IDs and names).

This new permission allows the service-api client used by ssh-portal-api
to query the Keycloak groups API.

### DIFF
--- a/services/keycloak/lagoon-realm-2.16.0.json
+++ b/services/keycloak/lagoon-realm-2.16.0.json
@@ -528,6 +528,11 @@
       "realmRoles": [
         "default-roles-lagoon"
       ],
+      "clientRoles": {
+        "realm-management": [
+          "query-groups"
+        ]
+      },
       "notBefore": 0,
       "groups": []
     }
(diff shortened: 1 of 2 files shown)


smlx:
Does this get automatically applied on upgrade? Or do we need to add instructions to the release notes to manually update the permissions for the service account in keycloak?

shreddedbacon (changes requested):
You'll also need to add a migration to the startup script like this https://github.com/uselagoon/lagoon/blob/ssh-portal-group-db/services/keycloak/startup-scripts/00-configure-lagoon.sh#L186-L205 and insert it after this https://github.com/uselagoon/lagoon/blob/ssh-portal-group-db/services/keycloak/startup-scripts/00-configure-lagoon.sh#L233

This is because the realm import only runs on first time installations. Existing installations still need to run migrations for now.

shreddedbacon (approved):
Thanks!

PR #3727, opened 2024-05-21
fix: switch the priority, development should be lower than production

shreddedbacon:
# General Checklist

# Database Migrations

Just a small fix to switch the priority, the defaults should be `production=6` and `development=5`.

1 being the lowest, 6 being higher. Thus production environment deployments should be higher priority than development.

If you've not modified the build priority ever, then [this snippet](https://gist.github.com/shreddedbacon/33f291ff8c915ff19eed35b68cd16f80) can be used to reset the priority to be correct


tobybellwood (approved):
Nice and easy, easy and nice!

Issue #3504, opened 2023-08-03
Add support for `DRUPAL_PACKAGE_MANAGER_NOT_SUPPORTED_HELP_URL` env variable


Schnitzel:
Drupal is working on adding automated updates to Drupal Core https://www.drupal.org/project/automatic_updates which basically means Drupal will try to update it's code base automatically when a new version of Core or Modules exist.

Lagoon does not have support for this, for 3 reasons:
1. Automatic Updates need write permissions in the filesystem to actually update the code base, our containers don’t allow you to change the filesystem in a running container, so the upgrade will fail.
1. Even if Lagoon would allow a customer to change the file system you could not commit back the code into the git repo of the code base (as in the container you don’t have git commit access)
1. with containers the application code base actually exists multiple times in each container (nginx, php, cli), to properly upgrade this you would need to upgrade all 3 places at the same time (!) which is impossible

Therefore Lagoon cannot support this. Drupal actually will check for write permissions first before attending an update and will fail, so it can't happen that sites are in a weird state. Nonetheless Drupal added a functionality that allows a Hosting Platform to disable the auto update functionality (actually all of package manager functionality) and show the user a link to documentation why this feature is not supported, see https://git.drupalcode.org/project/automatic_updates/-/commit/dee9269

In short this involves setting an env variable `DRUPAL_PACKAGE_MANAGER_NOT_SUPPORTED_HELP_URL` with an URL to a documentation that explains why this does not work.

We should add support for this in Lagoon. I'm not sure how exactly we want to set this env variable: Inject in lagoon-env Config Map? 
One thing to keep in mind: technically auto updates work in local development environments as there the issues 1-3 above are not existent (write access exists, git access exists, code base is shared through the host), so adding the env to the Lagoon Base Images might not be the best solution.

rocketeerbkw:
Attempting to test automatic updates, I've found a number of other reasons why it can't run OOTB:

* The underlying system which does the updates is the `package_manager` module which has a dependency on `composer`. Since we only install composer in the cli image, Drupal running in the php container can't actually do the updates

    ![image](https://github.com/uselagoon/lagoon/assets/92499/3d09f27a-4853-4e41-a4d9-b316fae3685d)

* There is also a requirement for `rsync` to be installed in the php image which it isn't by default
* The package_manager module doesn't work on alpine linux https://www.drupal.org/project/automatic_updates/issues/3395782
*  Any module that scaffolds files will disable package_manager
    
[...]

PR #3729, opened 2024-05-24
Fixes some wording in documentation related to Drupal automatic updates

bryangruneberg:
- Simple fix
- PR title should be good for release notes

### DIFF
--- a/docs/applications/drupal/automatic-updates.md
+++ b/docs/applications/drupal/automatic-updates.md
@@ -23,7 +23,7 @@ The following update methods been disabled by Lagoon.
 ## Drupal Automatic Updates
 
 The [Automatic Updates](https://www.drupal.org/project/automatic_updates)
-contrib module is disabled by and it will also be disabled when it moves
+contrib module is disabled by Lagoon. The module will also be disabled when it moves
 into Drupal core.
 
 ## Drush


bomoko:
I'm trying to see if this change is true - I can't quite tell if we've done any work on Lagoon itself to disable automatic updates, or if it's just updated when lagoon integrations is installed (see https://github.com/uselagoon/lagoon/issues/3504)

docs/applications/drupal/automatic-updates.md
```diff
@@ -23,7 +23,7 @@ The following update methods been disabled by Lagoon.
 ## Drupal Automatic Updates
 
 The [Automatic Updates](https://www.drupal.org/project/automatic_updates)
-contrib module is disabled by and it will also be disabled when it moves
+contrib module is disabled by Lagoon. The module will also be disabled when it moves
```
rocketeerbkw, line 26: I think I meant to say "disabled by default." We will use an environment variable to disable, so users would be able to override that if they want (for example, using/testing automatic updates locally).

tobybellwood: We should probably explain here how it is disable - what the variable is, and where it is set

shreddedbacon:
Is this still required? Seems not?

PR #3733, opened 2024-06-06
feat: add new organization admin role

shreddedbacon:
# General Checklist

# Database Migrations

This adds a new role for organizations that allows a user to be granted `admin` permission, which is the same as `owner` except that admins cannot make changes to the management users of an organization, ie, they will not be able to add or remove other users as organization admins or owners. 

The role is only designed to allow interaction with projects, groups (and the group users), and notifications.

When adding a user to an organization as a manager, there is a new boolean called `admin`, used in a similar way to the `owner` boolean. Only `platform-owner` and `organization-owners` will be able to add users as admins.

```
addUserToOrganization(
	input: {
		user: {
			email: "orgadmin@example.com"
		}
		organization: 1
		admin: true
	}
) {
	id
}
```

Also, I have renamed the realm import file to better represent that it is a base import, rather than a specific version.

### DIFF
--- a/services/api/src/resources/user/resolvers.ts
+++ b/services/api/src/resources/user/resolvers.ts
@@ -221,7 +221,7 @@ export const deleteUser: ResolverFn = async (
 // addUserToOrganization adds a user as an organization owner
 export const addUserToOrganization: ResolverFn = async (
   _root,
-  { input: { user: userInput, organization: organization, owner: owner } },
+  { input: { user: userInput, organization: organization, admin: admin, owner: owner } },
   { sqlClientPool, models, hasPermission, userActivityLogger },
 ) => {
 
@@ -238,6 +238,7 @@ export const addUserToOrganization: ResolverFn = async (
   let updateUser = {
     id: user.id,
     organization: organization,
+    admin: false,
     owner: false,
   }
   if (owner) {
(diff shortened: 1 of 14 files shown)


services/api/src/resources/user/resolvers.ts
```diff
@@ -246,9 +247,16 @@ export const addUserToOrganization: ResolverFn = async (
...
   } else {
-    await hasPermission('organization', 'addViewer', {
-      organization: organization
-    });
+    if (admin) {
+      await hasPermission('organization', 'addOwner', {
+        organization: organization
+      });
+      updateUser.admin = true
+    } else {
```
tobybellwood, line 255: This isn't being set - so admins don't get the admin bit set (but owners seemingly do?
```
{
	"data": {
		"allOrganizations": [
			{
				"id": 1,
				"name": "lagoon-demo-organization",
				"projects": [
					{
						"name": "lagoon-demo-org"
					}
				],
				"owners": [
					{
						"email": "orgowner@example.com",
						"owner": true,
						"admin": true
					},
					{
						"email": "orgviewer@example.com",
						"owner": null,
						"admin": null
					},
					{
						"email": "orgadmin@example.com",
						"owner": null,
						"admin": null
					}
				]
			}
		]
	}
}
```

shreddedbacon: :eyes:

shreddedbacon: [this is the fix](https://github.com/uselagoon/lagoon/pull/3733/files#diff-9601ce160bccb3faab48a4a65bd13ff8250a9e35949b996d62ff08fe33f8218eR323), was a copypasta error before where it was adding it to `filteredOwners` instead of `filteredAdmins`

tobybellwood (approved):
This works as advertised. Nice changes to the seed-data make it easier to understand what's happening

PR #3738, opened 2024-06-12
fix: return an error if organization doesnt exist on addproject

shreddedbacon:
# General Checklist

# Database Migrations

Just a simple fix to make sure an unauthorized error is returned if an organization doesn't exist when adding a project. Previously it would get to the check for quota and fail with an obscure error about the project quota.
```
Cannot read properties of undefined (reading 'quotaProject')
```


tobybellwood (approved):
SImple change, increases usability

PR #3740, opened 2024-06-13
Change: Extends allNotifications with optional filters

CGoodwin90:
# General Checklist

# Database Migrations

Extends `allNotifications` to have an optional `name` & `type` input.

### DIFF
--- a/services/api/src/resources/notification/resolvers.ts
+++ b/services/api/src/resources/notification/resolvers.ts
@@ -707,7 +707,7 @@ export const getAllNotifications: ResolverFn = async (
     rows = await query(sqlClientPool, Sql.selectNotificationByNameAndType(args.name, args.type));
 
     if (rows.length === 0) {
-      throw new Error(`No notification found for ${args.name} & ${args.type}`);
+      throw new Error(`Unauthorized: You don't have permission to "view" on "notification"`);
     }
 
     if (rows.length > 0) {
--- a/services/api/src/resources/notification/resolvers.ts
+++ b/services/api/src/resources/notification/resolvers.ts
@@ -701,9 +701,43 @@ export const getAllNotifications: ResolverFn = async (
   args,
   { sqlClientPool, hasPermission }
 ) => {
+  let rows;
+
+  if (args.name && args.type) {
+    rows = await query(sqlClientPool, Sql.selectNotificationByNameAndType(args.name, args.type));
+
+    if (rows.length > 0) {
+      rows[0].type = args.type;
+    }
+
+    await hasPermission('notification', 'view', {
+      notification: rows[0].id,
+    });
+
+    return rows;
+  }
+
+  if (args.name) {
+    await hasPermission('notification', 'viewAll');
+
+    const rows = await Helpers(sqlClientPool).selectAllNotifications(args.name);
+
+    if (rows.length === 0) {
+      throw new Error(`No notification found for ${args.name}`);
+    }
+
+    return rows;
+  }
+
+  if (args.type) {
+    await hasPermission('notification', 'viewAll');
+
+    const rows = await query(sqlClientPool, Sql.selectAllNotifications(args.type));
+
+    return rows;
+  }
+
   await hasPermission('notification', 'viewAll');
-
-  const rows = await Helpers(sqlClientPool).selectAllNotifications();
-
+  rows = await Helpers(sqlClientPool).selectAllNotifications();
   return rows;
 };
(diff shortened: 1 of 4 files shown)


services/api/src/resources/notification/resolvers.ts
```diff
@@ -711,9 +711,43 @@ export const getAllNotifications: ResolverFn = async (
   args,
   { sqlClientPool, hasPermission }
 ) => {
-  await hasPermission('notification', 'viewAll');
+  let rows;
+
+  if (args.name && args.type) {
+    rows = await query(sqlClientPool, Sql.selectNotificationByNameAndType(args.name, args.type));
+
```
shreddedbacon, line 708: need to throw an error if no rows

shreddedbacon: or return `null` or something

services/api/src/resources/notification/resolvers.ts
```diff
@@ -711,9 +711,55 @@ export const getAllNotifications: ResolverFn = async (
...
+    if (rows.length === 0) {
+      throw new Error(`No notification found for ${args.name} & ${args.type}`);
```
shreddedbacon: To prevent some issues with enumeration etc, it would be better to throw a generic unauthorized error like this
```
throw new Error(`Unauthorized: You don't have permission to "view" on "notification"`);
```

services/api/src/resources/notification/resolvers.ts
```diff
@@ -711,9 +711,55 @@ export const getAllNotifications: ResolverFn = async (
...
+  }
+
+  if (args.name) {
+    await hasPermission('notification', 'viewAll');
+
+    rows = await Helpers(sqlClientPool).selectAllNotifications(args.name);
 
-  const rows = await Helpers(sqlClientPool).selectAllNotifications();
+    if (rows.length === 0) {
+      throw new Error(`No notifications found for ${args.name}`);
```
shreddedbacon, line 730: As this is an admin only query, this error is probably fine to throw

shreddedbacon: Same with the other two errors further down, as they're covered with a `hasPermission('notification', 'viewAll');`

shreddedbacon (changes requested):

shreddedbacon:
Just re-based and pushed (didn't change any code) before finalising review

shreddedbacon:
This works fine, but I'm wondering if this is the best approach to this now. 

The `name` and `type` filters are nice, but I don't know if this is something that is really useful for anyone that isn't a platform-owner. Type is also sort of irrelevant because of the way the resolver works with fragments.

With organizations also being a thing, the likelihood of someone using this that isn't a platform-owner is probably low.

I can't recall if there was a specific reason for these changes or not though, but maybe we can get away without this change for now?

CGoodwin90:
> This works fine, but I'm wondering if this is the best approach to this now.
> 
> The `name` and `type` filters are nice, but I don't know if this is something that is really useful for anyone that isn't a platform-owner. Type is also sort of irrelevant because of the way the resolver works with fragments.
> 
> With organizations also being a thing, the likelihood of someone using this that isn't a platform-owner is probably low.
> 
> I can't recall if there was a specific reason for these changes or not though, but maybe we can get away without this change for now?

It was just so we had a way to retrieve unassigned notifications, right now once created they're essentially lost to the void until assigned to a project. I think there were some commands in the cli where this might be useful, but not super necessary. Happy to proceed without it for now though.

shreddedbacon:
> It was just so we had a way to retrieve unassigned notifications, right now once created they're essentially lost to the void until assigned to a project. I think there were some commands in the cli where this might be useful, but not super necessary. Happy to proceed without it for now though.

Ahh that's right. Yeah, I think it might need some adjustments, so maybe we can bump this to 2.22.0 instead. I'll add some review comments with some suggested changes.

shreddedbacon:
I'm not sure if this is required now. The main change here is just adding name and type filtering. Currently, it seems possible to retrieve all notifications regardless of whether it is assigned to a project or not.

The filtering commands really only benefit platform owners, and unless a user knows exactly the name of the notification they are looking for, it will just error out.

Since the response notifications are typed, it is already somewhat possible to filter for the type by only recalling the type you want in the resolver. The only thing here that could be useful would be the `name` filter. But someone could just use tools on the output from the API/CLI to search for these too.

Thoughts?

CGoodwin90:
Yeah happy to close this one, filtering can be done on the output easy enough.

PR #3741, opened 2024-06-13
docs: update available base images

tobybellwood:
A small maintenance docs update


shreddedbacon (approved):
I checked with Tony Belfort just to make sure my understanding of the image updates are correct, and this looks all good to me

Edit: Tony approached me and told me I got his name wrong, so I have now corrected this.

PR #3744, opened 2024-06-17
fix: remove usergroup injection from org project groups to prevent duplicates

shreddedbacon:
# General Checklist

# Database Migrations

If an organization owner was a member of a group and also the organization owner, when viewing the groups of a project it would result in duplicate groups appearing due to the way the resolver consolidated groups.

This fix is to remove the part of the resolver where the users groups are injected into the resulting projects groups so that only groups attached to the project are returned.

Due to the way groups work in Lagoon prior to organizations, there exists the possibility that a group that did not exist within the organization could have been attached to a project, however this is unlikely (but not impossible)


tobybellwood (approved):
This resolves the issue as experienced.

PR #3745, opened 2024-06-18
test: Update upstream branch and kubernetes versions

tobybellwood:
This PR consumes the upstream Lagoon-v.2.20.0 release branch from lagoon-charts, as well as updating the installed versions of packages.

Of note:
* Kubernetes 1.30 is now used for testing
* docker compose v2 is now the default for `make up` commands in the makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -384,13 +384,13 @@ api-logs-development: build-ui-logs-development
 
 ## CI targets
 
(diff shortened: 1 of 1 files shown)


shreddedbacon (approved):
Yep, makes sense, bonus that tests pass :heavy_check_mark:

PR #3747, opened 2024-06-19
Update default value for Varnish cache in docs

kasperg:
According to [Docker](https://github.com/uselagoon/lagoon-images/blob/main/images/varnish/6.Dockerfile#L103) [files](https://github.com/uselagoon/lagoon-images/blob/main/images/varnish/7.Dockerfile#L55) CACHE_SIZE has a default value of 500M. The documentation says the default value is 100M.

Update the documentation to refect the actual values.


tobybellwood:
Thanks @kasperg - nice pickup. I've also just realised that it's 950M (and a different backend) in the persistent versions, but I can do those updates

PR #3753, opened 2024-06-27
chore: remove all dangerous deleteAll and removeAll mutations

shreddedbacon:
# General Checklist

# Database Migrations

This removes all the dangerous deleteAll and removeAll mutations from the API. These were only ever used during testing, and aren't used that often anymore in tests anyway.

Tearing down and recreating a Lagoon locally is much easier now than it used to be.


services/keycloak/startup-scripts/00-configure-lagoon.sh
```diff
@@ -256,7 +256,7 @@ function migrate_admin_organization_permissions {
 
   echo Configuring Organization admin permissions
 
-  echo Delete existing organization management
+  echo Delete deleteall sshkeys
```
tobybellwood: que?

shreddedbacon: :eyes:

shreddedbacon: yah previous merge conflict miss

tobybellwood (approved):
Works as expected, upgrades track ok too. LGTM

PR #3755, opened 2024-06-28
test: reuse sshkeys from ci for local dev

tobybellwood:
This PR re-purposes the two-previously unused SSH keys from the ci users to users set up for all local-dev lagoon.

It also renames the keys for clarity.


shreddedbacon (approved):
Fine by me if the tests pass

PR #3756, opened 2024-06-28
feat: add inviteUser boolean to addUserToGroup

shreddedbacon:
# General Checklist

# Database Migrations

Just a simple one to make adding users to groups a little bit nicer if the user doesn't exist. Adds an `inviteUser` boolean to `addUserToGroup` input that will create the user account if one doesn't exist.

The UI could then expose a checkbox when adding a user to a group that asks if they want to invite the user if the user doesn't exist.

If `inviteUser` is set and the account does not already exist, it will be created, and a password reset email will be sent.


tobybellwood (approved):
This works as expected, with the appropriate safeguards around permissions

PR #3758, opened 2024-06-30
chore: adds fix-key-perms to set sshkey perms correctly for local-dev

tobybellwood:
This PR implements a make routine to set the correct 600 permission on the local-dev SSH private keys. Github stores files as 644 by default, so the private keys need to be chmodded to 600 before they can be used for SSH'ing.

Given how rarely this is needed, I've not added the routine to any others, and kept it standalone

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -347,6 +347,13 @@ down:
 kill:
 	docker ps --format "{{.Names}}" | grep lagoon | xargs -t -r -n1 docker rm -f -v
 
+.PHONY: fix-key-perms
+fix-key-perms:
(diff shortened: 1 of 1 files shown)


shreddedbacon:
I've not encountered an issue with this, but then I don't use those keys outside of the tests.

Happy to see it merge though as its a just a small addition. But maybe it could be worth adding some comments in the makefile to its purpose/usage

shreddedbacon:
Not sure if this is still required. Can we close it out?

The [local-stack generates keys](https://github.com/uselagoon/lagoon/blob/v2.29.2/Makefile#L1201)

PR #3759, opened 2024-07-02
fix: check if user has attributes before accessing attributes

shreddedbacon:
# General Checklist

# Database Migrations

#3675 introduced `last_accessed` attribute to users. Most users get an attribute added to them when they are created, but users created via an IDP do not. This is just a quick fix to check that attributes exist before trying to access the attribute value.

Also moves the lastaccessed call to apollo so it is only called once on the initial request, rather than every haspermission check


tobybellwood (approved):
LGTM

PR #3763, opened 2024-07-04
feat: add new add/remove admin to organization, deprecate old add/remove

shreddedbacon:
# General Checklist

# Database Migrations

The old process of adding a user as an owner of an organization was previously an `owner` flag, then with #3733 added `admin`. This is now different to other things that have a `role` concept.

This PR introduces a new OrganizationRole type and 2 new mutations that deprecate the older functions with the new types.
Both old and new can still be used, but the recommendation would be to use the new ones for simplicity.

## Changes
* `addUserToOrganization` -> `addAdminToOrganization`
  * The `organization` input is now an `id` or `name` input, either can be specified
  * The `role` input is an ENUM of `VIEWER`, `ADMIN`, `OWNER`
* `removeUserFromOrganization` -> `removeAdmimFromOrganization`
  * The `organization` input is now an `id` or `name` input, either can be specified
* The `OrgUser` type now has `organizationRole` which will return the users role
  * the `admin` and `owner` fields are deprecated, but still populated for backwards compatability


tobybellwood (approved):
Given this a run-through, all looks ok to me. Deprecations handled well, sensible consistency change

PR #3776, opened 2024-07-22
fix: permission checks for updating or viewing projects in orgs

shreddedbacon:
# General Checklist

# Database Migrations

## The problem
Prior to organizations being introduced, users with `update:project` on projects had permission to add, edit, and delete `DeployTargetConfigs` on projects. The roles that allow this currently are `maintainer` and `owner` roles (and higher level roles).

After the organizations roll out, the `update:project` permission on the above interactions was changed (probably incorrectly) to only allow organization administrators with the `updateProject:organization` permission to make changes to these configs.

Other roles that allow `update:project` still permit users with this permission to perform the same changes they used to do. This includes directly modifying the project settings, adding/removing metadata, and also modifying an environment to use a different deploytarget directly.

This permission change is to still allow users that previously had the `update:project` permission to modify `DeployTargetConfigs` to perform these actions, while still allowing organization adminstrators the ability to do so.

Modifying `DeployTargetConfigs` of a project in an organization will still require the deploytargets to be available to the organization as before, there is no change to this behaviour.

## Changes

* `checkOrgProjectViewPermission`
  * this is changed to check firstly if a user has the `view:project` permission, if not it will check if the project is in an organization or not and will then perform the organization check.
  * previous behaviour was to check organization permission, and then fall back to `view:project`
* `checkOrgProjectUpdatePermission`
  * this is changed to now check if a user has the `update:project` permission, if not it will check if the project is in an organization or not and will then perform the organization check.
  * previous behaviour was to check organization permission if in an org, but no fall back to `update:project`. It would only check `update:project` if the project was not in an organization.

### DIFF
--- a/services/api/src/resources/project/helpers.ts
+++ b/services/api/src/resources/project/helpers.ts
@@ -14,21 +14,22 @@ export const Helpers = (sqlClientPool: Pool) => {
       // then falling through to the default project view for general users
       const rows = await query(sqlClientPool, Sql.selectProject(pid));
       const project = rows[0];
-      if (project.organization != null) {
-        try {
+      try {
+        // finally check the user view:project permission
+        await hasPermission('project', 'view', {
+          project: project.id
+        });
+        return
+      } catch (err) {
+        if (project.organization != null) {
           await hasPermission('organization', 'viewProject', {
             organization: project.organization
           });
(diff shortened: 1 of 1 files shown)


bomoko (approved):
makes sense.

PR #3781, opened 2024-07-25
tests: Rework/retire/update test suite

tobybellwood:
This PR undertakes a number of updates to the default Lagoon test suite:

* It retires the legacy `drush sa` tests that utilise the drush-alias service, as this is no longer actively developed in Lagoon-core
* It updates the other `drush la` test to use the latest Drupal release (currently 10) - based on the standard drupal-base example
* It updates all the node.js based tests to node22 and expressJS 4.19
* It updates the dockerfile and docker-compose.yml syntax in the test suite to meet modern standards.


Jenkinsfile
```diff
@@ -179,7 +183,7 @@ pipeline {
         stage ('3: run third test suite') {
           steps {
             catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
-                sh script: "make -j$NPROC k3d/retest TESTS=[gitlab,github,bitbucket,drush] BRANCH_NAME=${SAFEBRANCH_NAME}", label: "Running third test suite on k3d cluster"
+                sh script: "make -j$NPROC k3d/retest TESTS=[gitlab,github,bitbucket,python,drush] BRANCH_NAME=${SAFEBRANCH_NAME}", label: "Running third test suite on k3d cluster"
```
shreddedbacon, line 186: Did we remove the python test in the past, and this is adding it back?

tests/tests/drush.yaml
```diff
@@ -12,31 +12,9 @@
...
-    drupal_version: 9
-    db: mariadb-single
-    php_version: 8.0
     git_repo_name: drush.git
     project: ci-drush-la-{{ cluster_type }}
 
 - ansible.builtin.import_playbook: api/delete-project.yaml
   vars:
     project: ci-drush-la-{{ cluster_type }}
-
```
shreddedbacon, line 24: Is this just removing the old drush site aliases because they don't work in newer versions of drupal we have in examples?

shreddedbacon (approved):
Everything here seems to be fine. Couple questions to confirm some unknowns for me, but not blocking questions.

tests/tests/drush.yaml
shreddedbacon: :see_no_evil: reading the PR description answers this :smiling_face_with_tear:

Jenkinsfile
tobybellwood: yup - it uses a slightly different branch push/pull - I will work out how to unify them, but in this release, at least it's back up to currency.

PR #3782, opened 2024-07-26
build: Update golang used in services to 1.22

tobybellwood:
This PR updates the golang version in use by the lagoon services to 1.22

The following issues were discovered:

* the `github.com/Khan/genqlient v0.3.0` module in services/workflows could not be updated due to a change in the MakeRequest structure in v0.5.0 that will require attention
* the `github.com/cheshir/go-mq/v2` module in services/workflows fails vet with a "passes lock by value" error - unsure about this
* the `k8s.io/*` modules in task-activestandby fail tests (around IngressList structure) when updated to 1.30.x - this will require attention (I think to modify the expected result JSON)

edit: raised the above issues for tracking in #3783


shreddedbacon (approved):
Nothing standing out to me, looks good.

shreddedbacon (approved):

PR #3784, opened 2024-07-30
fix: check for only matching fingerprint on key lookups

shreddedbacon:
# General Checklist

# Database Migrations

Updates the query when checking for ssh keys to query for the specific fingerprint, rather than all keys.

Also includes a minor logic fix on the platformowner/viewer permission check for groups


tobybellwood (approved):
Tested working, LGTM

PR #3788, opened 2024-08-05
fix: updateDeployTargetConfig to not use params that aren't being updated

rocketeerbkw:
# General Checklist

# Database Migrations

Attempting to update a deploy target config may fail with the following error: `Error: graphql: The provided deploytarget is not valid for this organization`.

This happens when the deploy target is not passed in as a value to update. The fix is to check if this is being updated before checking the permissions.

With correct typing, this would've been caught in strict mode. Even the smallest "strictNullChecks" typescript option looked like too much work to enable at this time, so I just updated the types so that this will be valid if we ever get to that point in the future.

# Closing issues

n/a

### DIFF
--- a/services/api/src/resources/deploytargetconfig/resolvers.ts
+++ b/services/api/src/resources/deploytargetconfig/resolvers.ts
@@ -1,12 +1,10 @@
 import * as R from 'ramda';
+import { Pool } from 'mariadb';
 import { ResolverFn } from '../';
-import { Environment } from '../../resolvers';
 import { Helpers as environmentHelpers } from '../environment/helpers';
 import { isPatchEmpty, query, knex } from '../../util/db';
-import { getProjectByEnvironmentId } from '../project/resolvers';
 import { Helpers } from './helpers';
 import { Sql } from './sql';
-import { deployTargetBranches } from '@lagoon/commons/src/deploy-tasks';
 import { Sql as EnvironmentSql } from '../environment/sql'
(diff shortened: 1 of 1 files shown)


services/api/src/resources/deploytargetconfig/resolvers.ts
```diff
@@ -262,44 +260,68 @@ export const deleteDeployTargetConfig: ResolverFn = async (
...
+  deployTargetProjectPattern?: string;
+  pullrequests?: string;
+  weight?: number;
+}
+
 export const updateDeployTargetConfig: ResolverFn = async (
   root,
-  { input },
-  { sqlClientPool, hasPermission, userActivityLogger }
+  { id, patch }: { id: number; patch: UpdateDeployTargetConfigPatch },
```
shreddedbacon: updateDeployTargetConfig is broken for me when I use it locally, it complains with
```
Cannot destructure property 'branches' of 'patch' as it is undefined.
```

Patching locally with the following resolves the issue for me
```
  { input: {id, patch }}: {input: { id: number; patch: UpdateDeployTargetConfigPatch }},
```

shreddedbacon (changes requested):

services/api/src/resources/deploytargetconfig/resolvers.ts
rocketeerbkw: Thanks. My kingdom for auto-generated resolver types.

shreddedbacon (approved):
Works now :partying_face:

PR #3793, opened 2024-08-08
Docs: Removing Alanna from CoC and adding Lagoon alumni.

AlannaBurke:
Removes Alanna from CoC page and adds Alumni to maintainers.


tobybellwood (approved):
Thank you, and thank you for everything in the last 4-5 years :medal_sports:

PR #3794, opened 2024-08-14
fix: check if user has attributes before accessing attributes

shreddedbacon:
# General Checklist

# Database Migrations

In the `usersByOrganization` query, there are checks against attributes of a user from keycloak. If a user has no attributes though, this check fails.

This fix checks if attributes are defined before trying to access them.


tobybellwood (approved):
This makes sense, thanks!

PR #3795, opened 2024-08-21
Improve SSH docs around ptys and interactive shells

smlx:
There has been some confusion around why sometimes you don't get a shell prompt when using Lagoon SSH. This change aims to clarify what happens when you use ssh and when you need to force pty allocation.

I noticed that we advise people to specify port 22 even though that is the default. So I removed that bit.

I also took the opportunity to remove the advice to ignore host keys, since both the ssh core and ssh portal services provide static host keys these days. If you don't want to remove that yet I can push it in a separate PR.

# General Checklist

- [ ] ~Affected Issues have been mentioned in the Closing issues section~

# Database Migrations

- [ ] ~If your PR contains a database migation, it **MUST** be the latest in date order alphabetically~

# Closing issues

n/a

### DIFF
--- a/docs/interacting/ssh.md
+++ b/docs/interacting/ssh.md
@@ -104,7 +104,7 @@ ssh {% if defaults.sshport != 22 %}-p {{ defaults.sshport }} {% endif %}-t drupa
 
 ## Copying files
 
-The common case of copying a file into your `cli` pod can be acheived with the usual SSH-compatible tools.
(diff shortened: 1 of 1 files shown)


shreddedbacon:
The port needs to go back in.

The `{{ defaults.sshport }}` are used as a way to provide overrides for people that may not use port 22 and host their own version of the docs.

smlx:
Ok thanks for the tip. I added a conditional around the port templating.

tobybellwood (approved):
Thanks!

PR #3797, opened 2024-08-23
fix: add view-users role to the service-api client

smlx:
Later versions of Keycloak require view-users as well as query-groups
role permissions in order to query the groups in a realm.

This seems to be basically undocumented. :'(


shreddedbacon:
This makes sense. Covers off the change required to the realm import on initial setup, and the migration for existing installations.

shreddedbacon (approved):

PR #3798, opened 2024-08-23
chore: fix debug command

smlx:
The pemissions on the KUBECONFIG file in jenkins changed to 0600, so
just `sudo cat` it out instead of scp.


tobybellwood (approved):
yup - thanks for doing this - I'd been working around it for months!

PR #3799, opened 2024-08-26
docs: remove uptimerobot references and add note about monitoring

shreddedbacon:
# General Checklist

# Database Migrations

Removes references to UptimeRobot and updates documentation for the `monitoring-path` on routes to include a note that Lagoon does not provide monitoring capabilties out of the box, and that only labels and annotations are provided.

It is up to platform operators to use these labels and annotations to inform monitoring systems as they see fit.

As always though, people are free to add their own monitoring that they manage themselves outside of Lagoon if a platform operator does not offer a monitoring solution.

### DIFF
--- a/docs/concepts-basics/lagoon-yml.md
+++ b/docs/concepts-basics/lagoon-yml.md
@@ -275,12 +275,20 @@ that's important!):
 
 ### Monitoring a specific path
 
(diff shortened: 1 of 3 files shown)


tobybellwood (approved):
Thanks!

PR #3801, opened 2024-09-04
change: set keycloak frontend url in api for well-known discovery

shreddedbacon:
# General Checklist

# Database Migrations

This change requires [these changes](https://github.com/uselagoon/lagoon-charts/pull/682) in the charts.

The main things changing are to allow internal comms for the API to Keycloak, while still presenting the keycloak frontend url for the well-known discovery endpoint.

This still populates the forgot password reset emails with the publicly accessible Keycloak URL so that users will be able to perform password resets. This can be confirmed in both the `make up` and `make k3d/local-stack` and visiting either the mailhog/mailpit that is started and doing a password reset. In `make k3d/local-stack` a user is invited as part of the seeding steps, and checking in the mailpit that starts as part of that stack you'll see the reset email with the "public" URL for Keycloak.

In `make k3d/local-stack` there is the ability to access keycloak via https, but it will fail to load the resources because the frontend url for keycloak is set to the http endpoint in the [Makefile](https://github.com/uselagoon/lagoon-charts/blob/45c04f8bf6033188bcbcceaf7b0e121a7ff2f2d3/Makefile#L241). In production, this would be set to the `https` endpoint under a `keycloakFrontEndURL` value in the core values (but [defaults to the ingress with https](https://github.com/uselagoon/lagoon-charts/blob/45c04f8bf6033188bcbcceaf7b0e121a7ff2f2d3/charts/lagoon-core/templates/keycloak.deployment.yaml#L48) if not provided). This is unchanged behaviour, just calling this out here though.


tobybellwood (approved):
Tested working ok

PR #3803, opened 2024-09-10
feat: add lagoon-ui-oidc client

shreddedbacon:
# General Checklist

# Database Migrations

This creates a copy of the `lagoon-ui` client called `lagoon-ui-oidc` that has public authentication disabled. This client will eventually be used by the next UI version, while leaving the existing client in place for the current UI so that both versions of the UI could be used while development is undertaken, with the eventual goal of the original `lagoon-ui` client going away.

The [charts will need to be updated](https://github.com/uselagoon/lagoon-charts/pull/691) to include support for configuring the secret for this client with this variable `KEYCLOAK_LAGOON_UI_OIDC_CLIENT_SECRET`


tobybellwood (approved):
This is a step towards Uiv2

PR #3807, opened 2024-09-17
chore: update k8s versions and upstream branch

tobybellwood:
This PR updates the versions of K3D, Kubernetes and Kubectl to match the versions tested in lagoon-charts.

Additionally, it repoints the CHARTS_TREEISH back to main now v2.21 has been released


shreddedbacon (approved):
Makes sense

PR #3812, opened 2024-09-19
build: update rabbitmq/broker to 3.12.14

tobybellwood:
This PR updates the version of RabbitMQ used in the broker service to the latest release in the 3.12.x series.

We have to upgrade Minor version by minor version to ensure feature flags are updated


shreddedbacon (approved):
Reading https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.12.14 I'm not seeing anything that stands out as potentially breaking for us.

PR #3813, opened 2024-09-20
feat: add lagoon-cli keycloak client

shreddedbacon:
# General Checklist

# Database Migrations

This adds a keycloak client specifically for the lagoon-cli to use when the oauth integration is available to use. 

It only has `http://127.0.01` redirect uri support, which works with the random ports that the cli generates to handle the callback with.


tobybellwood (approved):
This is the building block for the next part of Lagoon CLI.

PR #3816, opened 2024-10-01
refactor: updates for api-db and keycloak-db to support MySQL8

tobybellwood:
# General Checklist

# Database Migrations

# Description

In order to support MySQL8 as well as MariaDB-based databases, a couple of schema changes are needed. These are mostly to control the size of indexes, and set enum types to be compatible.

There is no current change planned, but should a Lagoon admin wish to migrate to a cloud database for Lagoon **[NOTE THAT THIS IS NOT CURRENTLY POSSIBLE]**, these changes will have to be in place prior to migration. Additionally, MySQL strict mode is not supported (due to Lagoon's love of zero dates).

There are a couple of local-development additions to support mysql for testing etc.

# Requires

The following pullrequest is for the supporting chart changes.

* https://github.com/uselagoon/lagoon-charts/pull/745


shreddedbacon:
I tweaked this a bit to leave the existing docker-compose.yaml file alone and leverage make targets with a compose override for mysql specific changes to keep the overrides isolated.

This lets you now use the defaults `make up` or `make ui-logs-development` to spin up with a `mariadb` and use the `-mysql` suffixed versions of these to run with the mysql overrides in place.

Having the ability to do both allows for verifying that not only do the migrations succeed in mysql, that they hopefully also succeed in mariadb.

services/api/database/migrations/20220908065119_initial_db.js
```diff
@@ -9,18 +9,18 @@
...
                 table.string('name', 300).notNullable();
-                table.specificType('description', 'text').defaultTo('').notNullable();
+                table.specificType('description', 'text').notNullable();
                 table.string('image', 2000).defaultTo('');
                 table.string('service', 100);
                 table.string('type', 100).notNullable();
                 table.integer('environment');
                 table.integer('project');
-                table.string('group_name', 2000);
+                table.string('group_name', 300);
```
shreddedbacon, line 18: this is fine, keycloak appears to limit this to 256

services/api/database/migrations/20220908065119_initial_db.js
```diff
@@ -126,17 +126,17 @@
                 table.integer('environment');
                 table.string('severity', 300).defaultTo('');
                 table.decimal('severity_score', 1, 1).defaultTo(0);
-                table.string('identifier', 300).notNullable();
-                table.string('lagoon_service', 300).defaultTo('');
+                table.string('identifier', 100).notNullable();
```
shreddedbacon, line 129: i have no idea what the `identifier` field is used for to know if this will have any other impacts by reducing down to 100

services/api/database/migrations/20220908065119_initial_db.js
```diff
@@ -126,17 +126,17 @@
...
+                table.string('lagoon_service', 100).defaultTo('');
```
shreddedbacon, line 130: `lagoon_service` should be fine going to 100, pretty sure we're limited to 63 characters here anyway due to kubernetes labels and out `lagoon.sh/service` label usage. No idea what the docker-compose service name limitations are :hidethepain:

services/api/database/migrations/20220908065119_initial_db.js
```diff
@@ -126,17 +126,17 @@
...
                 table.string('source', 300).defaultTo('');
                 table.string('associated_package', 300).defaultTo('');
-                table.specificType('description', 'text').defaultTo('');
-                table.string('version', 300).defaultTo('');
+                table.specificType('description', 'text');
+                table.string('version', 100).defaultTo('');
```
shreddedbacon, line 134: I also suspect `version` is fine to reduce down, but :shrug:

shreddedbacon:
The only thing I haven't tried locally is starting on a MariaDB container, then attempting to change to MySQL. 

I'm going to guess this will be a real pain to script for local development since our compose stack isn't really designed with in place upgrading in mind. 

#3829 does offer the ability to do stable to unstable chart upgrades that could potentially test the viability of a MariaDB pod to MySQL pod upgrade in a helm pre or post upgrade task somehow. This would be super risky though. We may have to continue to support MariaDB pod only, but ensure any future migrations are valid for MySQL.

services/api/src/models/environment.ts
```diff
@@ -153,7 +153,7 @@ export const Environment = (clients: {
     if (
       deleted_date < interested_month_start &&
       moment(deleted_date).format('YYYY-MM-DD HH:mm:ss') !==
-        '0000-00-00 00:00:00'
+        null
```
shreddedbacon: i'm unsure about this particular one, would need more verification

shreddedbacon:
There might be some issues with MySQL strict mode that make this difficult to support alongside MariaDB. Mostly that the `timestamp` defaulting to `null` in strict mode defaults to `null` but in our MariaDB images it still sets to `0`. So we need to probably go with strict mode disabled. If we had a way to completely migrate from MariaDB to MySQL then we could move to properly supporting strict mode in both the migration scripts and the API. But for now, strict mode disabled with the changes in this branch appear to be valid for both MariaDB and MySQL 8.

Luckily(?) AWS appear to have [strict mode disabled by default in RDS MySQL 8](https://repost.aws/questions/QUZg91IPXvT_yGxfWqiI6DWA/is-it-possible-to-turn-off-strict-sql-mode-in-rds-for-mysql-v8-0#ANL_i7fdP5RRGYPCsFdcyX9Q). Unsure about other providers. But we could document this as a requirement when we start to think about supporting external DB providers in the charts.

services/api/database/migrations/20241231000100_mysql8_compatibility.js
```diff
@@ -0,0 +1,73 @@
...
+exports.up = async function(knex) {
+    return knex.schema
+    .alterTable('advanced_task_definition', function (table) {
+        table.specificType('description', 'text').notNullable().alter();
+        table.string('group_name', 300).alter();
+        table.specificType('command', 'text').alter();
+        table.timestamp('deleted').notNullable().alter();
+    })
+    .alterTable('environment', function (table) {
+        table.timestamp('deleted').notNullable().alter();
```
tobybellwood: Need to update deploy_base_ref and deploy_head_ref here to 250 (from 100)
https://github.com/uselagoon/lagoon/pull/3816/files#diff-94fc09ab7f90affd501ddddf3ae66b4d0b41b4dd22b9b2849f2f692d3b8f3d75R79-R80

services/api/database/migrations/20241231000100_mysql8_compatibility.js
```diff
@@ -0,0 +1,73 @@
...
+    })
+    .alterTable('environment_fact', function (table) {
```
tobybellwood: Missing ennvironment_backup deleted field update https://github.com/uselagoon/lagoon/pull/3816/files#diff-94fc09ab7f90affd501ddddf3ae66b4d0b41b4dd22b9b2849f2f692d3b8f3d75R101

shreddedbacon: We dropped the column in a previous migration https://github.com/uselagoon/lagoon/blob/ab7caf5d186d7c530f669732367d45fec4e1107d/services/api/database/migrations/20240507000000_delete_orphaned_backup_resources.js#L16

shreddedbacon:
One thing to be aware of with keycloak using mysql, is that it also has to connect to the lagoon api db and uses the hardcoded `LAGOON_DB_VENDOR` value. We'd need to make sure that if the api-db is changed to mysql, that we have a way to pass this to the keycloak pod too. Locally you've got this https://github.com/uselagoon/lagoon/pull/3816/files#diff-abd63f9e26078ca107dc1fa168d6cd47c47f8bac403e4fa46865129f0c7804c1R17 so there would need to be associated changes to the charts to know to pass this through. I know its not really part of this pullrequest though because this is just the first stage of being able to support mysql or an external provider generally

shreddedbacon:
I refactored the mysql overrides to make it simpler to switch, and also able to run through the entire test suite with mysql with the use of `DATABASE_VENDOR` in the [Makefile](https://github.com/uselagoon/lagoon/pull/3816/commits/d60014a95b18e71808970c823e911e4f64814514#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R98-R99). This way the image built and all the configuration options required are done during make without the need for a mysql specific docker compose file, or any other make targets.

Edit: Some minor chart changes are required to handle this switch. Just verifying the required changes to be as minimal/backwards compatible

shreddedbacon:
Found [some things](https://github.com/uselagoon/lagoon-images/pull/1190) in the upstream images that are a bit annoying and breaks the mysql image from working in tests and charts.

I [modified the charts](https://github.com/uselagoon/lagoon-charts/compare/mysql-image-support) with some minimal database vendor changes and added the overrides that are in the images pull request to this branch. Ideally, the images PR would be merged and we just consume those newer images when they would be released. 

I also added a password entrypoint in the api-db and keycloak-db so that the charts don't need to specify a vendor in the password, the vendoring happens in the entrypoint now. The api-db and keycloak-db chart values have vendor types defined defaulting to mariadb.

This branch has had the [vendor default](https://github.com/uselagoon/lagoon/pull/3816/commits/d60014a95b18e71808970c823e911e4f64814514#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R99) set to be `mysql` which the database migration and rollback action fails with currently. We will have to revert the default vendor back to `mariadb`, but it might be worth fixing the rollbacks to work with mysql too.  (EDIT 31 December: reverted the default back to `mariadb`)

I just want to ensure that the entire jenkins test suite runs through with images built on mysql and I'll feel much better about it.
Edit: tests [passed](https://ci.lagoon.sh/blue/organizations/jenkins/lagoon/detail/PR-3816/15/pipeline) with mysql, thats a good sign.

shreddedbacon:
Now that https://github.com/uselagoon/lagoon-images/pull/1190 has been merged, there are some bits left over that didn't need to exist anymore, now removed in [this commit](https://github.com/uselagoon/lagoon/pull/3816/commits/de6c3bdba5566412a5f58f010656cad786c19c87)

shreddedbacon (approved):
As discussed during our team meeting, everyone agreed this is fine to proceed with.

PR #3827, opened 2024-10-18
fix: use api-db for project and org group attributes instead of keycloak

shreddedbacon:
# General Checklist

# Database Migrations

Fixes some functions that were still checking group attributes for projects and organization from keycloak instead of the api database where this information is now stored.

The project and organization attributes being added to groups in keycloak were only done for legacy purposes and will be removed eventually.


tobybellwood (approved):
Been through this forwards and backwards, makes sense, and tests ok.

PR #3829, opened 2024-10-21
test: refactor local development for more configuration

shreddedbacon:
# General Checklist

# Database Migrations

Refactors the make targets to allow for setting up a local-stack with the latest stable version of the `lagoon-core`, `lagoon-remote`, and `lagoon-build-deploy` charts, instead of bleeding edge. This also allows for the local-stack to be spun up much faster for not having to build the images prior.

These changes allow for someone install a local-stack with the stable version, then once it is running perform a Lagoon upgrade by re-installing just the Lagoon components (core, remote, build-deploy) with their latest charts or fresh built service images.

It is also possible to deselect which DBaaS providers are installed as part of a `local-stack` if you don't need them. This can be useful for mac users who currently have issues due to MongoDB not working on arm based systems, by default on mac or arm based systems the MongoDB provider will no longer be installed. The test suites may not work properly for any MongoDB based tests though, so be aware of this.

Harbor can also have problems installing on arm based systems. Support is now included to detect arm and will install an unauthenticated registry instead.

K8up can also be installed optionally, and the version that the remote-controller is configured to use can be selected. This can be useful for validating that backups work, or changes to backups work between v1 and v2 of k8up, or testing migrations of resources for v1 to v2.

Additionally, the k3d config is setup to allow for the host docker socket to be used, this allows local-stacks being torn down and started again to leverage the host docker so that non-built images don't need to be re-pulled within the cluster if they're already present on the host. This should speed up the creation of subsequent `local-stack` clusters.


tobybellwood (approved):
happy to merge this off the back of uselagoon/lagoon-charts#695

PR #3835, opened 2024-10-30
docs: move historical releases into archive

shreddedbacon:
# General Checklist

# Database Migrations

I get sick of the releases menu being so large. This moves old releases under an `Archived Releases` menu under releases to make the list a bit easier to look at.

### DIFF
--- a/docs/releases/.pages
+++ b/docs/releases/.pages
@@ -2,4 +2,6 @@ title: Releases
 order: desc
 nav:
   - Policy: policy.md
-  - ...
+  - ... | regex=2.(?:1[8-9]|[2-9]\d|[1-9]\d{2,}).*.md
+  - Archived Releases:
+    - ... | **/*.md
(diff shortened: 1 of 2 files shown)


tobybellwood:
Maybe we can move only the n-4 (or similar) older releases - our status pages and releases still point to a URL without the /archived part - and setting up redirects is a pain each time?

shreddedbacon:
> Maybe we can move only the n-4 (or similar) older releases - our status pages and releases still point to a URL without the /archived part - and setting up redirects is a pain each time?

Let me think about it a bit more. Still a draft.

shreddedbacon:
Updated to [use regex](https://regex101.com/r/2GDOnC/1) to match versions greater than a certain version to show under `Releases` anything else gets slapped under `Archived Releases` but retains the existing paths so no redirects required.

PR #3840, opened 2024-11-10
feat: view, add, and remove platform roles on users

### LINKED ISSUE #3635: Manage who has platform-owner role via API
To further prevent the need to log in to keycloak and provide more visibility of which users have the higher level `platform-owner` role, this should be able to be viewable via the API.

Being able to remove or add this role to users via the API is also desirable so that there is less need to log in to keycloak.

shreddedbacon:
# General Checklist

# Database Migrations

# Managing Platform Wide Users

This feature adds the functions required to view, and manage the roles of platform wide users.

No longer required to log in to Keycloak to manage the users in these roles. The `allPlatformUsers` query supports filtering by `email`, `id`, `role`, and if supported `gitlabId`, allowing for finding who is in a role, or if a user has any roles, quickly.

## New: Platform Organization Owner

Another platform wide role called `platform-organization-owner` is also included in this PR. This role would allow a user to create, delete, and update organizations, and list all deploytargets within Lagoon. This role is also able to assign and unassign deploytargets on an organization. 

Additionally, the ability to interact with organizations by way of what the standard `organization-owner` role would grant is part of this role. To interact with projects within an organization, similar to the standard `organization-owner` role, a user with this role will still need to be added to a group. 

This role is useful for granting a lower level of access to certain members of a team managing a platform to manage organizations, without needing to grant full platform owner role.

### CLI examples for verification

With a lagoon CLI configured for local API access (or if using local-stack in k3d) it is possible to run some commands to list, add, and remove a user. If this is approved, then the CLI will get updated to include these as commands.

```
# list all current platform owners
lagoon -l local-k3d raw --raw 'query allPlatformUsers{
    allPlatformUsers(role: OWNER){
        email
        platformRoles
    }
}'

# add owner role to guest user
lagoon -l local-k3d raw --raw 'mutation addPlatformRoleToUser{
    addPlatformRoleToUser(user:{email:"guest@example.com"}, role: OWNER){
        email
        platformRoles
    }
}'

# list all owners again and see the guest user
lagoon -l local-k3d raw --raw 'query allPlatformUsers{
    allPlatformUsers(role: OWNER){
        email
        platformRoles
    }
}'

# remove owner role from guest user
lagoon -l local-k3d raw --raw 'mutation removePlatformRoleFromUser{
    removePlatformRoleFromUser(user:{email:"guest@example.com"}, role: OWNER){
        email
        platformRoles
    }
}'
```

# Closing issues

closes #3635

### DIFF
--- a/services/api/src/models/user.ts
+++ b/services/api/src/models/user.ts
@@ -26,6 +26,7 @@ export interface User {
   owner?: boolean;
   admin?: boolean;
   organizationRole?: string;
+  platformRoles?: [string];
 }
 
 interface UserEdit {
@@ -42,6 +43,7 @@ interface UserEdit {
 
 export interface UserModel {
   loadAllUsers: () => Promise<User[]>;
+  loadAllPlatformUsers: () => Promise<User[]>;
   loadUserById: (id: string) => Promise<User>;
   loadUserByUsername: (email: string) => Promise<User>;
   loadUserByIdOrUsername: (userInput: UserEdit) => Promise<User>;
@@ -55,11 +57,20 @@ export interface UserModel {
     userGroups: Group[]
   ) => Promise<string[]>;
   addUser: (userInput: User, resetPassword?: Boolean) => Promise<User>;
+  addPlatformRoleToUser: (userInput: User, role: PlatformRole) => Promise<User>;
+  removePlatformRoleFromUser: (userInput: User, role: PlatformRole) => Promise<User>;
   updateUser: (userInput: UserEdit) => Promise<User>;
   deleteUser: (id: string) => Promise<void>;
   resetUserPassword: (id: string) => Promise<void>;
   userLastAccessed: (userInput: User) => Promise<Boolean>;
   transformKeycloakUsers: (keycloakUsers: UserRepresentation[]) => Promise<User[]>;
+}
+
+// these match the names of the roles created in keycloak
+enum PlatformRole {
+  VIEWER = 'platform-viewer',
+  OWNER = 'platform-owner',
+  ORGANIZATION_OWNER = 'platform-organization-owner'
 }
 
 interface AttributeFilterFn {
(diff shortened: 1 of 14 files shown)


services/api/src/typeDefs.js
```diff
@@ -464,6 +464,22 @@ const typeDefs = gql`
...
 
+  type PlatformUser {
+    id: String
+    email: String
+    firstName: String
+    lastName: String
+    comment: String
+    gitlabId: Int
+    platformRoles: [PlatformRole]
+  }
```
rocketeerbkw: I don't see the utility in having a second user type when the `platformRoles` field can be added to regular users (it should just return an empty array).

services/api/src/typeDefs.js
```diff
@@ -464,6 +464,22 @@ const typeDefs = gql`
...
+
+  enum PlatformRole {
+    VIEWER
+    OWNER
+    ORGANIZATIONOWNER
```
rocketeerbkw: Our other enums use `_` as a word delimiter, I think `ORGANIZATION_OWNER` is more readable.

services/api/src/models/user.ts
```diff
@@ -348,6 +352,102 @@ export const User = (clients: {
...
+        platformUsers.push(keycloakPlatformOrgOwners[f1])
+      } else {
+        platformUsers[found].platformRoles.push("organizationowner")
+      }
+    }
+    const users = await transformKeycloakUsers(platformUsers);
+    return users;
+  };
+
+  const platformRoleSwitch = (role) => {
```
rocketeerbkw: Optional:

This kind of conversion can be done at the graphql boundary. Then anywhere that graphql accepts or returns the enum, the actual string value used in the resolvers logic can be the same as what's in keycloak, and no further conversion would be needed. One small adjustment of the `PlatformRole` resolver:

```
PlatformRole: {
  VIEWER: 'platform-viewer',
  OWNER: 'platform-owner',
  ORGANIZATION_OWNER: 'platform-organization-owner',
},
```

rocketeerbkw (changes requested):

rocketeerbkw (approved):

PR #3841, opened 2024-11-17
refactor: use client to auth to keycloak instead of admin user

### LINKED ISSUE #3217: Use service account for Keycloak admin API access
Lagoon currently uses Keycloak's built-in `admin-cli` client and a username/password to authenticate against the Keycloak Admin API. One problem with this is that the token obtained has to be repeatedly refreshed since it is intended for the admin CLI tool and has a very short expiration.

Instead, could Lagoon use an administrative [service-account client](https://www.keycloak.org/docs/16.1/server_development/#authenticate-with-a-service-account)? This would avoid the repeated refreshing of admin tokens.

There would need to be some bootstrap logic to create this service-account client on initial install of Lagoon. The built-in `admin-cli` username/password can still be used for that. I guess this would go into the Keycloak [startup script](https://github.com/uselagoon/lagoon/blob/main/services/keycloak/startup-scripts/00-configure-lagoon.sh).

This might help with #2930.

shreddedbacon:
# General Checklist

# Database Migrations

This configures a new client in the `master` realm called `admin-api` which is now used by the Lagoon API when performing actions against Keycloak. This reduces the requirement for the username and password to be used.

The startup scripts are also changed to include attempting to auth with the new admin-api service account before falling back to using the username and password. This means that the very first startup requires the use of the username and password to seed the admin-api client, but subsequent startups and further upgrades will use the client to authenticate. This will allow for users to configure the admin user or other users in the master realm with two factor authentication if required.

Unfortunately, and this is the same with the current Keycloak admin password, rotating it via scripts is currently not possible. If the client secret for this new client needs to be rotated, it will have to be done via an admin user and also in helm at the same time.

TODO: Add support for configuring the secret in the helm chart

# Closing issues

closes #3217


rocketeerbkw (approved):
Another incremental step to reduce who needs to access keycloak admin account :+1:

shreddedbacon:
Can't merge this until there is the matching chart changes. I'll put that together and update once it is ready.

PR #3842, opened 2024-11-18
Update makefile and docs post 2.22 release

tobybellwood:


shreddedbacon:
Fine for now I think, we can always adjust the notes later on.

shreddedbacon (approved):

PR #3662, opened 2024-03-04
feat: use api-sidecar-handler to offload sshkey handling and add more types

shreddedbacon:
# General Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for inclusion in changelog

# Database Migrations

- [ ] If your PR contains a database migation, it **MUST** be the latest in date order alphabetically

The node package `sshpk` doesn't support all the various key types out there. The aim of this PR is to introduce a sidecar to required services that runs a simple http server that can replace the functionality of the `sshpk` package.

Each `api` and `webhooks2tasks` pod will get this sidecar so that they aren't reliant on a single service for this functionality, and as there is no state it doesn't need to be a single service and fits nicely as a sidecar.

It introduces new APIs to interact with user ssh keys that doesn't require the user to define the type independently. The input is just `publicKey` which can be the full `ssh-ed25519 A....z` format public key (including comment). The database table for the type is also converted from `enum` to `string` to support any new types, and since the older APIs had enum enforcement, that still applies for anyone that uses the older APIs.

The old APIs remain for now, but have been flagged as deprecated. They will be removed in a future release to give time for tooling and UI to be updated to use the newer APIs.

### Add public key
```
addUserSSHPublicKey(input: {
  name: "my-users-publickey-example"
  publicKey: "ssh-rsa AAAAB3NzaC1yc2EA<SNIP>LOiWw=="
  user: {
    email: "my-user@example.com"
  }
}) {
  id
}
```
### Update public key
```
updateUserSSHPublicKey(input: {
  id: 1
  patch: {
    publicKey: "ssh-rsa AAAAB3NzaC1yc2EA<SNIP>LOiWw=="
  }
}) {
  id
}
```
### Delete public key
```
deleteUserSSHPublicKey(input: {
    id: 1
})
```

# Closing issues

closes #2189 
closes #1584 
addresses #2384 by deprecating `deleteSshKey` which takes the `name` input

PR #3844, opened 2024-11-19
fix: api-sidecar-handler returns newlines in ssh key data

rocketeerbkw:
# General Checklist

# Database Migrations

A new API was added in https://github.com/uselagoon/lagoon/pull/3662 that handles public/private key pair cryptography. This API returns newlines in key values, which break parts of Lagoon that does key matching.

This PR removes the newlines from the API response and runs a database migration to clean up any existing newlines.

# Closing issues

n/a


shreddedbacon (approved):
Nice, easy fix

PR #3847, opened 2024-11-21
refactor: lowercase email address search to match keycloak

shreddedbacon:
# General Checklist

# Database Migrations

Creating a user in Lagoon with an uppercase email address ends up with Keycloak saving/storing it as lowercase. If a user then tries to query Lagoon for this user using the uppercase email, the Keycloak lookup fails because it does not match the lowercase saved email.

Keycloak saves all email addresses and usernames as lowercase when users are created, even if the provided email address contains uppercase. Confusingly though, it still respects case sensitivity in the API query params when doing lookups. It appears Keycloak have chosen to do this for some reason, probably for consistency, and there isn't any feature or parameter to change this behaviour.

There are numerous github issues and stack overflow posts where people have encountered issues with needing uppercase usernames or email address in Keycloak with no real clear indication if it can be changed. 

While the mailbox local part of an email address should be case sensitive per RFC 5321 2.3.11 and 2.4, however, the majority of email service providers will ignore this to reduce confusion. For example, `User@example.com` and `user@example.com` according to RFC should be treated as different mailboxes.

The simplest solution for us is to just make sure we lowercase email addresses when we're performing user lookups against Keycloak, which is what this pull request does.

Additionally, I've renamed some of the user lookup functions to better suit their use, and also updated the `allUser` query to be more efficient when filtering on an email address or user ID by only looking up the requested filter query.

### DIFF
--- a/services/api/src/models/user.ts
+++ b/services/api/src/models/user.ts
@@ -43,8 +43,8 @@ interface UserEdit {
 export interface UserModel {
   loadAllUsers: () => Promise<User[]>;
   loadUserById: (id: string) => Promise<User>;
-  loadUserByUsername: (email: string) => Promise<User>;
-  loadUserByIdOrUsername: (userInput: UserEdit) => Promise<User>;
+  loadUserByEmail: (email: string) => Promise<User>;
+  loadUserByIdOrEmail: (userInput: UserEdit) => Promise<User>;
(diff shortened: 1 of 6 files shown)


rocketeerbkw (approved):
Easy win to reduce user confusion :+1:

PR #3849, opened 2024-12-03
Update RabbitMQ to 3.13 on broker

tobybellwood:
Note that 3.13 includes an experimental feature flag for Khepri, which is not compatible with the management plugin, or upgradeable to Rabbit 4.x - so I've added logic to the feature flag enabler to ignore experimental FFs.


shreddedbacon (approved):
Ran this through a simulated upgrade using `make k3d/stable-local-stack` (with some minor changes to the seed data mutations for the platform roles, since they don't exist in 2.22.0) and then using `make k3d/install-lagoon` to run a helm upgrade with the images built in this branch and the latest charts version.

Everything appears to be functional.

PR #3859, opened 2024-12-23
chore: add view-users permission to lagoon-opensearch-sync

tobybellwood:
as per https://github.com/uselagoon/lagoon-opensearch-sync/issues/149 the lagoon-opensearch-sync service-account requires the view-users role in addition to the query-groups.

Follows same logic as https://github.com/uselagoon/lagoon/pull/3797


shreddedbacon (approved):
Yep, makes sense.

PR #3709, opened 2024-04-15
feat: introduce retention policy support

shreddedbacon:
# General Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for inclusion in changelog

# Database Migrations

- [x] If your PR contains a database migation, it **MUST** be the latest in date order alphabetically

Introducing retention policy support to Lagoon. This first iteration only contains Harbor and history based retention policy support.

Retention policy support allows platform owners to configure policies globally, organizationally, or directly on projects. Policy inheritance exists-ish, and a project will override organization, which will override global.

# Harbor Policies
Harbor policies allow for retaining the Last Pulled number of images based on rules that can be defined. This allows platform owners to be able to write rules that could retain all images of a project for the last # pulled images. Or configure multiple rules based on [doublestar pattern matching](https://github.com/bmatcuk/doublestar#patterns) that Harbor uses to prune some images more aggressively than others.

Harbor policies are enforced when they are linked, unlinked, updated, or when a deployment reaches the final stages. It uses hooks on certain events, and then send messages to the remotes to inform their configured Harbor of any policy changes that are required.

This requires that the `remote-controller` be updated with [support for receiving the retention policy payloads](https://github.com/uselagoon/remote-controller/pull/243)

# History Policies
History policies allow for retaining the last X number of builds and tasks and associated files/logs for these. It is possible to configure this for `COUNT`, `DAYS`, `MONTHS`. This allows for platform owners to retain a small number of builds and tasks within an environment if they choose to.

History policies are enforced whenever a new task or build is created.

See the [README](https://github.com/uselagoon/lagoon/blob/retention-policies/services/api/src/resources/retentionpolicy/README.md) for more information about retention policies.

PR #3866, opened 2025-01-06
feat: more filters for deploymentsByFilter

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Extends `deploymentsByFilter` with `month` and `includeDeleted` filtering.

* `month` can be used to filter builds for a specific month
* `includeDeleted` will allow to filter for environments that were deleted

Additionally, updates the resolver to exclude results for projects that are deleted, this is because environment deployment history is not currently deleted when a project is deleted, but this could change with #3709

### DIFF
--- a/services/api/src/resources/deployment/resolvers.ts
+++ b/services/api/src/resources/deployment/resolvers.ts
@@ -151,7 +151,7 @@ export const getDeploymentsByFilter: ResolverFn = async (
   { sqlClientPool, hasPermission, models, keycloakGrant, keycloakUsersGroups, adminScopes }
 ) => {
 
-  const { openshifts, deploymentStatus = ["NEW", "PENDING", "RUNNING", "QUEUED"] } = input;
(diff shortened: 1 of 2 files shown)


services/api/src/resources/deployment/resolvers.ts
```diff
@@ -181,13 +181,29 @@ export const getDeploymentsByFilter: ResolverFn = async (
       queryBuilder = queryBuilder.whereIn('environment.project', userProjectIds);
   }
 
+  // collect builds for a specific year/month
+  if (month) {
```
bomoko: perhaps this should be renamed, given we're using the year and month for the filter

bomoko (changes requested):

services/api/src/resources/deployment/resolvers.ts
shreddedbacon: I've updated it now to be `startDate` and `endDate` filters

bomoko (approved):

bomoko (approved):
oh, great change!

PR #3869, opened 2025-01-22
build: correct path in local-git for entrypoint

tobybellwood:
Alpine 3.21 changed the path for some utility packages previously under /sbin to /usr/sbin - and the one we use is in the local-git image entrypoint


shreddedbacon (approved):
Simple fix

PR #3860, opened 2024-12-27
chore: update keycloak to version 26

shreddedbacon:
# General Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for inclusion in changelog

# Database Migrations

- [ ] If your PR contains a database migation, it **MUST** be the latest in date order alphabetically

# Description

Update keycloak to v26.0.7

Supporting pullrequests:
* https://github.com/uselagoon/lagoon-ui/pull/307
  * changes the ui to use keycloak-js directly as the provided keycloak.js client in keycloak has been removed
* https://github.com/uselagoon/lagoon-charts/pull/733
* https://github.com/uselagoon/lagoon-ssh-portal/pull/497
  * with the enforcement of https on keycloak, ssh-portal fails to verify the issuer due to self signed certificate and crashes

PR #3875, opened 2025-02-03
test: simulate an upgrade in ci

shreddedbacon:
# General Checklist

# Database Migrations

# Description

The purpose of this is to install the latest stable release, and then perform an upgrade to the version in branch. This helps to validate if changes to Lagoon in either core, or the charts will cause a failure of an upgrade.

The process is to firstly install the latest stable release using the following (some extra options which may change with #3860)
```
make k3d/stable-local-stack
```
This will set up the k3d cluster, install the prerequisite components (ingress, harbor, etc) and then install the latest lagoon charts release versions (core, remote, build-deploy)

Once that is running, perform the upgrade using the following
```
make k3d/install-lagoon
```
This will just re-install, or upgrade, the core, remote, and build-deploy charts from the source repositories.

Also updated to allow for overriding the build-deploy-image without having to set the `buildImage` in the API. The `build-deploy-image` is now pulled and then pushed into the registry that is deployed as part of the local-stack and there are new make commands that can be used to push a different image into the registry to override it as required. This will hopefully allow for easier development and testing of build-deploy-images locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -818,10 +818,13 @@ k3d/stable-install-lagoon:
 
 # k3d/local-stack will deploy and seed a lagoon-core with a lagoon-remote and all basic services to get you going
 # and will provide some initial seed data for a user to jump right in and start using lagoon
+INSTALL_SEED_DATA = true
 .PHONY: k3d/local-stack
 k3d/local-stack: k3d/setup
 	$(MAKE) k3d/install-lagoon SKIP_DETAILS=true
+ifeq ($(INSTALL_SEED_DATA),true)
 	$(MAKE) k3d/seed-data
+endif
 	$(MAKE) k3d/get-lagoon-details
 
 # k3d/stable-local-stack is the same as k3d/local-stack except that it starts it with the latest stable chart versions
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -50,12 +50,13 @@ UPSTREAM_TAG ?= latest
...
 # edge is the most current merged change
+BUILD_DEPLOY_IMAGE_REPO = uselagoon/build-deploy-image
 BUILD_DEPLOY_IMAGE_TAG ?= edge
 
 # UI_IMAGE_REPO and UI_IMAGE_TAG are an easy way to override the UI image used
 # only works for installations where INSTALL_STABLE_CORE=false
-# UI_IMAGE_REPO = uselagoon/ui
-# UI_IMAGE_TAG =
+UI_IMAGE_REPO = uselagoon/ui
+UI_IMAGE_TAG = main
```
tobybellwood, line 59: acknowledging that this override needs to stay here until the next stable release is out of the door to ensure Keycloak library backwards compatibility

tobybellwood (approved):
This looks good to me - doesn't add much to the test time, but an important step in checking upgrades.

PR #3876, opened 2025-02-04
feat: built-in audit logging

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Introduce audit logging that saves events in the API instead of relying on the events being stored in an external logging tool (like opensearch) as they currently are.

They are searchable by date range and other fields (email, user id, event type, resource type, etc..)

Things to consider
* Truncating or ageing out events over time to prevent the table from becoming too large
  * Maybe periodically export to s3 for historical purposes? It could be possible to just dump events from the API, and then manually truncate the table to retain certain date range
  * Maybe #3709 could have some role here in the future to handle configuring the audit log retention policy?

### DIFF
--- a/services/api/database/migrations/20250218000000_audit_log.js
+++ b/services/api/database/migrations/20250218000000_audit_log.js
@@ -1,0 +1,38 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function(knex) {
+    organizations = await knex.schema.hasTable('audit_log');
+    if (!organizations) {
+        return knex.schema
+        .createTable('audit_log', function (table) {
+            table.increments('id').notNullable().primary();
+            table.specificType('usid', 'CHAR(36)'); // will be uuid from keycloak
+            table.string('email_address', 300);
+            table.specificType('resource_id', 'CHAR(36)'); // could be uuid (group/user) or int (project/env/org)
+            table.enu('resource_type', ['backup', 'bulkdeployment', 'deployment', 'deploytarget', 'deploytargetconfig', 'environment', 'group', 'notification', 'organization', 'project', 'sshkey', 'task', 'user', 'variable','workflow']);
+            table.string('resource_details', 300);
+            table.specificType('linked_resource_id', 'CHAR(36)'); // could be uuid (group/user) or int (project/env/org)
+            table.enu('linked_resource_type', ['backup', 'bulkdeployment', 'deployment', 'deploytarget', 'deploytargetconfig', 'environment', 'group', 'notification', 'organization', 'project', 'sshkey', 'task', 'user', 'variable','workflow']);
+            table.string('linked_resource_details', 300);
+            table.string('audit_event', 300);
+            table.specificType('impersonator_id', 'CHAR(36)'); // will be uuid from keycloak
+            table.string('impersonator_username', 300);
+            table.string('ip_address', 300);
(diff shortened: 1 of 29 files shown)


services/api/src/resources/audit/resolvers.ts
```diff
@@ -0,0 +1,78 @@
...
+export const getAuditLogs: ResolverFn = async (
+  root,
+  { input },
+  { sqlClientPool, hasPermission }
+) => {
+  await hasPermission('project', 'viewAll');
+
+  let queryBuilder = knex('audit_log');
+
+  if (input && input.userId) {
```
bomoko: you could potentially just wrap all these where statements in a single "if (input) {}" block, that way if there is no input object, you skip all the subsequent conditionals.

services/api/database/migrations/20250218000000_audit_log.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function(knex) {
+    organizations = await knex.schema.hasTable('audit_log');
+    if (!organizations) {
+        return knex.schema
+        .createTable('audit_log', function (table) {
+            table.increments('id').notNullable().primary();
+            table.specificType('usid', 'CHAR(36)'); // will be uuid from keycloak
```
bomoko, line 11: out of interest - why `usid` rather than `user_id`

bomoko (approved):
Looks good.

I've run this locally, although I haven't manually tested _all_ paths. I've done a bunch - logic makes sense and smoketesting is good.

Can also confirm impersonation logging is working as well.

Nice one.

services/api/database/migrations/20250218000000_audit_log.js
shreddedbacon: usid is what the other tables in the api-db use ([sshkey](https://github.com/uselagoon/lagoon/blob/audit-logging/services/api/database/migrations/20220908065119_initial_db.js#L287), [lastaccessed](https://github.com/uselagoon/lagoon/blob/audit-logging/services/api/database/migrations/20241229000000_user_last_accessed.js#L10))

services/api/src/resources/audit/resolvers.ts
shreddedbacon: Yeah, I can do that :D

shreddedbacon:
Just updated it with the limit changes

services/api/src/resources/audit/resolvers.ts
```diff
@@ -12,69 +12,89 @@ export const getAuditLogs: ResolverFn = async (
...
-  }
+    if (input.linkedResourceType) {
+      queryBuilder = queryBuilder.where('linked_resource_type', input.linkedResourceType);
+    }
 
-  if (input && input.source) {
-    queryBuilder = queryBuilder.where('source', input.source);
-  }
+    if (input.linkedResourceDetails) {
+      const l =
```
bomoko: Does this `l` do anything (I don't think I saw it the first time round)

bomoko (approved):
Looks good except for the rogue `l` if it is indeed rogue.

services/api/src/resources/audit/resolvers.ts
shreddedbacon: nothing to see here :hide-the-pain:

shreddedbacon:
> Looks good except for the rogue `l` if it is indeed rogue.

Yep, rogue :laughing:

PR #3877, opened 2025-02-06
docs: document organization env vars

### LINKED ISSUE #3732: Confusing information regarding .env and .lagoon.env based env files
The following section in the documentation references adding `.env` and `.lagoon.env` files into the git repo, but neglects to inform the user that they also need to be added to the `Dockerfile`, as this is roughly mentioned at the top of the page.

https://docs.lagoon.sh/concepts-advanced/environment-variables/#environment-files-existing-directly-in-the-git-repo

It might be worth including a note in the sub section that the files need to be included in the built images, or structured a bit differently to make it clearer in both areas that the files need to be added to the containers via the `Dockerfile` for each service that they need to be available in.

rocketeerbkw:
# General Checklist

# Description

Complete rewrite of docs for env vars, adding org env var information.

# Closing issues

closes #3732

### DIFF
--- a/docs/concepts-advanced/environment-variables.md
+++ b/docs/concepts-advanced/environment-variables.md
@@ -1,170 +1,352 @@
 # Environment Variables
 
-It is common to store API tokens or credentials for applications in environment variables.
-
-Following best practices, those credentials are different per environment. We allow each environment to use a separate set of environment variables defined in environment variables or environment files.
-
-As there can be environment variables defined in either the Dockerfile or during runtime (via API environment variables), we have a hierarchy of environment variables: environment variables defined in lower numbers are stronger.
-
-1. Environment variables (defined via Lagoon API) - environment specific.
-2. Environment variables (defined via Lagoon API) - project-wide.
-3. Environment variables defined in Dockerfile (`ENV` command).
-4. Environment variables defined in `.lagoon.env.$LAGOON_GIT_BRANCH` or `.lagoon.env.$LAGOON_GIT_SAFE_BRANCH` (if the file exists and where `$LAGOON_GIT_BRANCH` `$LAGOON_GIT_SAFE_BRANCH` are the name and safe name of the branch this Docker image has been built for), use this for overwriting variables for only specific branches.
-5. Environment variables defined in `.lagoon.env` (if it exists), use this for overwriting variables for all branches.
-6. Environment variables defined in `.env`.
-7. Environment variables defined in `.env.defaults`.
-
-`.lagoon.env.$LAGOON_GIT_BRANCH`, `.lagoon.env.$LAGOON_GIT_SAFE_BRANCH`, `.env`, and `.env.defaults` are all sourced by the individual containers themselves as part of running their entrypoint scripts. They are not read by Lagoon, but by the containers `ENTRYPOINT` scripts, which look for them in the containers working directory. If environment variables don't appear as expected, check if your container has a `WORKDIR` setting that points to somewhere else.
+Lagoon supports the use of environment variables, which are commonly used by applications to store configuration separate from code. There are multiple methods to set an env var and each Lagoon environment can have separate values.
 
 ## Environment Variables (Lagoon API)
 
-We suggest using the Lagoon API environment variable system for variables that you don't want to keep in your Git repository (like secrets or API keys), as they could be compromised by somebody having them on their local development environment or on the internet, etc.
+Use Lagoon API env vars for config that you don't want to keep in your Git repository (like secrets or API keys). This prevents sensitive information from getting leaked due to accidental exposure of the code.
 
-The Lagoon API allows you to define project-wide or environment-specific variables. Additionally, they can be defined for a scope-only build-time or runtime. They are all created via the Lagoon GraphQL API. Read more on how to use the GraphQL API [in our GraphQL API](../interacting/graphql.md) documentation.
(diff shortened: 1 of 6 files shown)


docs/concepts-advanced/environment-variables.md
```diff
@@ -1,170 +1,352 @@
...
-  addEnvVariable(
-    input:{
-      type:ENVIRONMENT,
-      typeId:546,
-      scope:RUNTIME,
-      name:"MYVARIABLENAME",
-      value:"MyVariableValue"
+Buildtime env vars are only available during Docker image builds, and can be accessed by using the `ARG` instruction. Check the [Docker documention on `ARG`](https://docs.docker.com/reference/dockerfile/#arg) for all the ways to use this method.
+
+Typically the `ARG` will go after the FROM. Read [the Docker documentation about ARG and FROM](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact).
```
bomoko, line 128: I wonder if it wouldn't be useful for us to have some kind of example illustrating the difference between build and runtime variables?

docs/concepts-advanced/environment-variables.md
```diff
@@ -1,170 +1,352 @@
...
 DB_USER=$DB_USERNAME # Redefine DB_USER with the value of DB_USERNAME e.g. if your application expects another variable name for the Lagoon-provided variables.
 ```
 
-### `.lagoon.env.$BRANCHNAME`
+### `.lagoon.env` and `.lagoon.env.BRANCHNAME`
+
+Use `.lagoon.env` when you want to always set env vars, but only for Lagoon images.
 
-If you want to define environment variables different per environment you can create a `.lagoon.env.$BRANCHNAME` e.g. for the main branch `.lagoon.env.main`. This helps you keep environment variables apart between environments.
+Use `.lagoon.env.BRANCHNAME` when you want to set different env vars per environment. `BRANCHNAME` will be replaced by the env vars `$LAGOON_GIT_BRANCH` or `$LAGOON_GIT_SAFE_BRANCH`. For example, if you deployed the branch `main`, the file name would be `.lagoon.env.main`. Branch names that contain special characters are sanitized, so check the value of `$LAGOON_GIT_SAFE_BRANCH` to determine what the filename should be.
```
bomoko, line 283: `"BRANCHNAME" will be replaced by the` 
I think that this can be stated slightly more strongly - `must be` rather than `will be` for instance

docs/concepts-advanced/environment-variables.md
```diff
@@ -1,170 +1,352 @@
...
+### Docker Compose
+
+Lagoon doesn't use env vars set using Docker Compose when deploying your projects. You may still choose to set them for local development purposes. Review the [Docker Compose env var documentation](https://docs.docker.com/compose/how-tos/environment-variables/) for more information.
 
-## Special Environment Variables
 
-### `PHP_ERROR_REPORTING`
+## Build Variables
+
+Use build vars when you want to set an env var that is valid at buildtime, but only for a single deployment. Bulk deployments also support build vars.
```
bomoko, line 302: With this, it might also make sense to add an example use case to the documentation.

bomoko: in fact, just explaining what we're doing in the example would be great

bomoko (approved):
Looks good!

I'm approving, with some comments - none of them should stop merging.

docs/concepts-advanced/environment-variables.md
```diff
@@ -1,170 +1,352 @@
...
+        name
+        scope
+        value
+      }
+    }
+    ```
+
+### Scopes
+
+Scopes are used to set an env var for a specific context. An env var can have a scope of `buildtime`, `runtime`, `global`, `container_registry` or `internal_container_registry`.
```
shreddedbacon:
```suggestion
Scopes are used to set an env var for a specific context. An env var can have a scope of `build`, `runtime`, `global`, `container_registry` or `internal_container_registry`.
```

docs/concepts-advanced/environment-variables.md
```diff
@@ -1,170 +1,352 @@
...
+
+#### Buildtime
```
shreddedbacon, line 124: Shouldn't this be `build` since that is what the scope is called?

shreddedbacon (changes requested):
The references to `buildtime` as the scope probably need to be corrected to the scope `build`. The flags for `--scope buildtime` will fail. And so will some of the mutations referencing this as `BUILDTIME`

docs/concepts-advanced/environment-variables.md
rocketeerbkw: "branchname must be replaced," sounds like a request to the reader, but this is documenting what happens in the container, so I'm not sure I would change anything, assuming I understand the feedback.

docs/concepts-advanced/environment-variables.md
rocketeerbkw: I'm open to suggestions. I picked the names of env vars that would make sense for the scopes used, but that assumes the reader would recognize them.

docs/concepts-advanced/environment-variables.md
rocketeerbkw: I wrote it that way originally, but found it cumbersome to read/write. I much prefer "buildtime" for just normal documentation language and it matches "runtime." I got a little carried away with the find/replace, will fix up the places where the api scope name is used.

shreddedbacon (approved):
The fixes for the build scope are good for me.

Only the two pending discussions from @bomoko which if they're all good, then this is fine for now? We could always adjust the examples and wording in a follow up if required?

PR #3880, opened 2025-02-10
feat: add keycloak plugin for better SSO UX

rocketeerbkw:
# General Checklist

# Database Migrations

# Description

Adds a [keycloak plugin](https://sventorben.github.io/keycloak-home-idp-discovery/) that can detect which identity provider to use based on the user email address, instead of forcing the user to select their own provider.

Configuration of the login flow is manual for this PR, there is an example at https://sventorben.github.io/keycloak-home-idp-discovery/configuration.html#authenticator-execution.

For local testing, you can run `make compose/example-sso` and then create a new login flow for the `lagoon` realm. Note that the lagoon theme is quite out of date and doesn't work well with the new flow.

### DIFF
--- a/services/keycloak/Dockerfile
+++ b/services/keycloak/Dockerfile
@@ -82,6 +82,8 @@ ENV TMPDIR=/tmp \
 
 VOLUME /opt/keycloak/data
 
+RUN curl -sSLo /opt/keycloak/providers/keycloak-home-idp-discovery.jar https://github.com/sventorben/keycloak-home-idp-discovery/releases/download/v26.0.1/keycloak-home-idp-discovery.jar
+
 COPY entrypoints/kc-startup.sh /lagoon/kc-startup.sh
 COPY entrypoints/wait-for-mariadb.sh /lagoon/entrypoints/98-wait-for-mariadb.sh
 COPY entrypoints/default-keycloak-entrypoint.sh /lagoon/entrypoints/99-default-keycloak-entrypoint.sh
--- a/local-dev/k3d-seed-data/seed-example-sso.sh
+++ b/local-dev/k3d-seed-data/seed-example-sso.sh
@@ -2,6 +2,7 @@
 
 CONFIG_PATH=/tmp/kcadm.config
 
+function auth {
 # login to keycloak
 if ! /opt/keycloak/bin/kcadm.sh config credentials --config $CONFIG_PATH --server http://localhost:8080/auth --realm master --client admin-api --secret ${KEYCLOAK_ADMIN_API_CLIENT_SECRET}
 then
@@ -12,6 +13,9 @@ then
    exit 1
    fi
 fi
+}
+
+auth
 
 if /opt/keycloak/bin/kcadm.sh get realms/sso --config $CONFIG_PATH > /dev/null; then
     echo "Realm sso is already created, skipping"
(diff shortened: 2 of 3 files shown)


services/keycloak/Dockerfile
```diff
@@ -88,6 +88,7 @@ COPY entrypoints/default-keycloak-entrypoint.sh /lagoon/entrypoints/99-default-k
 
 COPY startup-scripts /opt/keycloak/startup-scripts
 COPY themes/lagoon /opt/keycloak/themes/lagoon
+COPY providers /opt/keycloak/providers
```
tobybellwood: I maybe think we should be pulling a versioned copy of this from the source repo release eg
https://github.com/sventorben/keycloak-home-idp-discovery/releases/download/v26.1.1/keycloak-home-idp-discovery.jar (if that's the version added here)

rocketeerbkw: They break BC very frequently (v24.0.0, v26.0.0, v26.1.1, etc), and their version numbers don't follow exactly that of keycloak. We can pull it dynamically as long as we don't dependabot update it?

rocketeerbkw: To clarify, the version numbers aren't the same, but the plugin tracks the version of keycloak closely, and will break BC with old keycloak versions.

For example in 26.1.0:
> This and all upcoming releases will no longer be backwards compatible with Keycloak versions < 26.1

tobybellwood: yah - we do the same here https://github.com/uselagoon/lagoon/blob/main/services/broker/Dockerfile#L23

shreddedbacon: This would potentially problematic when we upgrade keycloak versions then? If we say, update keycloak to 27, will this provider still work, or would it need to be updated?

tobybellwood: yes, it will likely need to be updated - but having the versioned plugin documented should make it easier to verify compatibility (same as we do for broker - but broker hard breaks on an incompatible plugin)

rocketeerbkw: > This would potentially problematic when we upgrade keycloak versions then?

Yes, we're now blocked on upgrading keycloak until the plugin gets support for that version.

shreddedbacon: It might be worth including a comment in the dockerfile where this is defined
```
FROM quay.io/keycloak/keycloak:26.0.7
```
That says to ensure that a compatible version of this new provider is available before updating keycloak

bomoko: I'll just comment that besides this discussion, from my side everything looks good. At least, it seems to do what it says on the box when I spin it up locally.

Assuming that once the item above is addressed, we're good to go?

shreddedbacon:
I was curious to see this work so I put together a quick browser flow that uses it ([gist here](https://gist.github.com/shreddedbacon/b717cdd4646cb7ea86abf3886fbdf3c9))

I found that without turning on the `forwardUnverifiedEmail` and `forwardToLinkedIdp` options, that only the first login for an IDP user would succeed. Logging out then back in again without these options would result in the user being asked to login with their "shadow" user in the lagoon realm (because the shadow user, and the sso realm user has an unverified email address)

shreddedbacon (approved):
Happy for this to go in, having played with the functionality.

PR #3881, opened 2025-02-11
Chore/remove drutiny

### LINKED ISSUE #3082: Insights version 2 work - remove Lagoon Core Harbor/Trivy/Drutiny integration

bomoko:
This PR supersedes #3197 by completely removing Drutiny webhook processing

closes #3082


shreddedbacon:
Do we need to announce the removal of this? Or is it unused code being removed?

bomoko:
> Do we need to announce the removal of this? Or is it unused code being removed?

It's a good question - I'd like to ask @tobybellwood what he thinks here too - it might make sense to announce the deprecation and then remove it in the next version?

shreddedbacon (approved):
This seems fine. Just removal of unused functionality.

The question of whether this needs to be announced as a deprecation or just straight to removal can be done outside of this pullrequest if necessary though

tobybellwood:
Given zero effective usage and removal of other related functionality, this can just go without a warning.

fubarhouse:
giphy

PR #3883, opened 2025-02-14
docs: Add custom MkDocs theme override with tracking pixel

Schnitzel:
This adds a scarf.sh analytics pixel to understand how people are using our documentation


tobybellwood (approved):
Merging with seperate step to ensure it only runs in uselagoon repo

PR #3888, opened 2025-02-21
docs: storage type information

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Updates the docs with some information about the two current storage types that Lagoon offers, bulk and block.

I tried to keep it from being too descriptive of specific solutions implemented, mostly because it is very dependent on the underlying cloud providers and the provisioner in used for each storage class.

I included some generic `x_link` variables that can be used to populate with information or a link to further reading if required, this would be useful in the AIO docs to provide a link to a page that contains more specific cloud provider implementation information.


bomoko (approved):
Looks good

PR #3890, opened 2025-02-25
buid: update rabbitmq to v4

tobybellwood:


shreddedbacon:
Should also update the readme to remove the section about the HA queue policy since this is no longer in use https://github.com/uselagoon/lagoon/blob/main/docs/docker-images/rabbitmq.md#included-rabbitmq-default-schema-definitionsjson

Edit: These docs are for the `lagoon-images` image, not the lagoon-core broker. So changes to those docs will need to be done if/when the lagoon-image version changes. Ignoring for this update, as this is for core only.

shreddedbacon:
This doesn't require the [TLS changes in charts](https://github.com/uselagoon/lagoon-charts/pull/753) to work. 

The way that the charts works is by adding a configuration file that works in both the current version of rabbit used by Lagoon, and v4.0.x.

The additions to the script are to clean up an old priority that isn't used.

rocketeerbkw (approved):
Tested new version locally and didn't encounter any issues. Also verified delayed-message-exchange plugin still works.

PR #3894, opened 2025-03-04
fix: get environment details for auditlog

shreddedbacon:
# General Checklist

# Database Migrations

# Description

The `updateDeployment` mutation audit log incorrectly used an `environment` check which would be an `int` to check for its `id` and `name` which results in a graphql error. Also it removes what seems to be an unnecessary permission check, if you can deploy, you can view.


tobybellwood (approved):
nice pickup

PR #3895, opened 2025-03-04
fix: groups with greater than 100 members

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

In Keycloak 23, the API was updated to page results, with a default limit of `100` results for group member listings. When upgrading keycloak, this was not taken into account, so Lagoon groups with greater than 100 members are not able to list all members, and users outside the limit don't get the correct permissions.

# Closing issues

n/a


shreddedbacon (approved):
Easy fix, ran through this and works as intended.

PR #3897, opened 2025-03-07
tests: adds the ability to check url content for two options

tobybellwood:
With a recent change, the order of the routes returned in LAGOON_ROUTES changed, and this broke the test suite.

In this PR I have added both variants to the check-url-content routine, just in case we need to go back to the previous version.

Note: active-standby-kubernetes test failure is expected here temporarily


shreddedbacon:
Yah send it

PR #3900, opened 2025-03-12
Removes workflow references

bomoko:
We have decided to remove the Workflows system from Lagoon.

This PR will remove it from Core.

# General Checklist

# Database Migrations

# Description


shreddedbacon:
The [charts](https://github.com/uselagoon/lagoon-charts/pull/760) will need to be merged before this is, so that the `CHARTS_TREEISH` in this branch can be reverted to `main` before it is merged.

rocketeerbkw (changes requested):
There is one final bit of workflows left related to audit logs. There are type references in `node-packages/common/src/types.ts`, `services/api/src/typedefs.js`, and `services/api/src/resolvers.js`. The database schema for audit logs also includes `workflow` as an enum value. Potentially there would be audit logs if workflow mutations were called that could be purged during the db migration as well.

I don't see the need to keep these around since you don't need to review past audit logs for a feature that doesn't exist?

rocketeerbkw (approved):
Awesome, time for :wastebasket:

PR #3906, opened 2025-03-21
fix: removeUserFromGroup cache duplicates when user not in group

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

Running the `removeUserFromGroup` mutation, when the user isn't a member of the group, causes the group member cache to balloon in size. There was a bug that caused the entire member list to be appended instead of being reset each time it is called.

# Closing issues

n/a


shreddedbacon (approved):
This resolves the issue of ballooning group user cache.

PR #3913, opened 2025-04-24
docs: `mongo` service type is now called `mongodb`

rocketeerbkw:
# Description

The docs use `mongo` for the service type name, but the build-deploy-tool uses `mongodb`.

See also https://github.com/uselagoon/build-deploy-tool/pull/432


shreddedbacon (approved):
Easy one, once the other PR is merged this should be all good.

PR #3914, opened 2025-05-01
refactor: use redis for subscriptions

shreddedbacon:
# General Checklist

# Database Migrations

# Description

This removes the `graphql-rabbitmq-subscriptions` and replaces it with `graphql-redis-subscriptions`. The intention is to resolve long standing issues of queue exhaustion that the rabbitmq subscriptions experienced.

Local testing shows much better stability than the rabbitmq version.

### DIFF
--- a/services/api/src/clients/pubSub.ts
+++ b/services/api/src/clients/pubSub.ts
@@ -1,12 +1,13 @@
 import { path } from 'ramda';
 import { withFilter } from 'graphql-subscriptions';
-import { AmqpPubSub } from 'graphql-rabbitmq-subscriptions';
+import { RedisPubSub } from 'graphql-redis-subscriptions';
 import { ForbiddenError } from 'apollo-server-express';
 import { logger } from '../loggers/logger';
-import { getConfigFromEnv } from '../util/config';
 import { query } from '../util/db';
 import { Sql as environmentSql } from '../resources/environment/sql';
 import { ResolverFn } from '../resources';
+import { config } from './redisClient';
+import Redis from 'ioredis';
 
 export const EVENTS = {
   DEPLOYMENT: 'api.subscription.deployment',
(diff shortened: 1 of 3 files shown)


services/api/src/clients/pubSub.ts
```diff
@@ -1,32 +1,40 @@
...
-import { AmqpPubSub } from 'graphql-rabbitmq-subscriptions';
+import { RedisPubSub } from 'graphql-redis-subscriptions';
 import { ForbiddenError } from 'apollo-server-express';
 import { logger } from '../loggers/logger';
-import { getConfigFromEnv } from '../util/config';
 import { query } from '../util/db';
 import { Sql as environmentSql } from '../resources/environment/sql';
 import { ResolverFn } from '../resources';
+import { config } from './redisClient';
+import Redis from 'ioredis';
```
rocketeerbkw: `ioredis` is the older library, and while it's technically still maintained, it's suggested to use `node-redis` for all new work. The docs for the subscription library default to ioredis but it can also support node-redis, which is what we are using already in the [redisClient.ts](https://github.com/uselagoon/lagoon/blob/main/services/api/src/clients/redisClient.ts).

The `publisher`/`subscriber` need independent objects, so a small refactor of redisClient to export a `createRedisClient` method OR just copy/paste it.

services/api/src/clients/pubSub.ts
```diff
@@ -1,32 +1,40 @@
...
-  logger
+export const pubSub = new RedisPubSub({
+  publisher: new Redis(options),
+  subscriber: new Redis(options),
+  connectionListener: (err) => {
+    if (err) {
+        logger.info("Pubsub: error connecting to redis", err);
+    } else {
+      logger.info("Pubsub: Successfuly connected to redis");
+    }
```
rocketeerbkw: This doesn't appear to log anything at all?

rocketeerbkw (changes requested):
Tested locally and subscriptions are working. Would be more future proof to change the redis library, which I don't expect to change anything functionally.

services/api/src/clients/pubSub.ts
shreddedbacon: Couldn't find any clear examples of how to not use ioredis. But did find an example that just passed the connection string into the redispubsub and everything still works.

services/api/src/clients/pubSub.ts
shreddedbacon: Was from an example I found when I first ran it, appears irrelevant so just removed it and everything still works.

services/api/src/clients/pubSub.ts
shreddedbacon: Feel free to fix it to how you think it should be otherwise

rocketeerbkw (approved):
Still works fine locally.

PR #3915, opened 2025-05-02
refactor: deprecate openshiftprojectpattern

### LINKED ISSUE #3905: Deprecate project/namespace pattern fields
The `openshiftProjectPattern`/`kubernetesNamespacePattern` fields are not useful to end users and don't seem to add any benefits for Lagoon administrators.

We can remove any use of patterns in the code base and mark the fields as deprecated and remove the logic for setting/getting.

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Deprecate the `openshiftProjectPattern` (and `kubernetesNamespacePattern`) field in the API as this is no longer used.

# Closes

closes #3905


node-packages/commons/src/tasks.ts
```diff
@@ -531,7 +536,7 @@ export const getControllerBuildData = async function(deployTarget: any, deployDa
   // if no build image is determined, the `remote-controller` defined default image will be used
   // once it reaches the remote cluster.
 
-  // @TODO: openshiftProject here can't be generated on the cluster side (it should be) but the addOrUpdate mutation doesn't allow for openshiftProject to be optional
+  // @TODO: envNamespaceName here can't be generated on the cluster side (it should be) but the addOrUpdate mutation doesn't allow for envNamespaceName to be optional
```
rocketeerbkw: `envNamespaceName` is not a variable, is this comment out of date? I'm not sure that the `@TODO` would be in this case, since the previous comment also doesn't apply anymore?

rocketeerbkw (changes requested):
There's one final reference that I could find https://github.com/uselagoon/lagoon/blob/main/node-packages/commons/src/api.ts#L98.

The `DeployTargetConfig` type also has a `deployTargetProjectPattern` which can get the same treatment. It wasn't listed in the original issue, so that could be a follow up instead.

node-packages/commons/src/tasks.ts
shreddedbacon: Yah, that shouldn't have been there

rocketeerbkw (approved):
Thanks! Tested locally and there are no errors when trying to CRUD any of the deprecated fields.

PR #3918, opened 2025-05-19
chore: add old queue cleanup to broker

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Cleans up old queues from the broker. Mostly this is to follow up with the removal of `workflows` but has included all the previously removed service queues.


bomoko (approved):
Looks good as is, and it makes sense why it's here - but the file is ill named, it would be good if we could change it.

PR #3919, opened 2025-05-22
docs: document valkey and redis-8 Lagoon images

rocketeerbkw:
# Description

* Updates docs for the Redis 8 image and how it differs from redis <= 7.
* Adds a page for the Valkey image.

**Note:** Redis 8 images haven't been released yet, so this could wait a bit before merging.


tobybellwood (approved):
THanks for starting, I added more bits and coverage after the 25.5.0 release

PR #3920, opened 2025-05-28
Feature - role emails

bomoko:
# General Checklist

# Database Migrations

# Description

This PR will add the ability to email users when their organization roles change.


services/api/src/resources/user/resolvers.ts
```diff
@@ -485,10 +487,23 @@ export const addAdminToOrganization: ResolverFn = async (
       details: `${user.email} role ${role}`,
     },
   };
+
+  var addAdminToOrganizationEmail = true;
+  try {
+    let dbUserDetails = await models.UserModel.getFullUserDetails(user);
+    console.log("dbUserDetails", dbUserDetails);
```
bomoko: kill this

services/api/src/resources/user/resolvers.ts
```diff
@@ -485,10 +487,23 @@ export const addAdminToOrganization: ResolverFn = async (
...
+    if( dbUserDetails && dbUserDetails['orgEmailOptin'] == false) {
+      addAdminToOrganizationEmail = false;
+    }
+  } catch (e) {
+    console.log("Error while trying to get user details", e);
```
bomoko: change this to logger

services/api/src/resources/user/resolvers.ts
```diff
@@ -485,10 +487,23 @@ export const addAdminToOrganization: ResolverFn = async (
       details: `${user.email} role ${role}`,
     },
   };
+
+  var addAdminToOrganizationEmail = true;
```
bomoko: switch to false by default

services/logs2notifications/internal/handler/useraction_events_template.go
```diff
@@ -0,0 +1,387 @@
...
+
+      <table border="0" cellpadding="0" cellspacing="0" role="presentation" style width="100%">
+        <tbody>
+
+              <tr>
+                <td align="left" vertical-align="top" style="font-size:0px;padding:0px;word-break:break-word;">
+
+      <div style="-webkit-hyphens: none; -moz-hyphens: none; hyphens: none; word-break: keep-all; font-family: Montserrat, Arial, sans-serif; font-size: 16px; font-weight: 400; line-height: 21px; text-align: left; color: inherit;"><p class="body_text" style="-webkit-hyphens: none; -moz-hyphens: none; hyphens: none; word-break: keep-all;">
+
+      <!-- This is the main content of the email, which will be rendered in HTML format. -->
```
shreddedbacon, line 272: As we discussed, it might be good to separate the main content generation out so that the template is re-usable across the standard email notifications as well as user actions

services/logs2notifications/internal/handler/useraction_events_template.go
```diff
@@ -0,0 +1,387 @@
...
...IkmSJElSaxZi6k2SBcCBwA6ts8yQH9PNCftm6yCSJEmSJI0LCzH1LsmmdPPFNm6dZYpdBbwROLKqbmsdRpIkSZKkcWIhpiaSzAN2BN4OzG8cZ5rcCrwXeEtV/b51GEmSJEmSxpGFmJpKsiawN7AbsHLjOJPuOGC3qjqndRBJkiRJksaZhZjGQpJ16W5AfH7rLBPoXLoi7H9bB5EkSZIkaRJYiGmsJNkCOBTYsHWWCXAN8GbgvVV1c+swkiRJkiRNCgsxjZ0kKwI7AW8F1mocZxwFOBLYp6quaB1GkiRJkqRJYyGmsZVkLWA/YGdghbZpxsa3gNdU1Rmtg0iSJEmSNKksxDT2kmwIvAvYsnWWhi4C9gSOqaq0DiNJkiRJ0iSzENNESFLAVsAhwLqN4/TpeuBA4OCquqF1GEmSJEmSJPUsySpJ9kxyTabfx5L8VeuvuSRJkiRJksZAkvlJjkyysG1nNRLfT/Kk1l9jSZIkSZIkjaEkGyU5uXGBNSyXJfmXJPNaf10lSZIkSZI0xpJUkhcmuahpnbXsbkpyYJI1W38tJUmSJEmSNEGSrJZk3yTXN6235ubzSdZp/bWTJEmSJEnSBEuyIMknGhdd9+SsJFu0/lpJkiRJkiRpiiTZNMkPGhdfi7sqyauSrNj66yNJkiRJkqQplGRekpekG1jf0q1J3p1krdZfE0mSJEmSJM2AJGumG1x/U4My7GtJNmz9NZAkSZIkSdIMSrJuki/0VISdn+R5Sar131uSJEmSJEkzLskW6Qbbj8K1SV6XZJXWf09JkiRJkiTpT5KsmGTndIPuh2Fhkg8meUDrv5skSZIkSZJ0l5KslW7g/a3LUYadkuTxrf8ukiRJkiRJ0lJLsmGSr86xCLs4yT/GOWGSJEmSJEmaREkqyXOTnHcPRdgNSfZLsnrrzJIkSZIkSdJyS7JKkj2TXLOEMuzoJH/dOqMkSZIkSZI0dEnmJzky3cD805M8uXUmSZIkSZIkaeSSrJNkhdY5JEmSJEmSJEmSJEmSJEmSJEmSJEl35f8DjIblfEYaRkoAAAAASUVORK5CYII=" style="border:0;display:block;outline:none;text-decoration:none;height:auto;width:100%;font-size:16px;" width="290" height="auto">
```
shreddedbacon: May want to also make this image configurable

services/logs2notifications/internal/handler/useraction_events_template.go
shreddedbacon, line 320: As we discussed, it might be good to separate the main content generation out so that the template is re-usable across the standard email notifications as well as user actions

services/api/database/migrations/20250520225815_add_emailoptout_to_user.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function(knex) {
+    return knex.schema.alterTable('user', function(table) {
+    table.boolean('org_email_optin').notNullable().defaultTo(true);
```
shreddedbacon: Might be worth changing this up a bit now to make further options in the future, thinking things like

```
table.boolean('opt_email_org_role').notNullable().defaultTo(true); // when a users role within the organization management changes
table.boolean('opt_email_sshkey').notNullable().defaultTo(true); // when an ssh key is added or removed from the user
table.boolean('opt_email_group_role').notNullable().defaultTo(false); // when the users role within a group is changed. this should probably default to false because it could be quite noisy for some users
```

services/api/src/models/user.ts
```diff
@@ -29,6 +29,7 @@ export interface User {
   admin?: boolean;
   organizationRole?: string;
   platformRoles?: [string];
+  emailOptIn?: boolean;
```
shreddedbacon: I'm wondering if this should be an interface, to allow for future expansion
```
interface IEmailNotifications {
  organizationRoleChanges?: boolean;
  sshKeyChanges?: boolean;
  groupRoleChanges?: boolean;
}

export interface User {
  ...
  emailNotifications?: IEmailNotifications;
}
```

services/logs2notifications/internal/handler/main.go
```diff
@@ -284,6 +286,18 @@ func (h *Messaging) processMessage(message []byte) {
 				}
 			}
 		}
+
+		// Here we deal explicitly with a class of 'user_action' events
+		if notification.Meta.Level == "user_action" {
+			if notification.Meta.Event == "api:addAdminToOrganization" {
```
shreddedbacon: This currently only fires on if a user is added as a manager to an organization. What about removal (https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/user/resolvers.ts#L409)

Additionally, it could be worth notifying on the following:
Would we want to notify users if their platform roles changed? Maybe? This would probably be non-optional though so we don't have to expose the option to adjust it to general users
* add platform role to user: https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/user/resolvers.ts#L633
* remove platform role from user: https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/user/resolvers.ts#L654

Would probably be good to notify when ssh keys are added or removed from an account, github does this, others probably do too
* add sshkey: https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/sshKey/resolvers.ts#L97
* remove sshkey: https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/sshKey/resolvers.ts#L272

Possible future notify if a group role was changed? These actual `addUserToGroup` and `removeUserFromGroup` resolvers may need new logic in them to check if the group/role are unchanged though, otherwise if you run the same command twice with the same inputs the action is still performed and the user would receive an email saying their role was changed to the same role.
* added user to group: https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/group/resolvers.ts#L645
* removed user from group: https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/group/resolvers.ts#L706

services/api/src/resources/user/resolvers.ts
shreddedbacon: rather than have the payload flag being `addAdminToOrganizationEmail` would be more generic to have it be named `sendUserEmail` and then if the other optional email notifications are added in the future, the logic would be to determine which type of notification it is in the resolver and set this flag.

then the `logs2notifications` service useraction event handler would only need to check if `sendUserEmail=true`

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -0,0 +1,121 @@
...
+
+type addAdminToOrganization struct {
+	Meta struct {
+		Payload struct {
+			AddAdminToOrganizationEmail bool                           `json:"addAdminToOrganizationEmail"`
+			User                        addAdmintoOrganizationUser     `json:"user"`
+			Resource                    addAdminToOrganizationResource `json:"resource"`
+		} `json:"payload"`
+	} `json:"meta"`
+}
```
shreddedbacon: When `auditLogs` were added, we set up a [new data structure](https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/user/resolvers.ts#L339-L350) in the activity payload that contained information specific to this.

We should probably do something similar for the email notifications in the resolvers that would send emails with a new structure that contains this payload data. rather than relying on information that may or may not be present in all other resolvers.

In a previous comment I mentioned replacing `addAdminToOrganizationEmail` with `sendUserEmail`, but it could even be that if the new payload is in the message then send the email instead. No need for a specific flag then, as the injected payload could be "the flag"

shreddedbacon (changes requested):

services/api/database/migrations/20250520225815_add_emailoptout_to_user.js
shreddedbacon: Also note with the `opt_email_group_role` possible notifications, another reason for default `false` would be for some of AIO customers that may use other systems for managing users that could already send these notifications. having it as default opt out would be best. and then users would be able to opt in specifically for this.

services/logs2notifications/internal/handler/useraction_events_template.go
bomoko: Done

services/logs2notifications/internal/handler/useraction_events_template.go
bomoko: done

shreddedbacon:
Hitting a few issues testing this out. Will need to make some changes to the [charts](https://github.com/uselagoon/lagoon-charts/pull/792) to fix logs2notifications email service for running in the k3d/local-stack.

# SSH keys
When I add a key to my user, it sends the email successfully. Removing the key though it uses `deleteSshKeyById` and I see no email. The logs2notifications pod reports these logs
```
2025/06/11 00:49:03 Sending email for 'api:addSshKey' event, user:platformowner@example.com
2025/06/11 00:49:36 Skipping email for 'api:deleteSshKeyById' event - payload: {{ 0 } <nil>}
```
Looking at the resolvers, `deleteSshKey` appears to be configured to add the email data to the [userActivityLogger](https://github.com/uselagoon/lagoon/blob/feature/roleemails/services/api/src/resources/sshKey/resolvers.ts#L283-L299) except it looks to be missing in `deleteSshKeyById` [here](https://github.com/uselagoon/lagoon/blob/feature/roleemails/services/api/src/resources/sshKey/resolvers.ts#L342-L356).

# Organization role change

I'm not really sure where to look at this one
```
$ lagoon -l local-k3d add organization-administrator --email guest@example.com --role owner --organization-name lagoon-demo-organization
Error: graphql: Cannot read properties of undefined (reading 'organizationRoleChanges')
```

# Group role change
Tweaking the user to opt in to group role emails, i tried to add then remove a user from a group. No email sent, logs from logs2notifications
```
2025/06/11 01:13:52 Skipping email for 'api:addUserToGroup' event - payload: {{ 0 } <nil>}
2025/06/11 01:14:21 Skipping email for 'api:removeUserFromGroup' event - payload: {{ 0 } <nil>}
```
I'm guessing similar story where the email data is not added to the user activity logger correctly if the user has opt in to those emails

bomoko:
yeah @shreddedbacon - it was actually a combo deal - mostly the issue was the fact that I changed the structure of User to include the submodel - this fixes that (https://github.com/uselagoon/lagoon/pull/3920/commits/0a70ec93e3510a11fedd75085ff1b86b4a2c60fb)

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -0,0 +1,288 @@
...
+	// let's check if we need to send an email
+	if payload.Meta.Payload.UserActionEmailDetails == nil {
+		if h.EnableDebug {
+			log.Printf("Skipping email for '%v' event - payload: %v\n", notification.Meta.Event, string(rawPayload))
+		}
+		return nil
+	} else {
+		if h.EnableDebug {
+			log.Printf("Sending email for '%v' event, user:%s\n", notification.Meta.Event, payload.Meta.Payload.UserActionEmailDetails.Email)
+		}
```
shreddedbacon: With the enable debug being set to false now. I'd change this up to log when it sends, but leave the `skipping` wrapped in debug

shreddedbacon: It might be worth looking at slog or something that has logging levels at some point later on

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -0,0 +1,288 @@
...
+	content := `
+<p>Hello,</p>
+
+<p>
+  You (<strong>{{.Name}}</strong>) have {{.Action}} an SSH key {{if .Keyname}} ({{.Keyname}}) {{end}} in your Lagoon account.
+</p>
+
+<p>
+  If you have any questions or need assistance, please contact your organization manager.
+</p>
```
shreddedbacon, line 135: Probably don't need this block when adding or removing an ssh key, the user would have done this themselves (or a platform user)

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -0,0 +1,288 @@
...
+	content := `
+<p>Hello,</p>
+
+<p>
+  You (<strong>{{.Name}}</strong>) have been assigned the platform role of <strong>{{.Role}}</strong>.
+</p>
+
+<p>
+  If you have any questions or need assistance, please contact your organization manager.
+</p>
```
shreddedbacon: Same with this one, probably not required. This wouldn't be an organization manager doing this action.

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -0,0 +1,288 @@
...
+	return errors.New(fmt.Sprintf("Unable to match incoming notification: %v", notification))
+}
+
+func (h *Messaging) groupAddRemoveEmailMessage(valuesStruct UseractionEmailDetails, event string) error {
+
+	content := `
+<p>Hello,</p>
+
+<p>
+  You (<strong>{{.Name}}</strong>) have been {{.Action}} the group "{{.Groupname}}" in your Lagoon account.
```
shreddedbacon: Styling to match others
```suggestion
  You (<strong>{{.Name}}</strong>) have been {{.Action}} the group <strong>{{.Groupname}}</strong>.
```

shreddedbacon: I wonder if its also possible to include which role the user was granted here when the action is adding

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -0,0 +1,288 @@
...
+	return nil
+}
+
+func (h *Messaging) addEditRemoveSshKeyEmailMessage(valuesStruct UseractionEmailDetails, event, action string) error {
+
+	content := `
+<p>Hello,</p>
+
+<p>
+  You (<strong>{{.Name}}</strong>) have {{.Action}} an SSH key {{if .Keyname}} ({{.Keyname}}) {{end}} in your Lagoon account.
```
shreddedbacon: Styling to match other messages

```suggestion
  You (<strong>{{.Name}}</strong>) have {{.Action}} an SSH key {{if .Keyname}} <strong>{{.Keyname}}</strong> {{end}}
```

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -0,0 +1,288 @@
...
+	content := `
+<p>Hello,</p>
+
+<p>
+  You (<strong>{{.Name}}</strong>) have been removed from the platform role of <strong>{{.Role}}</strong>.
+</p>
+
+<p>
+  If you have any questions or need assistance, please contact your organization manager.
+</p>
```
shreddedbacon: Same, probably not required.

services/api/src/typeDefs.js
```diff
@@ -1944,21 +1951,30 @@ const typeDefs = gql`
...
     email: String!
     firstName: String
     lastName: String
     comment: String
     gitlabId: Int
     resetPassword: Boolean
+    emailNotifications: UserEmailOptionInput
   }
 
+
```
shreddedbacon: Additional line not needed

services/api/src/models/user.ts
```diff
@@ -9,11 +9,19 @@ import { Sql } from '../resources/user/sql';
 import { getConfigFromEnv } from '../util/config';
 import { Helpers as groupHelpers } from '../resources/group/helpers';
 import { logger } from '../loggers/logger';
+import { sendToLagoonLogs } from '@lagoon/commons/dist/logs/lagoon-logger';
```
shreddedbacon: This appears to be unused?

shreddedbacon (changes requested):
Few comments

For updating the user locally until the UI is sorted, this example can be used with the CLI
```
lagoon -l local-k3d raw --raw 'mutation updateUser{
  updateUser(input:{
    user: {email:"platformowner@example.com"}
    patch: {
      emailNotifications: {
        groupRoleChanges: true
      }
    }
  })
  {
    id
    emailNotifications{
      organizationRoleChanges
      sshKeyChanges
      groupRoleChanges
    }
  }
}'
```

services/logs2notifications/internal/handler/useraction_events.go
shreddedbacon: I wonder if it makes sense to also change the greeting to this instead (same for all others too)
```suggestion
<p>Hello <strong>{{.Name}}</strong>,</p>

<p>
  You have been {{.Action}} the group "{{.Groupname}}" in your Lagoon account.
```

shreddedbacon (approved):
This all works as intended. I think it would be good to make sure that the [UI component](https://github.com/uselagoon/lagoon-ui/pull/324) associated to this is released at the same time though, mostly so that users have a way to opt out if they wish without having to use the API directly

PR #3921, opened 2025-06-02
refactor: Update Apollo Server to v3

CGoodwin90:
Updates `apollo-server-express `from `v2` - `v3,` this is a decent shift from an all-in-one package to a more modular approach. File uploads and subscriptions are no longer built in and required manual setup via `graphql-upload` & `graphql-ws`. Most parts of the apollo server implementation have been refactored to accomodate the changes in the new package.

Requires UI update https://github.com/uselagoon/lagoon-ui/pull/325


shreddedbacon:
Just FYI, you can change this line https://github.com/uselagoon/lagoon/blob/gql-update/Makefile#L59 in the Makefile to point to a different UI image so that when you run `make k3d/local-stack` it will start up using the image defined, so changing it to `pr-325` for your UI changes.

services/api/src/resources/task/helpers.ts
```diff
@@ -99,7 +99,7 @@ export const Helpers = (sqlClientPool: Pool, hasPermission, adminScopes) => {
       let rows = await query(sqlClientPool, Sql.selectTask(insertId));
       const taskData = R.prop(0, rows);
 
-      pubSub.publish(EVENTS.TASK, taskData);
+      pubSub.publish(EVENTS.TASK, {taskChanged: taskData});
```
rocketeerbkw: The data payload here is different from all other `pubSub.publish` calls and it appears that adding tasks doesn't trigger a subscription.

```suggestion
      pubSub.publish(EVENTS.TASK, taskData);
```

rocketeerbkw (changes requested):
API still working correctly overall. Task file uploads and subscriptions also working as expected. No regressions in authz checks.

There is just one change needed for task subscriptions, otherwise LGTM.

shreddedbacon (approved):
Approving based on @rocketeerbkw last comment now that changes applied

Issue #3029, opened 2022-02-15
Official keycloak Node.js client library deprecated


rocketeerbkw:
https://www.keycloak.org/2022/02/adapter-deprecation.html

We use [keycloak-nodejs-connect](https://github.com/keycloak/keycloak-nodejs-connect) in the [api to authenticate](https://github.com/uselagoon/lagoon/blob/main/services/api/src/util/auth.ts#L46-L48) keycloak generated jwt tokens. We'll need to use something else in the long term.

Timeline from announcement:
- February 2022: Adapters deprecated
- September 2022: No more major/minor releases of adapters
- December 2022: No more micro releases of adapters

tobybellwood:
To bundle with some of the keycloak updates in #3492 or similar

PR #3923, opened 2025-06-09
Upgrade keycloak to 26.2.5

rocketeerbkw:
# General Checklist

# Description

- Upgrades keycloak to the most recent version 26.2.5
- Upgrades keycloak related libraries in the api to most recent versions
- Validates api tokens against keycloak session database
- Removes some dead code

Keycloak has still not recommended a replacement for keycloak-connect libarary and it appears to still get updated, so didn't take care of this issue yet https://github.com/uselagoon/lagoon/issues/3029

### DIFF
--- a/services/api/src/clients/keycloakClient.ts
+++ b/services/api/src/clients/keycloakClient.ts
@@ -44,10 +44,23 @@ keycloakGrantManager.validateToken = function validateToken(
     } else if (token.content.typ !== expectedType) {
       reject(new Error('invalid token (wrong type)'));
     } else if (token.content.iat < this.notBefore) {
-      reject(new Error('invalid token (future dated)'));
+      reject(new Error('invalid token (stale token)'));
     } else if (!token.content.iss.includes('auth/realms/lagoon')) {
       reject(new Error('invalid token (wrong ISS)'));
     } else {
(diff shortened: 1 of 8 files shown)


services/api/src/clients/keycloakClient.ts
```diff
@@ -44,10 +44,23 @@ keycloakGrantManager.validateToken = function validateToken(
     } else if (token.content.typ !== expectedType) {
       reject(new Error('invalid token (wrong type)'));
     } else if (token.content.iat < this.notBefore) {
-      reject(new Error('invalid token (future dated)'));
+      reject(new Error('invalid token (stale token)'));
     } else if (!token.content.iss.includes('auth/realms/lagoon')) {
       reject(new Error('invalid token (wrong ISS)'));
     } else {
+      const audienceData = Array.isArray(token.content.aud) ? token.content.aud : [token.content.aud]
```
bomoko, line 51: noting that this is pulled from https://github.com/keycloak/keycloak-nodejs-connect/blob/main/middleware/auth-utils/grant-manager.js

services/api/src/lib/keycloak-grant-manager.js
bomoko, line 1: Simply updating the source - I wonder if it would be good to note where this comes from (if I we don't already)?

bomoko (approved):
code looks good

PR #3924, opened 2025-06-11
Change: Adds Organization on Projects

CGoodwin90:
# General Checklist

Adds a new resolver, `getOrganizationByProject`, to fetch basic organization metadata associated with a project. 

Required for UI PR: https://github.com/uselagoon/lagoon-ui/pull/326


services/api/src/typeDefs.js
```diff
@@ -1074,6 +1078,12 @@ const typeDefs = gql`
     envVariables: [EnvKeyValue]
   }
 
+  type OrganizationMetaData {
```
shreddedbacon: I know that the ID is already retrievable under the `organization` field, but it should also be in this too. Just so if we in the future decide to deprecate or remove the old field, the ID is still retrievable

services/api/src/resources/organization/resolvers.ts
```diff
@@ -358,6 +358,20 @@ export const getProjectsByOrganizationId: ResolverFn = async (
...
+  args,
+  { sqlClientPool }
+) => {
+
+  if (!project || !project.organization) {
+    return null
+  }
+
+  const organisationData = await query(sqlClientPool, Sql.selectOrganizationMetaData(project.organization));
+  return organisationData[0];
```
shreddedbacon:
```suggestion
  const organizationData = await query(sqlClientPool, Sql.selectOrganizationMetaData(project.organization));
  return organizationData[0];
```

s -> z

services/api/src/resolvers.js
```diff
@@ -423,6 +424,7 @@ const resolvers = {
     groups: getGroupsByProjectId,
     privateKey: getPrivateKey,
     publicKey: getProjectDeployKey,
+    organizationMetaData: getMinimalOrganizationByProject,
```
shreddedbacon: Seeing it as `organizationMetaData` now just feels off to me. It is a bit of a bummer that `organization` is used :hide-the-pain:

I wonder if `organizationData` is maybe better, since it isn't really "metadata" we're retrieving. Thoughts?

shreddedbacon (changes requested):

services/api/src/resolvers.js
tobybellwood: I also think (and it was probably me that suggested it :see_no_evil: ) that `organizationMetaData` isn't the right term - if we run with a broader concept of attaching metadata to projects/environments/services - then it should extend to orgs too - so this would be confusing. I think `organizationData` is probably closer (or `organizationDetails`)

shreddedbacon: `organizationDetails` sounds good

bomoko: yeah, it's more like an organization summary, or stub. It's not actually metadata at all.

CGoodwin90: Yep I like `organizationDetails` better. Or even `organizationSummary`?

services/api/src/typeDefs.js
```diff
@@ -606,6 +606,10 @@ const typeDefs = gql`
     """
     organization: Int
     """
+    MetaData of organization
+    """
+    organizationMetaData: OrganizationMetaData
```
shreddedbacon: Blaize and I had a chat, we think this should change to return the full `Organization` type, effectively removing the need for this new type.

Then in the `getMinimalOrganizationByProject` resolver, add a check that will check if any other fields than the following are requested, then a `hasPermission` check against the organization id would be performed.

Here is an example of how to check if a field was requested https://github.com/uselagoon/lagoon/blob/v2.25.0/services/api/src/resources/env-variables/resolvers.ts#L100

So effectively you would do something like this in the resolver
```
if fields notin id, name, friendlyName, description {
  hasPermission
end
```

services/api/src/resources/organization/resolvers.ts
```diff
@@ -358,6 +358,33 @@ export const getProjectsByOrganizationId: ResolverFn = async (
...
+
+  const extraFields = requestsExtraFields(info, ['id', 'name', 'friendlyName', 'description']);
+
+  let orgQuery = Sql.selectOrganizationDetails(project.organization);
+
+  if (extraFields) {
+    await hasPermission('organization', 'view', {
+      organization: project.organization,
+    });
+    orgQuery = Sql.selectOrganization(project.organization);
```
shreddedbacon: This shouldn't be required. If a user without permission tries to request the extra fields, they'll get a permission denied and no data will be returned from the previous select on line 374

shreddedbacon: We might want to rethink description as well, so for now I think remove it so its only id and name/friendlyname.

We can re-evaluate description at a later period.

services/api/src/resources/organization/resolvers.ts
```diff
@@ -358,6 +358,30 @@ export const getProjectsByOrganizationId: ResolverFn = async (
   return rows;
 };
 
+export const getOrganizationByProject: ResolverFn = async (
```
shreddedbacon, line 363: I'd just add a comment to this that this resolver should only ever be called by another resolver that has done a permission check on the project first.

shreddedbacon (approved):

PR #3925, opened 2025-06-16
fix: task should fail if the upload of files to the Lagoon API fails

seanhamlin:
# General Checklist

# Description

At the moment, if the file upload to the Lagoon API fails (e.g. HTTP 502) then the next command `rm` succeeds, and thus the task gets a `success` status applied, even though it failed.

By moving to `&&` this will prevent the `rm` from running, and `exit 1` will be seen, and thus the task should appear as failed.

We also need to ensure that `curl` responds with a non-zero exit code, hence adding `--fail-with-body`, this flag was added in 2021 - see https://daniel.haxx.se/blog/2021/02/11/curl-fail-with-body/


shreddedbacon:
Just rebasing to get other fixes for the tests to run

shreddedbacon (approved):
All good for me. Noting that if the task did fail at the upload stage and the `rm -rf $file` was not executed, when the task pod is removed, the temporary filesystem it used would also be removed. This effectively means the `rm` operation is redundant.

PR #3934, opened 2025-06-25
Revert "Update: AWS SDK v3 update"

bomoko:
Reverts uselagoon/lagoon#3927

While the AWS SDK v3 upgrade _mostly_ worked, this seems to have introduced an error with s3 download links, which, when clicked, are responding as such

image


bomoko:
In order to release, let's actually revert the work - and then follow up with an upgrade on the very next release.

CGoodwin90 (approved):

Issue #3939, opened 2025-06-30
Caching issue when you edit your name, not being reflected in the group membership UI


seanhamlin:
# Describe the bug

When you update your name using Lagoon CLI or Keycloak, this is not immediately reflected in the UI when looking at lists of users (e.g. group membership within an organization).

## To Reproduce

1. Update your name `lagoon update user -E sean.hamlin@amazee.io -C sean.hamlin@amazee.io --first-name 'Sean 2'`
2. Visit a group that you are a member of
3. Notice the name is not changed

## Expected behavior

Cache is evicted when no longer accurate.

## Workaround

Edit the group membership for a user, make no changes, just click save. Seems to do the required purging.

jackwrfuller:
I believe this is an issue with the redis cache of group membership not being invalidated when updating user information via the API or Keycloak.

In particular, when getting group membership information, I believe the cache `cache:keycloak:group-members:${group.id}` is polled. However, this is not deleted when the `updateUser` mutation is called via the API. My PR #3941 addresses this.

This appears to solve the issue when user information is updated via the API (e.g lagoon-cli). However, it does not fix the issue when user information is edited directly in keycloak. This is because (I believe) the API isn't aware of most events happening inside keycloak.

I think there are two potential solutions to this. The first is to create an event-listener within keycloak that listens for changes to user info, and tells the API about the event when it happens. I'm not excited by this option, since it would require a fair amount of work and there are probably security concerns.

Alternatively, let us think: what is the use-case of updating user info from within keycloak? The only one I can see is the `Your Account` page in the UI, which doesnt exist but instead redirects the user to keycloak. It seems to me that it would be easier to just create a `Your Account` in the UI which calls the `updateUser` mutation, thus avoiding working within keycloak entirely. 

I might have a go at this, unless there are objections.

jackwrfuller:
I've had a go at a simple 'Your Account' page, see the PR linked above. AFAIK there should be no reason for a user to need to edit their details within Keycloak now?

PR #3941, opened 2025-07-02
fix: invalidate group membership cache on user update

jackwrfuller:
# General Checklist

# Description

Addresses #3939.

When user information is updating via `updateUser` mutation, the redis caches that contain the user's information are not invalidated, and therefore contain out of date information.

This PR ensures that whenever user information is updated via this mutation, the correct caches are invalidated (i.e deleted).


shreddedbacon (approved):
This works as expected when using the `updateUser` mutation.

As [you've found in the UI](https://github.com/uselagoon/lagoon-ui/pull/331) the `your account` page leads to keycloak, and if a user were to update their details there, this would not be captured by this resolver. That is an issue we need to figure out in more detail, but for now this change in isolation is a good starting point.

PR #3944, opened 2025-07-07
fix: add typename to field check to resolve permission error

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Fixes an issue with the extra field check to include `__typename` in the included fields to fix permission error for non-organization owners

This can be verified in a `k3d/local-stack` using the `orguser` account to view the project overview page. With this patch in place the page loads, without there is a permission denied error.

### DIFF
--- a/services/api/src/resources/organization/resolvers.ts
+++ b/services/api/src/resources/organization/resolvers.ts
@@ -371,8 +371,7 @@ export const getOrganizationByProject: ResolverFn = async (
     return null
   }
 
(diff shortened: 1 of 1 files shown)


bomoko:
Hi @shreddedbacon
Thanks!

Okay - so when I actually click the link as the `orguser` user taking me through to the organization, I get an error, presumably the user doesn't have permissions to view the organization.

bomoko (approved):
This does fix the problem, but it introduces a new problem in the UI we need to address (described in comments)

As it goes, though, this does what it says on the box.

PR #3945, opened 2025-07-08
refactor: Better logs2notification templates

bomoko:
## New Functionality

This PR takes the work done in https://github.com/uselagoon/lagoon/pull/3920 and extends it, allowing the user to supply an email template via a cli argument and env var

Passing an email template is simply a case of starting logs2Notifications with either

* `--email-template-file=/file/on/disk.gotmpl`
* env var `EMAIL_TEMPLATE_FILE` set to the absolute path of the go-template

Futhermore, the service now requires the existence of two default template files
* `templates/defaultlogo.b64`  - a base64 encoded logo for the emails
* `templates/mail.gotmpl` - which houses the actual template for the mail system 

## Additionally

We introduce a few changes
* Better handling of the base template and base64 encoded image (loading from file)
* Email notifications now use the same mechanism as the user action notifications


services/logs2notifications/internal/handler/templates/files/defaultTemplateImageBase64.txt
shreddedbacon, line 1: Minor thing, could we rename this something simpler like `default_logo.b64`?

services/logs2notifications/internal/handler/templates/userinteraction_event_mail.gotmpl
shreddedbacon, line 1: For readability, could you run this through a formatter like https://webformatter.com/html?

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -261,7 +267,8 @@ func (h *Messaging) addAdminToOrganization(valuesStruct UseractionEmailDetails)
   If you have any questions or need assistance, please contact your organization manager.
 </p>
 `
-	mainHTML, err := templateGenerator(content, valuesStruct, h.EmailBase64Logo)
+	templateGenerator := NewTemplateDataGenerator(content, h.EmailTemplate, h.EmailBase64Logo)
+	mainHTML, err := templateGenerator.Generate(valuesStruct) //templateGenerator(content, valuesStruct, h.EmailBase64Logo)
```
shreddedbacon:
```suggestion
	mainHTML, err := templateGenerator.Generate(valuesStruct)
```

services/logs2notifications/internal/handler/useraction_events.go
```diff
@@ -205,7 +208,9 @@ func (h *Messaging) addPlatformRoleToUser(valuesStruct UseractionEmailDetails) e
 </p>
 `
 
-	mainHTML, err := templateGenerator(content, valuesStruct, h.EmailBase64Logo)
+	//mainHTML, err := templateGenerator(content, valuesStruct, h.EmailBase64Logo)
```
shreddedbacon: Guessing not required now?

shreddedbacon (changes requested):
All good for me. Few things.

But the main one as discussed was having the file default change so that it can be overridden easily via configmap only mounting, rather than having to also set the env var.

shreddedbacon (approved):
Nice, way better looking than the previous templates.

Will need follow up work to add configmap support to the charts for overriding the files. Would be nice to have that prior to the release, but not required as the logo variable can at least be used if the logo is needing to be changed.

PR #3952, opened 2025-07-24
chore: support seeding of user and org in local-stack

jackwrfuller:
# General Checklist

# Description

Enables local-stack support for the initial seeding feature that will be introduced in https://github.com/uselagoon/lagoon-charts/pull/802

A user and organisation can be created automatically in the following manner:

```bash
make k3d/local-stack LAGOON_SEED_USERNAME=user@example.com LAGOON_SEED_PASSWORD=password LAGOON_SEED_ORGANIZATION=example-org
```


shreddedbacon:
Might be worth updating the documentation here https://github.com/uselagoon/lagoon/blob/main/docs/installing-lagoon/lagoon-core.md#install-the-helm-chart to include a note or something about how someone could seed an initial user and organization

And rework https://github.com/uselagoon/lagoon/blob/main/docs/installing-lagoon/lagoon-core.md#log-in-to-the-ui to mention the seed user instead of `lagoonadmin`

jackwrfuller:
> Might be worth updating the documentation here https://github.com/uselagoon/lagoon/blob/main/docs/installing-lagoon/lagoon-core.md#install-the-helm-chart to include a note or something about how someone could seed an initial user and organization
> 
> And rework https://github.com/uselagoon/lagoon/blob/main/docs/installing-lagoon/lagoon-core.md#log-in-to-the-ui to mention the seed user instead of `lagoonadmin`

Done. I also slightly reworked the lagoon core helm installation instructions to hopefully be a tad clearer.

Rereading these docs reminded me that I think it would be helpful if there was a 'starterkit' `values.yml` that contained the bare minimum needed to get up and runnning. I imagine such a thing would be heavily inspired by the local-stack.

shreddedbacon:
> Done. I also slightly reworked the lagoon core helm installation instructions to hopefully be a tad clearer.

Thanks!

> Rereading these docs reminded me that I think it would be helpful if there was a 'starterkit' `values.yml` that contained the bare minimum needed to get up and runnning. I imagine such a thing would be heavily inspired by the local-stack.

We've talked about this in the past, and it is probably easier to do now than when we first discussed it. It would certainly be better than linking to the linter-values file, which can have all sorts of weird things in them to suit being used in CI. Maybe in a separate PR, but we should probably chat about it first.

shreddedbacon (approved):

PR #3957, opened 2025-07-25
tests: pin ansible version to 2.18.6

shreddedbacon:
Ansible 2.19.0 has a lot of breaking changes that requires considerable refactoring of our test suite to work - https://github.com/ansible/ansible/blob/v2.19.0/changelogs/CHANGELOG-v2.19.rst

This pins it to 2.18.6 which was the last working version for the test suite.

We should investigate whether we continue to use ansible and do the necessary refactoring, or re-write them with go


tobybellwood (approved):
Yeah, sensible.

Issue #3961, opened 2025-07-31
FR : Support Bitbucket Server clone links in webhook payloads


tarikflz:
### Problem

Currently, Lagoon’s Bitbucket webhook handler expects payload fields like:

- `repository.links.html.href`
- `repository.full_name`

However, these fields are not present in Bitbucket Server (formerly known as Stash) webhook payloads. Instead, Bitbucket Server provides SSH and HTTP(S) clone URLs under:

```json
"repository": {
  "links": {
    "clone": [
      { "href": "ssh://git@bitbucket.mycompany.com:7999/project/repo.git", "name": "ssh" },
      { "href": "https://bitbucket.mycompany.com/scm/project/repo.git", "name": "http" }
    ]
  }
}
```

This causes Lagoon’s webhook handler to throw an error:

```bash
Cannot read properties of undefined (reading 'href')
```

### Proposed Solution
Enhance the Bitbucket webhook handler to:
 - Check for `repository.links.clone`
 - Prefer the `ssh`clone link if available
 - Fallback to current behavior (`html.href + full_name`) for backward compability with Bitbucket Cloud
### Reference
📄 Bitbucket Server webhook documentation:
https://confluence.atlassian.com/bitbucketserver064/event-payload-974388877.html#Eventpayload-repositoryevents

### Benefit
This change would allow teams using Bitbucket Server or Data Center editions to integrate with Lagoon natively via webhooks, without custom patches or middleware.

PR #3962, opened 2025-07-31
feat: Add Bitbucket Server webhook compatibility

tarikflz:
# General Checklist

# Database Migrations

- [x] This PR does not contain a database migration

# Description

This PR improves Lagoon's Bitbucket webhook handler to support Bitbucket Server and Data Center instances, which do not include the `repository.links.html.href` or `repository.full_name` fields in webhook payloads.

Instead, Bitbucket Server provides clone URLs under `repository.links.clone`, including SSH and HTTP(S) links. This update:

- Extracts the SSH URL from `clone` if available and uses it directly.
- Falls back to the current behavior (`html.href + full_name`) for Bitbucket Cloud compatibility.
- Ensures `giturl` generation is resilient and safe across both Bitbucket variants.

### Reference

Official documentation for Bitbucket Server webhook payload format:  
📄 https://confluence.atlassian.com/bitbucketserver064/event-payload-974388877.html#Eventpayload-repositoryevents

Tested with:

- Bitbucket Server 7.x payloads
- Bitbucket Cloud payloads
- Lagoon webhook triggering via local DDEV setup

# Closing issues

This implements: https://github.com/uselagoon/lagoon/issues/3961

### DIFF
--- a/services/logs2notifications/internal/handler/email_events.go
+++ b/services/logs2notifications/internal/handler/email_events.go
@@ -186,16 +186,23 @@ var emailEvent = map[string]EventMap{
 	"github:pull_request:opened:handled":           {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestOpened"},
 	"gitlab:merge_request:opened:handled":          {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestOpened"},
 	"bitbucket:pullrequest:created:opened:handled": {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestOpened"}, //not in slack
+	"bitbucket:pr:opened:handled":                  {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestOpened"}, //not in slack
 	"bitbucket:pullrequest:created:handled":        {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestOpened"}, //not in teams
 
 	"github:pull_request:synchronize:handled":      {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"},
 	"gitlab:merge_request:updated:handled":         {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"},
 	"bitbucket:pullrequest:updated:opened:handled": {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in slack
 	"bitbucket:pullrequest:updated:handled":        {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in teams
+	"bitbucket:pr:modified:opened:handled":         {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in slack
+  "bitbucket:pr:modified:handled":                {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in teams
+  "bitbucket:pr:from_ref_updated:opened:handled": {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in slack
+  "bitbucket:pr:from_ref_updated:handled":        {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in teams
 
 	"github:pull_request:closed:handled":      {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestClosed"},
 	"bitbucket:pullrequest:fulfilled:handled": {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestClosed"},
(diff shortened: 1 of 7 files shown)


services/logs2notifications/internal/handler/email_events.go
```diff
@@ -186,26 +186,34 @@ var emailEvent = map[string]EventMap{
...
 	"bitbucket:pullrequest:created:handled":        {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestOpened"}, //not in teams
 
 	"github:pull_request:synchronize:handled":      {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"},
 	"gitlab:merge_request:updated:handled":         {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"},
 	"bitbucket:pullrequest:updated:opened:handled": {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in slack
 	"bitbucket:pullrequest:updated:handled":        {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in teams
+	"bitbucket:pr:modified:opened:handled":         {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in slack
+  "bitbucket:pr:modified:handled":                {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in teams
+  "bitbucket:pr:from_ref_updated:opened:handled": {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in slack
+  "bitbucket:pr:from_ref_updated:handled":        {Emoji: infoEmoji, Color: "#E8E8E8", Template: "mergeRequestUpdated"}, //not in teams
```
shreddedbacon: The indentation of these will need to be corrected in all of the events.go files you've modified. You can fix them easily by running
```
cd services/logs2notifications && go fmt ./...
```

services/webhook-handler/src/extractWebhookData.ts
```diff
@@ -58,26 +58,61 @@ export function extractWebhookData(req: IncomingMessage, body: string): WebhookR
...
+          throw new Error('Missing repository.full_name in webhook payload.');
+        }
+
+        if (!regexmatch[2]) {
+          giturl = `git@${domain}:${bodyObj.repository.full_name}.git`;
+        } else {
+          const port = regexmatch[2];
+          giturl = `ssh://git@${domain}${port}/${bodyObj.repository.full_name}.git`;
+        }
+      }
```
shreddedbacon, line 113: We have tests for bitbucket cloud based payloads (though these are probably a little bit out of date in some of the data)
* https://github.com/uselagoon/lagoon/blob/main/tests/tests/bitbucket.yaml
* https://github.com/uselagoon/lagoon/tree/main/tests/tests/bitbucket
* https://github.com/uselagoon/lagoon/tree/main/tests/tasks/webhook-bitbucket < this is where the payloads would need to be saved

Are you able to create tests the same way but call them `stash`. This would mean copying those bitbucket tests, renaming them to `stash` and then modifying the references like `ci-bitbucket` to be `ci-stash` and then updating the payloads with matching payload data from stash. You will need to modify the payload data to account for the local git that would be created for the tests to use, you can read the existing payload files for bitbucket to see how this is done.

The tests are a little bit difficult to understand, but the main thing would be to ensure that you've provided some example webhook payloads that would be used to verify that the payloads from Stash/BB Server will correctly provision environments.

You'd be able to run the tests locally too, but resetting after a failed test run is a bit cumbersome, its usually easier to just tear it all down and start over like this
```
make k3d/clean && make k3d/test TESTS=[stash]
```

Having these tests would ensure that both the `bitbucket` cloud tests still pass, and that the `stash` webhooks would work too.

services/webhooks2tasks/src/webhooks/projects.ts
```diff
@@ -291,6 +291,7 @@ export async function processProjects(
         break;
 
       case 'bitbucket:repo:push':
+      case 'bitbucket:repo:refs_changed':
```
shreddedbacon, line 294: These may need to be split off from being prefixed with `bitbucket` to being prefixed with `stash`. Since the two systems are different, this would make it clear they are different. You'd need to update all the events in logs2notifications to deal with this, and then move this https://github.com/tarikflz/lagoon/blob/feat/webhook-bitbucket-server/services/webhook-handler/src/extractWebhookData.ts#L61 into this if block https://github.com/tarikflz/lagoon/blob/feat/webhook-bitbucket-server/services/webhook-handler/src/extractWebhookData.ts#L85 and add `webhooktype = 'stash';` into the clone links block https://github.com/tarikflz/lagoon/blob/feat/webhook-bitbucket-server/services/webhook-handler/src/extractWebhookData.ts#L68

This way its made clear that an event would be stash/bitbucket server, instead of bitbucket cloud when the messages are flying around.

shreddedbacon (changes requested):

services/logs2notifications/internal/handler/email_events.go
tarikflz: applied

shreddedbacon:
May be replaced with https://github.com/uselagoon/lagoon/pull/3979

PR #3966, opened 2025-08-06
Change title of docker compose to be without a dot

eiriksm:
# General Checklist

# Database Migrations

# Description

I assume the title of this section should in fact be `docker-compose.yml` and not `.docker-compose.yml`


rocketeerbkw (changes requested):
Fun fact: the name of the compose file can be anything, since it is defined in [`.lagoon.yml`](https://docs.lagoon.sh/concepts-basics/lagoon-yml/#docker-compose-yaml).

I agree that it makes sense to use the standard naming convention though. Would you please also update it on the [japanese language page](https://github.com/uselagoon/lagoon/blob/main/docs/ja/concepts-basics/lagoon-yml.md?plain=1#L568)?

PR #3971, opened 2025-08-24
feat: show if user has 2fa or is federated user in api

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Currently the only way to check if a user has 2fa enabled is to log in to keycloak as admin to check users individually. This is not ideal, as it can be time consuming and requires users to log in to keycloak directly, which we want to limit.

Another is detecting if a user is using an identity provider instead of direct user access.

This feature will now show if a user has 1 or more 2fa features enabled by setting a `has2faEnabled` flag on their account in the API. It currently can detect if a user is using totp or webauthn, but the flag is just a `true|false`

The federated user check will also only detect if the user has any federated identities, and if so will flag the user as a federated user. This appears as `isFederatedUser` which is also a `true|false`

The [UI could also be extended](https://github.com/uselagoon/lagoon-ui/pull/340) to display this status so organization owners can see which of their users have 2fa enabled, or which are federated users (SSO).

Example in organization users page: 
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ BUILD_DEPLOY_IMAGE_TAG ?= edge
 # UI_IMAGE_REPO and UI_IMAGE_TAG are an easy way to override the UI image used
 # only works for installations where INSTALL_STABLE_CORE=false
 UI_IMAGE_REPO = uselagoon/ui
-UI_IMAGE_TAG = pr-340
+UI_IMAGE_TAG = main
 
 # The two variables below are an easy way to override the insights-handler image used in the local stack lagoon-core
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -56,7 +56,7 @@ BUILD_DEPLOY_IMAGE_TAG ?= edge
 # UI_IMAGE_REPO and UI_IMAGE_TAG are an easy way to override the UI image used
 # only works for installations where INSTALL_STABLE_CORE=false
 UI_IMAGE_REPO = uselagoon/ui
-UI_IMAGE_TAG = main
+UI_IMAGE_TAG = pr-340
```
shreddedbacon: This will need to be reverted before merging

rocketeerbkw (approved):
Discussed and approved.

PR #3972, opened 2025-08-26
docs: update local-stack backups

jackwrfuller:
# General Checklist

# Description

Fixes some code snippet display issues in the k8up section of the local-stack docs, as well as some other minor updates.

### DIFF
--- a/docs/contributing-to-lagoon/developing-lagoon.md
+++ b/docs/contributing-to-lagoon/developing-lagoon.md
@@ -230,14 +230,14 @@ Currently, Harbor does not have any arm64 based images. This means local-stack w
 
 If you want to skip installing Harbor entirely, even if your system is not arm based, you can use the following
 
-```bash title="Harbor disabled
+```bash title="Harbor disabled"
 INSTALL_UNAUTHENTICATED_REGISTRY=true
 ```
 
 !!! info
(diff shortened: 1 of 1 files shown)


docs/installing-lagoon/lagoon-backups.md
shreddedbacon, line 1: This describes how to install k8up with the new version, but doesn't include the flags needed on the `lagoon-remote` chart to tell it to use the newer version.  The older version of k8up is still the default version that will be used unless those flags are defined. We may may the newer version the default in the future, but there is no standard migration available for previously deployed resources at this time.

Additionally, [lagoon-core doesn't support](https://github.com/uselagoon/lagoon-charts/pull/796) the newer k8up version for the automated backups, so while newer k8up can be used, the older version is still the current supported version.

shreddedbacon: This can lead to a possible scenario where a user uninstalls the older version for the newer version, but not triggering any deployments so that the newer versions resources are created. Effectively killing all backups for an environment.

This is because we have no automated migration published that happens outside of a build.

shreddedbacon: I think for now, removing this part of the documentation changes and leaving it for local-stack only is the best move while we work on figuring out a proper upgrade path to v2

jackwrfuller: understood and updated.

shreddedbacon (approved):

PR #3973, opened 2025-08-26
Change: Deprecate `internal_container_registry`

### LINKED ISSUE #3878: Deprecate `internal_container_registry` env var scope
With the removal of harbor from Lagoon core, env vars with scope `internal_container_registry` are no longer used. We should deprecated it, possibly remove it.

At a minimum, we can remove all usage in resolvers and delete data in the api-db. Maybe we have to keep the enum value but we can return an error if we want to be super BC careful (not sure there's a valid reason though).

CGoodwin90:
# General Checklist

# Database Migrations

# Description

Adds a migration to remove `internal_container_registry` scoped variables & no-ops relevant resolver interactions.

closes https://github.com/uselagoon/lagoon/issues/3878

### DIFF
--- a/services/api/database/migrations/20250822070204_delete_internal_container_registry_scoped_env_vars.js
+++ b/services/api/database/migrations/20250822070204_delete_internal_container_registry_scoped_env_vars.js
@@ -3,9 +3,10 @@
  * @returns { Promise<void> }
  */
 exports.up = function(knex) {
-  return knex('env_vars')
-    .where('scope', 'internal_container_registry')
-    .del()
+  return knex.schema
+      .raw(`DELETE FROM env_vars WHERE scope = 'internal_container_registry'`)
+      .raw(`ALTER TABLE env_vars
+        MODIFY scope ENUM('global', 'build', 'runtime', 'container_registry') NOT NULL DEFAULT 'global';`);
 };
 
 /**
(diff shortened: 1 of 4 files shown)


services/api/database/migrations/20250822070204_delete_internal_container_registry_scoped_env_vars.js
rocketeerbkw, line 1: Since the resolvers won't accept this scope anymore, we can also remove the scope as a valid db enum. Example of doing that here https://github.com/uselagoon/lagoon/blob/main/services/api/database/migrations/20250312183338_delete_workflows.js#L12.

Also need to add it back in the `down` function.

services/api/src/resources/env-variables/resolvers.ts
```diff
@@ -311,6 +312,11 @@ export const addOrUpdateEnvVariableByName: ResolverFn = async (
     throw new Error('A variable name must be provided.');
   }
 
+  if (scope === 'internal_container_registry') {
+    logger.warn('Variable scope "internal_container_registry" is deprecated & can no longer be set.');
+    return;
```
rocketeerbkw: `logger.warn` only sends a message to stdout/container logs. If we throw an error instead, the API caller will see it too.

```suggestion
    throw new Error('Variable scope "internal_container_registry" is deprecated & can no longer be set.');
```

Same change needed in `addEnvVariable` too.

rocketeerbkw (changes requested):
There are also references that can be completely removed in the docs https://github.com/uselagoon/lagoon/blob/main/docs/concepts-advanced/environment-variables.md and internal api client https://github.com/uselagoon/lagoon/blob/main/node-packages/commons/src/api.ts#L145

rocketeerbkw (approved):
Working locally :+1:

PR #3978, opened 2025-09-08
fix: increase env limit for default org in local-stack

jackwrfuller:
# General Checklist

# Description

The local-stack is currently preconfigured with a default organisation with an environment limit of 4, but unfortunately it comes preexhausted, as `lagoon-demo-org` already has 4 environments. 

This causes friction when a user is using the local-stack to test a project, since they then need to increase this limit or delete environments. Having had to do this myself many times, I find it very annoying and believe the limit should just be increased out of the box.

### DIFF
--- a/local-dev/k3d-seed-data/00-populate-kubernetes.gql
+++ b/local-dev/k3d-seed-data/00-populate-kubernetes.gql
@@ -206,7 +206,7 @@ mutation PopulateApi {
     friendlyName: "Lagoon Demo Organization"
     description: "An organization for testing"
     quotaProject: 5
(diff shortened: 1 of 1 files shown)


shreddedbacon:
The organization created in the local-stack has an env quota set for a reason, to verify that the organization quota enforcement works.

If you're working locally, you can unlock it with the CLI.
```
lagoon -l local-k3d update organization --organization-name lagoon-demo-organization \
	--project-quota -1 \
	--environment-quota -1
```

Alternatively, you can modify the seed script to create a second organization, just don't set the quotas on that one.

PR #3982, opened 2025-09-12
fix: local-dev/jwt version check

shreddedbacon:
# General Checklist

# Database Migrations

# Description

This fixes a small version check error in the `local-dev/jwt` target to prevent downloading it every time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -617,11 +617,11 @@ endif
 
 .PHONY: local-dev/jwt
 local-dev/jwt:
-ifeq ($(JWT_VERSION), $(shell jwt -v 2>/dev/null | sed -nE 's/jwt ([0-9.]+).*/v\1/p'))
+ifeq ($(JWT_VERSION), $(shell jwt --version 2>/dev/null | sed -nE 's/jwt ([0-9.]+).*/\1/p'))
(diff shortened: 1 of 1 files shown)


shreddedbacon:
@jackwrfuller if you can review. I missed this in #3965 because while the check to re-use the tool fails, the tool still worked. I only noticed it when I was running targets again and saw it was downloading the jwt tool each time.

I'm not sure if this means that the version flag and output that the `macOS` version provides is different to the linux version?

jackwrfuller:
> @jackwrfuller if you can review. I missed this in #3965 because while the check to re-use the tool fails, the tool still worked. I only noticed it when I was running targets again and saw it was downloading the jwt tool each time.
> 
> I'm not sure if this means that the version flag and output that the `macOS` version provides is different to the linux version?

I've tested locally, and the target seems to operate correctly now. Removing the binary results in it being redownloaded, and subsequently running the target again does nothing:

Screenshot 2025-09-12 at 10 55 26 am

I have a question however: after running the target again, is `jwt` supposed to be symlinked system-wide to `./local-dev/jwt`? Because if so, that is not functioning for me.

jackwrfuller (approved):
LGTM

PR #3984, opened 2025-09-23
fix: check output of jenkins trigger for failures

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Fixes the jenkins pipeline trigger to check the output to see if it succeeds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -238,7 +238,9 @@ pipeline {
         TOKEN = credentials('git-amazeeio-helmfile-ci-trigger')
       }
       steps {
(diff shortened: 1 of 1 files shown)


tobybellwood (commented):
Do we want it to fail, or warn? It'd suck to have to rerun in Jenkins when a manual trigger would suffice?

shreddedbacon:
> Do we want it to fail, or warn? It'd suck to have to rerun in Jenkins when a manual trigger would suffice?

How to make it warn and be obvious that it failed to trigger?

Edit: Can do a catcherr like [this](https://github.com/uselagoon/lagoon/pull/3984/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R172) I guess.

But it won't be shown unless you're looking at the pipeline directly.

Alternatively, move the trigger to be the final stage, so a failure or not doesn't break the rest of the run?

tobybellwood (approved):
as per tests in other branch, LGTM

PR #3997, opened 2025-10-15
docs: active/stanby caveats, cli usage, and information refactor

rocketeerbkw:
# General Checklist

# Description

PR is primarily to document the gotchas around active/standby (eg, not updating env vars), but also updated to document Lagoon cli usage, and general information overhaul with examples that better match the webinar.

The drush aliases are only valid for the pre-drush-8 alias system, so I have just removed that information.

# Closing issues

n/a

### DIFF
--- a/docs/using-lagoon-advanced/active-standby.md
+++ b/docs/using-lagoon-advanced/active-standby.md
@@ -1,168 +1,208 @@
 # Active/Standby
 
+Lagoon supports a method of deployment that switches production routes between two production environments. This reduces downtime, and provides easier rollbacks. It is more widely know in the industry as [blue-green deployments](https://en.wikipedia.org/wiki/Blue%E2%80%93green_deployment).
+
+For a more detailed explanation, watch our webinar:
+
 <iframe width="560" height="315" src="https://www.youtube.com/embed/urq15chLvzQ" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 
 ## Configuration
 
-To change an existing project to support active/standby you'll need to configure some project settings with the Lagoon API.
+Configuring a project to use active/standby can be done in a few steps. You should have already [set up a new project](../using-lagoon-the-basics/setup-project.md) and [deployed the first production environment](../using-lagoon-the-basics/first-deployment.md).
 
-* `productionEnviromment` should be set to the branch name of the current active environment.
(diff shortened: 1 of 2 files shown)


shreddedbacon:
> The drush aliases are only valid for the pre-drush-8 alias system, so I have just removed that information.

Maybe we should deprecate these in the API to then?

rocketeerbkw:
> > The drush aliases are only valid for the pre-drush-8 alias system, so I have just removed that information.
> 
> Maybe we should deprecate these in the API to then?

Commented too fast :smile:, opened https://github.com/uselagoon/lagoon/issues/3998 to discuss.

docs/using-lagoon-advanced/active-standby.md
```diff
@@ -1,168 +1,180 @@
...
+      }
+    }
+    ```
 
-* Routes under `production_routes.active` will direct you to `production-brancha`.
-* Routes under `production_routes.standby` will direct you to `production-branchb`.
+### Routes
 
-During an active/standby switch, the routes will swap:
+Routes are configured in the `.lagoon.yml` file, and by default are tied to a specific environment. Lagoon also supports production `active` and `standby` routes which always point to the respective `active` and `standby` environments. These routes are configured using the `production_routes` section in your `.lagoon.yml`.
```
shreddedbacon: It is also probably worth calling out that the service that the route is attached to has to be the same service in both environments.

For example, if you have `prod-left` with an `nginx` service, then `prod-right` would also have to have the same service named `nginx`. The underlying technology of those services may be completely different though.

shreddedbacon: https://gist.github.com/shreddedbacon/6583df9778960832e2d6b8d2edc64d7a

rocketeerbkw: Can we lint that at deployment time (new issue)?

shreddedbacon: Maybe, with routes in the API though I'd move to encouraging people to use it for active/standby instead of `.lagoon.yml` routes. It is far less complicated to configure with routes in the API, and we can potentially extend it in the future to support a service with a different name easier.

shreddedbacon: For now, I think just call it out in the docs, and I'll try schedule a run through of API routes with the team so I can show what I mean about how much easier it is

rocketeerbkw:
Addressed review comment but merging w/o approval so the docs are more update for an upcoming conversation.

Issue #1696, opened 2020-03-03
Environment metrics are only available to admins


shreddedbacon:
**Describe the bug**
Querying environment hits appears to only be allowed by `platform-admin`. As an owner or maintainer of a project, I thought I would be able to see how many hits my environments have had.

**To Reproduce**
1. Run query as an owner or maintainer
```
query environmentByName ($project: Int!, $environment: String!, $month: Date!){
  environmentByName(name: $environment, project: $project) {
    hitsMonth(month: $month) {
      total
    }
  }
}
```
2. Unauthorized

**Expected behavior**
See the hits for the environment

rocketeerbkw:
Same for environment hours and storage, should those also be unblocked? The current system uses the same permission for view and add/update of metrics, so we'd need to create a new permission level to ensure only admins are allowed to mutate.

smlx:
I think the decoupling of hits with the logs backend needs to happen first before making this API more widely accessible. The current structure of the query seems to cause heavy load on Elasticsearch/Opensearch.

PR #4001, opened 2025-10-19
feat: environment storage filtering support

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Extends the storages resolver on environments to support optional date range, limits, and a specific claim name.
This also opens the resolver to non-platform users to query, as long as user has view permission on the environment.

# Closes
Maybe #1696 , we just ignore the hits endpoint as discussed in that issue since that endpoint isn't great as it is.

### DIFF
--- a/services/api/src/typeDefs.js
+++ b/services/api/src/typeDefs.js
@@ -869,7 +869,7 @@ const typeDefs = gql`
     """
     Reference to EnvironmentStorage API Object, which shows the Storage consumption of this environment per day
     """
-    storages(startDate: Date, endDate: Date, limit: Int, claim: String): [EnvironmentStorage]
+    storages(startDate: Date, endDate: Date, lastDays: Int, claim: String): [EnvironmentStorage]
     """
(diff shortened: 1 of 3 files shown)


services/api/src/typeDefs.js
```diff
@@ -869,7 +869,7 @@ const typeDefs = gql`
     """
     Reference to EnvironmentStorage API Object, which shows the Storage consumption of this environment per day
     """
-    storages: [EnvironmentStorage]
+    storages(startDate: Date, endDate: Date, limit: Int, claim: String): [EnvironmentStorage]
```
rocketeerbkw: This doesn't do what I expect. The other places where we use `limit` as a field limits the number of records returned, but this is how many days in the past to return results. So maybe a name like `lastDays` or `daysAgo` is better.

rocketeerbkw (changes requested):

rocketeerbkw (approved):
Works as expected ✅

PR #3929, opened 2025-06-18
Adds a user activity log entry + control length of signed link expiry

bomoko:
This PR introduces two features
1 - it logs when a signed download link is created
2 - it allows Lagoon platform admins the ability to control the timeout of the signed links generated for downloads.

# General Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [x] PR title is ready for inclusion in changelog

# Database Migrations

- [ ] If your PR contains a database migation, it **MUST** be the latest in date order alphabetically

# Description

Explain the **details** for making this change. What existing problem does the pull request solve?

# Closing issues

Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

PR #4002, opened 2025-10-19
feat: get backup by backup id and file by file id resolver

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Additional work to support #3929 with dedicated resolvers for requesting a file that only returns a single irem. Updating the UI to have a download button action that calls this specific resolver would indicate in the audit log a more direct action to request a file.

It would be worth exploring removing the `download` and `restoreLocation` field in the normal tasks and backups resolvers in the future though once the UI and other things are updated to use the dedicated download resolvers this PR would introduce. This would limit the possibility of false audit entries from someone simply running a query using the normal resolvers.

### DIFF
--- a/services/api/src/resources/file/resolvers.ts
+++ b/services/api/src/resources/file/resolvers.ts
@@ -16,15 +16,10 @@ const { Upload } = require('@aws-sdk/lib-storage');
 const isGCS = process.env.S3_FILES_GCS || 'false'
 const bucket = process.env.S3_FILES_BUCKET || 'lagoon-files'
 
-
-export const getDownloadLink: ResolverFn = async ({ s3Key }, input, { userActivityLogger }) => {
-  const command = new GetObjectCommand({
-    Bucket: bucket,
-    Key: s3Key,
-  });
+async function generateDownloadLink(s3Key: string, userActivityLogger?: Function ) {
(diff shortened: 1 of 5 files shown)


services/api/src/resources/file/resolvers.ts
```diff
@@ -46,6 +46,23 @@ export const getDownloadLink: ResolverFn = async ({ s3Key }, input, { userActivi
...
+) => {
+  const rowsPerms = await query(sqlClientPool, taskSql.selectPermsForTask(taskId));
+
+  await hasPermission('task', 'view', {
+    project: R.path(['0', 'pid'], rowsPerms)
+  });
+
+  const rows = await query(sqlClientPool, Sql.selectTaskFileById(taskId, fileId));
+
+  const downloadUrl = await getDownloadLink(rows[0], {}, userActivityLogger)
```
rocketeerbkw: The third argument to `getDownloadLink` should be a [`context`](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/index.ts#L14) but this is only passing `userActivityLogger`. The result is that no activity/audit logs are made when invoking `getDownloadLinkByTaskFileId`.

rocketeerbkw (changes requested):

rocketeerbkw (approved):
All file link fields/resolvers generate an audit log ✅

PR #4006, opened 2025-10-23
Feature - Environment Variable "Updated" field and pending update endpoint

bomoko:
This PR introduces two things.

First, the API exposes a new set of data on the environments - namely, whether they have any pendingChanges.
These are simple, `{type:,details}` messages that indicate _that_ there are pending changes to an environment (with the implication being that a redeploy is necessary).

Secondly - it introduces an "updated" field to the environment variables table that keeps track of when the last time an env var was update. This is used to check, against the last deployment, whether or not a redeploy is necessary.

# General Checklist

# Database Migrations

# Description

# Closing issues

### DIFF
--- a/services/api/src/resources/environment/environment_redeploy.ts
+++ b/services/api/src/resources/environment/environment_redeploy.ts
@@ -35,38 +35,29 @@ _,
 
 const getPendingEnvVarChanges = async(sqlClientPool, envId) => {
       const sql = `
-    SELECT DISTINCT
-      ev.name,
-      ev.updated,
-      e.id,
-      e.name as env_name,
-	    COALESCE(
-	      IF(ev.environment IS NULL,NULL,'Environment'),
-	      IF(ev.project IS NULL,NULL,'Project'),
-	      IF(ev.organization IS NULL,NULL,'Organization')
-      ) as varsource,
-      COALESCE(MAX(d.created) OVER (PARTITION BY e.id), '1970-01-01') as last_deployment
-    FROM environment as e
-    LEFT JOIN deployment as d ON e.id = d.environment
-    INNER JOIN project as p ON p.id = e.project
-    LEFT JOIN organization as o ON o.id = p.organization
-    LEFT JOIN env_vars as ev ON (
-      ev.environment = e.id OR
-      ev.project = p.id OR
-      ev.organization = o.id
-    )
-    WHERE ev.name IS NOT NULL AND e.id = ?
-  `;
+select e.id as env_id, e.name as env_name, ev.name as envvar_name, ev.updated as envvar_updated,
+  COALESCE(
+    IF(ev.environment IS NULL,NULL,'Environment'),
+      IF(ev.project IS NULL,NULL,'Project'),
+      IF(ev.organization IS NULL,NULL,'Organization')
+      ) as envvar_source
+FROM environment as e
+INNER JOIN project as p ON p.id = e.project
+LEFT JOIN organization as o ON o.id = p.organization
+LEFT JOIN env_vars as ev ON (
+   ev.environment = e.id OR
+   ev.project = p.id OR
+   ev.organization = o.id
+)
+WHERE ev.name IS NOT NULL AND e.id = ?
+AND ev.updated > (select coalesce(max(completed), '0000-00-00 00:00:00') from deployment where environment = ? and status = ?)
+ORDER BY ev.updated asc
+`;
 
-  const results = await query(sqlClientPool, sql, [envId]);
+  const results = await query(sqlClientPool, sql, [envId, envId, 'complete']);
 
-  // Filter in memory for env vars updated after last deployment
-  const pendingChanges = results.filter(row => {
-    const updated = new Date(row.updated);
-    const lastDeployment = new Date(row.lastDeployment);
-    return updated > lastDeployment;
-  }).map(row => {
-    return {type:`Environment Variable - ${row.varsource} level`, details: row.name};
+  const pendingChanges = results.map(row => {
+    return {type:`Environment Variable - ${row.envvarSource} level`, details: row.envvarName, date: row.envvarUpdated};
   });
 
   return pendingChanges;
--- a/services/api/src/resources/environment/environment_redeploy.ts
+++ b/services/api/src/resources/environment/environment_redeploy.ts
@@ -35,26 +35,62 @@ _,
 
 const getPendingEnvVarChanges = async(sqlClientPool, envId) => {
       const sql = `
-select e.id as env_id, e.name as env_name, ev.name as envvar_name, ev.updated as envvar_updated,
-  COALESCE(
-    IF(ev.environment IS NULL,NULL,'Environment'),
-      IF(ev.project IS NULL,NULL,'Project'),
-      IF(ev.organization IS NULL,NULL,'Organization')
-      ) as envvar_source
-FROM environment as e
-INNER JOIN project as p ON p.id = e.project
-LEFT JOIN organization as o ON o.id = p.organization
-LEFT JOIN env_vars as ev ON (
-   ev.environment = e.id OR
-   ev.project = p.id OR
-   ev.organization = o.id
+WITH last_completed AS (
+  SELECT COALESCE(MAX(d.completed), TIMESTAMP('1970-01-01 00:00:00')) AS ts
+  FROM deployment d
+  WHERE d.environment = ? AND d.status = 'complete'
 )
-WHERE ev.name IS NOT NULL AND e.id = ?
-AND ev.updated > (select coalesce(max(completed), '0000-00-00 00:00:00') from deployment where environment = ? and status = ?)
-ORDER BY ev.updated desc
+SELECT *
+FROM (
+  -- Environment-scoped
+  SELECT
+    e.id   AS env_id,
+    e.name AS env_name,
+    ev.name AS envvar_name,
+    ev.updated AS envvar_updated,
+    'Environment' AS envvar_source
+  FROM environment e
+  JOIN env_vars ev ON ev.environment = e.id
+  CROSS JOIN last_completed lc
+  WHERE e.id = ?
+    AND ev.name IS NOT NULL
+    AND ev.updated > lc.ts
+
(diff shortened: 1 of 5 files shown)


services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const results = await query(sqlClientPool, sql, [envId]);
+
+  // Filter in memory for env vars updated after last deployment
+  const pendingChanges = results.filter(row => {
+    const updated = new Date(row.updated);
+    const lastDeployment = new Date(row.last_deployment);
+    return updated > lastDeployment || true;
+  }).map(row => {
+    return {type:"Environment Variable", details: row.name};
```
shreddedbacon: It might be worth including if it is an env, project, or org var in this `type`?

It starts to get a bit iffy around org variables too, since most users probably have no idea they exist or where/how/who configures them. Since they're not exposed anywhere except the organization and to org owners.

bomoko: @shreddedbacon Sweet - I'll add that, good idea. It's not necessarily going to be shown in the UI immediately, but that's really useful info.

shreddedbacon: Yeah, I figured there would be some obfuscation in the UI, like a "hey you need to deploy" but it might be neat in the future to have a page in the sidebar that can appear if there are pending changes and list those pending changes :D

services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,73 @@
...
+    )
+    WHERE ev.name IS NOT NULL AND e.id = ?
+  `;
+
+  const results = await query(sqlClientPool, sql, [envId]);
+
+  // Filter in memory for env vars updated after last deployment
+  const pendingChanges = results.filter(row => {
+    const updated = new Date(row.updated);
+    const lastDeployment = new Date(row.last_deployment);
```
shreddedbacon: This gets "JSONified"
```suggestion
    const lastDeployment = new Date(row.lastDeployment);
```

services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,73 @@
...
+    return updated > lastDeployment || true;
```
shreddedbacon: This always returns true, even if the result should be false ([fiddle](https://jsfiddle.net/yjsvohaw/))
```suggestion
    return updated > lastDeployment;
```

services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,73 @@
...
+  }).map(row => {
+    return {type:`Environment Variable - ${row.varsource} level`, details: row.name};
```
shreddedbacon: I'm wondering if it would also be worth including the `date` of the change in this payload
```suggestion
    return {type:`Environment Variable - ${row.varsource} level`, date: row.updated, details: row.name};
```

services/api/src/typeDefs.js
```diff
@@ -897,6 +897,15 @@ const typeDefs = gql`
...
     kubernetesNamespacePattern: String @deprecated(reason: "No longer in use")
+    """
+    Pending changes tell us if we need to redeploy an environment
+    """
+    pendingChanges: [EnvironmentPendingChanges]
+  }
+
+  type EnvironmentPendingChanges {
+    type: String
+    details: String
```
shreddedbacon, line 908: Also expose the change date?
```suggestion
  type EnvironmentPendingChanges {
    type: String
    details: String
    date: String
```

services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,73 @@
...
+    FROM environment as e
+    LEFT JOIN deployment as d ON e.id = d.environment
+    INNER JOIN project as p ON p.id = e.project
+    LEFT JOIN organization as o ON o.id = p.organization
+    LEFT JOIN env_vars as ev ON (
+      ev.environment = e.id OR
+      ev.project = p.id OR
+      ev.organization = o.id
+    )
+    WHERE ev.name IS NOT NULL AND e.id = ?
```
shreddedbacon: I wonder if it may be more efficient to use the `last_deployment` date to select items with an updated date that is after this date, instead of selecting everything, then handling the date comparison in the pendingChanges filter?

Thoughts?

shreddedbacon (changes requested):

services/api/src/resources/environment/environment_redeploy.ts
shreddedbacon: Then before the result is returned, do a sort by date. This way you'd be able to see the order in which they were done if this is eventually exposed in the UI as its own page of sorts.

```
{
	"environmentByKubernetesNamespaceName": {
		"deployments": [
			{
				"created": "2025-10-24 01:47:32"
			},
			{
				"created": "2025-10-24 01:45:46"
			}
		],
		"id": 30,
		"pendingChanges": [
			{
				"date": "2025-10-24 01:53:46",
				"details": "MY_SPECIAL_VARIABLE",
				"type": "Environment Variable - Environment level"
			}
		]
	}
}
```

shreddedbacon:
I'm also wondering if this may need some changes around the state of the last deployment. For example, if it fails for whatever reason before the changes are actually applied by a build, the pending changes would show as none, but not been applied.

Would it be worth selecting the last deployment based on `status` being `complete`?

Edit: Now checks deployment status :heavy_check_mark:

services/api/src/resources/environment/environment_redeploy.ts
bomoko: oh ffs, this was me testing. Thanks.

services/api/src/resources/environment/environment_redeploy.ts
bomoko: Ah, nice! Thanks

services/api/src/resources/environment/environment_redeploy.ts
bomoko: yeah - I guess I was mostly just trying to avoid doing two queries - but I _think_ I could probably get away with a subquery as well - I'll play around with it now.

shreddedbacon: All good, was just a thought. Not sure if it would be more efficient. I guess it depends on how many variables someone has at all three layers to whether it would be slower in sql

services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,100 @@
...
+export const getPendingChangesByEnvironmentId: ResolverFn = async(
+{
+    id
+},
+_,
+{ sqlClientPool, hasPermission },
+) => {
+    // Note: as it stands, the only pending changes we have now have to do
+    // with env vars, but anything can be added in the form
+    // {type:"string", details:"string"}
```
shreddedbacon: Minor, spotted this after the initial suggestion to add date to the result.

```suggestion
    // {type:"string", details:"string", date: "string"}
```

shreddedbacon (changes requested):
This all works for me. I haven't fully gone over the SQL query as it is quite a monster.

I ran through a few different scenarios with ordering variable additions etc, and I think there is only one thing that I might change, and that would be that the `completed` timestamp of the build is probably the wrong one to use, and the `created` timestamp with the `completed` status should be used.

The reason being, when the build is created, the data sent to the build contains whatever was in the API at that time. If a variable is added AFTER that build is created, it is considered a pending change still, as the last build wouldn't have known about it.

services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,100 @@
...
+    let pendingChanges = await getPendingEnvVarChanges(sqlClientPool, id);
+    return pendingChanges;
+}
+
+const getPendingEnvVarChanges = async(sqlClientPool, envId) => {
+      const sql = `
+WITH last_completed AS (
+  SELECT COALESCE(MAX(d.completed), TIMESTAMP('1970-01-01 00:00:00')) AS ts
```
shreddedbacon:
```suggestion
  SELECT COALESCE(MAX(d.created), TIMESTAMP('1970-01-01 00:00:00')) AS ts
```

services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -0,0 +1,100 @@
...
+    AND ev.name IS NOT NULL
+    AND ev.updated > lc.ts
+) AS allenvs
+ORDER BY allenvs.envvar_updated DESC;
+`;
+
+  const results = await query(sqlClientPool, sql, [envId, envId, envId, envId]);
+
+  const pendingChanges = results.map(row => {
+    return {type:`Environment Variable - ${row.envvarSource} level`, details: row.envvarName, date: row.envvarUpdated};
```
shreddedbacon: Just having the details as the variable name looks weird, having some details about what the details are is handy.

Also switched up the `type` to be a bit simpler
```suggestion
    return {type:`${row.envvarSource} Environment Variable`, details: `Variable name: ${row.envvarName}`, date: row.envvarUpdated};
```

shreddedbacon (changes requested):
Otherwise this is :heavy_check_mark: from me

services/api/src/resources/environment/environment_redeploy.ts
bomoko: yeah, great stuff. Committed @shreddedbacon

shreddedbacon (approved):
Good stuff, this will be super useful to extend in #3990 to show pending route changes too.

shreddedbacon:
Change of plan, lets change the `type` to be enums so we have a way to filter etc

shreddedbacon (approved):
Great, enum makes a lot of sense here when we consider future pending changes. We can adjust the details section overtime, or even add new fields as necessary as other pending change types appear

shreddedbacon (approved):
:partying_face: Great!

rocketeerbkw:
Minimal but extensible addition :tada:

PR #4007, opened 2025-10-27
feat: add graphql-sse endpoint

shreddedbacon:
# General Checklist

# Database Migrations

# Description

This adds `graphql-sse` to the API to support subscriptions over server sent events. 

This will allow us to deprecate `graphql-ws` or remove it entirely. The UI will need to be updated to use `graphql-sse` for subscriptions too.

This has some debugging loggers in there that probably need to be removed, or have the data contained in them trimmed down. I've left them there for the moment for troubleshooting if required.

You can use this locally with curl to view deployment changes for an environment
```
curl -X POST -N -H "Accept: text/event-stream" \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer ${TOKEN}" \
    "http://localhost:3000/graphql/stream" \
    -d '{"query":"subscription {deploymentChanged(environment: 3){id,name}}"}'
```

### DIFF
--- a/services/api/src/server.ts
+++ b/services/api/src/server.ts
@@ -73,6 +73,7 @@ export const createServer = async () => {
           await User(modelClients).userLastAccessed(currentUser);
         } catch (e) {
           logger.error('Error loading user details for SSE subscription', e.message);
+          throw new AuthenticationError('Failed to load user context for SSE subscription');
         }
       }
 
@@ -138,6 +139,11 @@ export const createServer = async () => {
 
   const server = http.createServer(async (req, res) => {
     if (req.url?.startsWith('/graphql/stream')) {
+      const origin = process.env.CORS_ORIGIN || '*';
+      res.setHeader('Access-Control-Allow-Origin', origin);
+      res.setHeader('Access-Control-Allow-Credentials', 'true');
+      res.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS');
+      res.setHeader('Access-Control-Allow-Headers', 'Content-Type, Authorization, apollo-require-preflight');
       try {
         await sseHandler(req, res);
       } catch (error: any) {
@@ -209,6 +215,7 @@ try {
             await User(modelClients).userLastAccessed(currentUser);
           } catch (e) {
             logger.error('Error loading user details for subscription', e.message);
+            throw new AuthenticationError('Failed to load user context for subscription');
           }
         }
 
--- a/services/api/src/server.ts
+++ b/services/api/src/server.ts
@@ -16,6 +16,7 @@ import { keycloakHasPermission, legacyHasPermission } from './util/auth';
 import R from 'ramda';
 import { AuthenticationError } from 'apollo-server-express';
 const { useServer } = require('graphql-ws/lib/use/ws');
+const { createHandler } = require('graphql-sse/lib/use/http');
 
 export const createServer = async () => {
   logger.verbose('Starting the api...');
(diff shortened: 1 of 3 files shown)


services/api/src/server.ts
```diff
@@ -16,17 +16,145 @@ import { keycloakHasPermission, legacyHasPermission } from './util/auth';
...
   await configureApp();
 
   const port = toNumber(getConfigFromEnv('PORT', '3000'));
-  const server = http.createServer(app);
-  server.setTimeout(900000); // higher Server timeout: 15min instead of default 2min
 
-  // apolloServer.installSubscriptionHandlers(server);
+  const schema = await getSchema();
+
+  const sseHandler = createHandler({
```
CGoodwin90, line 30: Should we be including some kind of timeout? Not sure if `SSE` has any inbuilt defaults.

services/api/src/server.ts
```diff
@@ -16,17 +16,145 @@ import { keycloakHasPermission, legacyHasPermission } from './util/auth';
...
+          keycloakUsersGroups = await User(modelClients).getAllGroupsForUser(userId);
+          serviceAccount = await keycloakGrantManager.obtainFromClientCredentials(undefined, undefined);
+          currentUser = await User(modelClients).loadUserById(userId);
+          const userRoleMapping = await keycloakAdminClient.users.listRealmRoleMappings({ id: currentUser.id });
+          platformOwner = userRoleMapping.some(role => role.name === 'platform-owner');
+          platformViewer = userRoleMapping.some(role => role.name === 'platform-viewer');
+          groupRoleProjectIds = await User(modelClients).getAllProjectsIdsForUser(currentUser.id, keycloakUsersGroups);
+          await User(modelClients).userLastAccessed(currentUser);
+        } catch (e) {
+          logger.error('Error loading user details for SSE subscription', e.message);
```
CGoodwin90, line 75: We're just logging an error here and still creating the `context`, can we throw a new error if we can't get the user details?

services/api/src/server.ts
```diff
@@ -16,17 +16,145 @@ import { keycloakHasPermission, legacyHasPermission } from './util/auth';
...
+        errors: errors,
+        params: params
+      });
+    },
+    onComplete: (req, params) => {
+      logger.verbose('SSE connection completed', { params });
+    },
+  });
+
+  const server = http.createServer(async (req, res) => {
```
CGoodwin90, line 140: Do we need to worry about adding `CORS` with `SSE`, or does it handle it internally?

shreddedbacon: Unsure, how is CORS currently handled with the API for standard requests (not websockets)?

services/api/src/server.ts
shreddedbacon: We do the same thing in the existing check [here](https://github.com/uselagoon/lagoon/blob/v2.29.2/services/api/src/server.ts#L83) which is unmodified.

If it is changed here, it probably needs to also be changed in that section too. Do you have a suggestion on how that may work?

services/api/src/server.ts
shreddedbacon: Not sure, the whole idea of SSE though is essentially long running connections that are kept alive.

shreddedbacon:
Force push is just rebase, no code changes.

services/api/src/server.ts
CGoodwin90: Oh actually the 15m [timeout](https://github.com/uselagoon/lagoon/pull/4007/files#diff-a0949ca6ce80ca47adf27b3eb9d29920fa54cc25273251f72e3ea133afa416e3R157) is being applied for SSE too. I'm wondering if that should only be applied [here](https://github.com/uselagoon/lagoon/pull/4007/files#diff-a0949ca6ce80ca47adf27b3eb9d29920fa54cc25273251f72e3ea133afa416e3R154) if they're meant to be long running?

services/api/src/server.ts
CGoodwin90: I was thinking of simply throwing an error? 

```
  } catch (e) {
    logger.error('Error loading user details for SSE subscription', e.message);
    throw new AuthenticationError('Failed to load user context for SSE subscription');
  }
```

We're doing the same in the [onConnect](https://github.com/uselagoon/lagoon/blob/main/services/api/src/server.ts#L150) for the `websocket` setup & the [authenticate](https://github.com/uselagoon/lagoon/pull/4007/files#diff-a0949ca6ce80ca47adf27b3eb9d29920fa54cc25273251f72e3ea133afa416e3R47) for the `sseHandler`, can probably do the same for the `websocket context`. Not sure if there's further implications for doing this though?

services/api/src/server.ts
CGoodwin90: It's configured in the `express` app setup with basic defaults [here](https://github.com/uselagoon/lagoon/blob/v2.29.2/services/api/src/app.ts#L41) & disabled in the apollo(websocket) config. It's being bypassed for `SSE` [here](https://github.com/uselagoon/lagoon/pull/4007/files#diff-a0949ca6ce80ca47adf27b3eb9d29920fa54cc25273251f72e3ea133afa416e3R140), so we could add some standard headers so it's consistent? Just an example, same as what we have in the `express` setup?

```
  if (req.url?.startsWith('/graphql/stream')) {
    const origin = process.env.CORS_ORIGIN || '*';
    res.setHeader('Access-Control-Allow-Origin', origin);
    res.setHeader('Access-Control-Allow-Credentials', 'true');
    res.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS');
    res.setHeader('Access-Control-Allow-Headers', 'Content-Type, Authorization, apollo-require-preflight');
```

services/api/src/server.ts
shreddedbacon: Probably fine to throw an error, if the user isn't authenticated, it'll fail somewhere further down anyway.

services/api/src/server.ts
shreddedbacon: Will give it a whirl and see what happens :smiley:

services/api/src/server.ts
shreddedbacon: They won't time out if someone has an active stream open

services/api/src/server.ts
shreddedbacon: Not seeing any issues with setting these

CGoodwin90 (approved):

PR #4009, opened 2025-10-27
chore: yarn upgrades

tobybellwood:
This PR jumps into the API and laser-targets some easy updates/upgrades:

It primarily updates dependencies to newer versions and refactors the custom Lokka HTTP transport to remove reliance on the deprecated `@lagoon/lokka-transport-http` package. It also introduces a minor configuration change to the MariaDB client to improve insert ID handling.

Dependency updates:

* Upgraded `axios` from version `0.21.1` to `1.12.0` in both `node-packages/commons/package.json` and `services/auth-server/package.json`, and updated `node-fetch` from `2.6.0` to `3.3.2` in `node-packages/commons/package.json`. Also updated `mariadb` from `2.5.2` to `3.4.5` in `services/api/package.json`. [[1]](diffhunk://#diff-3a1708021bedbdbec11dc44ed6bd9c65f5306143e761cfbdeba0c51aec118e53L25-R30) [[2]](diffhunk://#diff-de93a8c8810753191dec926e55ad0483e22ca80e7e910ecbf791ee06f4ed041bL28-R28) [[3]](diffhunk://#diff-8aae71eebcc8f253ba93c534c95dc5bde66088ee656adfb7cf85cff10b3a7fdbL59-R59)

Lokka transport refactor:

* Removed the dependency on `@lagoon/lokka-transport-http` and implemented a custom `Transport` class in `node-packages/commons/src/lokka-transport-http-retry.ts`, including updated retry logic, improved error handling, and TypeScript typings. [[1]](diffhunk://#diff-d3ef2bc7e66d76fe12933466d4f8986d42d8698451e0e9bf3cbf671a60f5911cL1-R24) [[2]](diffhunk://#diff-d3ef2bc7e66d76fe12933466d4f8986d42d8698451e0e9bf3cbf671a60f5911cL26-L55) [[3]](diffhunk://#diff-d3ef2bc7e66d76fe12933466d4f8986d42d8698451e0e9bf3cbf671a60f5911cL66-L82)

MariaDB configuration improvement:

* Added the `insertIdAsNumber: true` option to the MariaDB client pool configuration in `services/api/src/clients/sqlClient.ts` to ensure insert IDs are returned as numbers.

PR #4013, opened 2025-11-07
fix: api error when accessing fields stored as `bigint`

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

The mariadb client version was upgraded from 2.x to 3.x in https://github.com/uselagoon/lagoon/pull/4009 which caused an error to be thrown for any api fields stored as mariadb `bigint`. Fix is to configure the client to use the same conversion mechanism as 2.x.

It's also [documented that `decimal` fields](https://mariadb.com/docs/connectors/mariadb-connector-nodejs/connector-nodejs-promise-api#migrating-from-2.x-or-mysql-mysql2-to-3.x) have the same config option, but when I tested those fields in the api I didn't get any errors, so leaving that one as-is.

# Closing issues

n/a

### DIFF
--- a/services/api/src/clients/sqlClient.ts
+++ b/services/api/src/clients/sqlClient.ts
@@ -9,7 +9,8 @@ export const config = {
   password: getConfigFromEnv('API_DB_PASSWORD', 'api'),
   database: getConfigFromEnv('API_DB_DATABASE', 'infrastructure'),
   connectionLimit: toNumber(getConfigFromEnv('API_DB_CONN_LIMIT', '80')),
-  insertIdAsNumber: true
+  insertIdAsNumber: true,
+  bigIntAsNumber: true,
 };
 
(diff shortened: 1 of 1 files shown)


shreddedbacon:
Not sure why jenkins is struggling to run the `k3d/stable-local-stack`. I was able to run `k3d/stable-local-stack` for this branch without issue locally.

shreddedbacon (approved):
Confirmed that the storage value is returned correctly.

Also verified with adding a problem that uses the decimal type, which also succeeds.
```
mutation {
	addProblem(input:{
  	environment: 3
    identifier: "3"
    data: "{\"data\":\"data\"}"
    severityScore: 0.452
    source: "manual"
  }) {
  environment{
  	name
  }
  data
  severityScore
  }
}
```

We should get some test coverage for these though, will see if this is easy enough to implement in a follow up.

PR #4017, opened 2025-11-12
feat/refactor: signed file upload urls and files from bucket

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Rather than upload files to the API for tasks, and have the API handle the file upload. A new resolver will now return information for being able to use pre-signed POST URLs to upload files directly to the bucket.

Additionally, this removes the API DB references for task files. Task file listing will now list the objects directly from the bucket directory for the task. This was done so that files uploaded by pre-signed URLs wouldn't have a reference needed to be managed, or wait for the file to be uploaded. This has the added benefit of being able to potentially add lifecycle policy to the task file directory in the bucket so that old files can be aged out without needing to worry about an API reference.

The `drush sql-dump` and `drush archive-dump` tasks have been updated with new tasks that utilise the new resolver.

Note: the local-stack has issues downloading the file from the UI due to minio being http and the UI being https. With this https://github.com/uselagoon/lagoon-charts/pull/840 all pods get the local-stack CA mounted into them, and this allows setting the minio endpoint used to be https and then everything works. Setting minio to https without that PR results in all sorts of failures due to untrusted certs. You can use devtools to inspect the response payload when clicking the file in the UI to see the download url, and then copy+paste it to verify the file downloads.


rocketeerbkw:
Discussed out of band, a few small changes to types are incoming.

rocketeerbkw (approved):
Tested locally, everything working.

PR #3983, opened 2025-09-16
feat: use keycloak quick-theme to generate simple lagoon-v2 theme

shreddedbacon:
# General Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for inclusion in changelog

# Database Migrations

- [ ] If your PR contains a database migation, it **MUST** be the latest in date order alphabetically

# Description

The current `lagoon` theme in keycloak has numerous bugs with rendering certain pages. This introduces a new theme `lagoon-v2` that uses [quick-theme](https://www.keycloak.org/ui-customization/quick-theme) to just change colors on the default theme with some image overrides.

This means not having to worry too much about changing too much, but still keeping some styling consistent with the UI.

The colors chosen in this may need adjusting, I just picked them close enough to what the current UI uses. It defaults to dark mode though, as this is not "automatic" in keycloak, it is set in the realm and defaults to dark mode

PR #4018, opened 2025-11-12
Fixes otp list in lagoon theme

bomoko:
In [this commit](https://github.com/keycloak/keycloak/commit/31aefd14893ac1ef1561fc40e76a44259c77e9dd) we see that there was a change in the way supported OTP integrations is exposed in the theming layer.

This PR brings the Lagoon theme in line with the current practice in Keycloak > v20.0.0

Testing instructions:
1 - spin up keycloak (either in localstack or using something like `make api-development`)
2 - create a user or pick an existing user in the `lagoon` realm (this can be done on the `master` realm, but be sure to change the theme for the realm to `lagoon`)
3 - edit the user, set a password, and on Details->Required user actions, select `Configure OTP`. Save.
4 - Log in as user from (2-3) and behold the OTP setup screen.

if you do steps (1-4) off `main` you'll hit an error after logging in, but with the change here you'll be able to set up your OTP and log in.

### DIFF
--- a/services/keycloak/themes/lagoon/login/login-config-totp.ftl
+++ b/services/keycloak/themes/lagoon/login/login-config-totp.ftl
@@ -10,7 +10,7 @@
                 <p>${msg("loginTotpStep1")}</p>
 
                 <ul id="kc-totp-supported-apps">
-                    <#list totp.policy.supportedApplications as app>
+                    <#list totp.supportedApplications as app>
                         <li>${msg(app)}</li>
                     </#list>
                 </ul>


bomoko:
Note, this _works_ but there will need to be a follow up PR that fixes the theming, since scrolling is broken.

shreddedbacon:
https://github.com/uselagoon/lagoon/pull/3983

bomoko:
> #3983

This is great @shreddedbacon . I'd still say let's get this in and fix the current theme so it doesn't 500 _if_ someone sets up OTP. But then roll out the new theme in the next release.

shreddedbacon:
Just spun this up, it does make the configuration for OTP load, but the page doesn't scroll properly.

Adding this style override to the main div adds the scrollbar. I have no idea if this impacts other parts of the login page, but I can't imagine it breaks anything else except maybe letting other things scroll if needed.

```
diff --git a/services/keycloak/themes/lagoon/login/template.ftl b/services/keycloak/themes/lagoon/login/template.ftl
index 8b9efa1cb..a98cbd971 100644
--- a/services/keycloak/themes/lagoon/login/template.ftl
+++ b/services/keycloak/themes/lagoon/login/template.ftl
@@ -113,7 +113,7 @@
         </#if>
       </header>  -->
 
-      <div id="kc-content" class="kc-content">
+      <div id="kc-content" class="kc-content" style="overflow-y: auto;">
         <div id="kc-content-wrapper" class="kc-content-wrapper">
 
           <#-- App-initiated actions should not see warning messages about the need to complete the action -->
```

shreddedbacon (approved):
Happy to forego the scrollbar fix and just fix this issue for now. Then focus on #3983

rocketeerbkw:
> I'd still say let's get this in and fix the current theme

There are way more broken parts than just this. Some pages are just missing altogether and throw errors. I'll merge with the understanding that the current theme is still not reliable and realistically not be used by anyone.

PR #4019, opened 2025-11-13
Change: Adds a migration to set a default value for metadata

CGoodwin90:
# General Checklist

# Database Migrations

# Description

Reverts the query change in [PR #4014](https://github.com/uselagoon/lagoon/pull/4014/files#diff-16177503259886f86d8e68b989d601c143690780b551d493170205e872a231a0R1067) & adds a migration to set a default value for `metadata`.

### DIFF
--- a/services/api/database/migrations/20251112052711_set_default_metadata.js
+++ b/services/api/database/migrations/20251112052711_set_default_metadata.js
@@ -1,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+// Need to utilize raw here instead of alterTable as MySQL requires
+// default values for JSON columns to be written as an expression
+exports.up = async function(knex) {
+  return knex.schema
+    .raw(`UPDATE project SET metadata = '{}' WHERE metadata IS NULL;`)
+    .raw(`ALTER TABLE project MODIFY metadata JSON NOT NULL DEFAULT ('{}');`);
+};
+
+/**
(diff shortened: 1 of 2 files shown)


services/api/database/migrations/20251112052711_set_default_metadata.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function(knex) {
+  return knex.schema
+    .raw(`ALTER TABLE project MODIFY metadata JSON NULL;`)
+    .raw(`UPDATE project SET metadata = NULL WHERE metadata = '{}';`);
```
bomoko, line 20: I'm not sure we need this - what are the implications of rolling this back? We don't necessarily want to empty these columns

bomoko (changes requested):

shreddedbacon (approved):
Does what it says on the tin

services/api/database/migrations/20251112052711_set_default_metadata.js
shreddedbacon: These rollbacks are honestly kinda silly though too, we've never needed to pull this out ever and there is no actual procedure for it. We generally just restore from a prior backup and run again.

bomoko: fair enough - then let's send it

bomoko (approved):

PR #4023, opened 2025-11-14
docs: release notes for 2.29.2

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

Adjusts release notes for failed 2.29.1 and adds page for 2.29.2.

# Closing issues

n/a

### DIFF
--- a/docs/releases/2.29.2.md
+++ b/docs/releases/2.29.2.md
@@ -1,0 +1,7 @@
+#### Release Links
+* lagoon [v2.29.2](https://github.com/uselagoon/lagoon/releases/tag/v2.29.1)
+* lagoon-core chart [1.58.0](https://github.com/uselagoon/lagoon-charts/releases/tag/lagoon-core-1.58.0)
+
+## Release Notes
+
+This release fixes issues in 2.29.0-2.29.1, but is otherwise a standard Lagoon release.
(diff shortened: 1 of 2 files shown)


docs/releases/2.29.2.md
```diff
@@ -0,0 +1,7 @@
+#### Release Links
+* lagoon [v2.29.2](https://github.com/uselagoon/lagoon/releases/tag/v2.29.1)
+* lagoon-core chart [1.58.0](https://github.com/uselagoon/lagoon-charts/releases/tag/lagoon-core-1.58.0)
+
+## Release Notes
+
+This release fixes issues in 2.29.0-2.29.1, but is otherwise a standard Lagoon release.
```
shreddedbacon, line 7: > is a standard Lagoon release

This seems a bit pointless to mention, what is a standard release vs one that has deprecation or other more detailed notes?

shreddedbacon: This wasn't included in the prior few release notes pages either

rocketeerbkw: My reading is that standard means there are no special circumstances that need to be mentioned in order to upgrade. The standard process can be followed. The language was taken from the last time I did a release 2.25.0 and the previous couple releases before that.

Before "standard" was being copy/pasted, they used to say "There are no required actions or considerations with this release. As always, we suggest upgrading all minor versions." I'm ok with any language that conveys "nothing special here."

shreddedbacon: In both cases the omission of the text is enough to represent that there are no actions required. If there are action items, they would be listed.

shreddedbacon: The policy describes what people need to do https://docs.lagoon.sh/releases/policy/ and calls out the following

> You should check all upgrade requirements and deprecation notices for all versions as you upgrade.

shreddedbacon: This needs to be merged. I'm going to just ignore the wording as much as it annoys me for now so that it can be merged.

shreddedbacon (approved):

PR #4024, opened 2025-11-16
test: optional traefik installation

shreddedbacon:
# General Checklist

# Database Migrations

# Description
This is an experimental branch to see which tests pass/fail when using Traefik instead of ingress-nginx, requires https://github.com/uselagoon/lagoon-charts/tree/traefik.

## Route annotations
Lagoon allows users to [set annotations](https://docs.lagoon.sh/concepts-basics/lagoon-yml/#ingress-annotations) on their routes. If you've set annotations that are specific to `ingress-nginx`, these will likely not work unless they're listed as supported in Traefik. 

Redirect annotations [described in the docs](https://docs.lagoon.sh/concepts-basics/lagoon-yml/#ingress-annotations-redirects) are also unsupported.

## Idling
`aergia-controller` now ships with support for `traefik` and `ingress-nginx`

# Linked
* https://github.com/uselagoon/build-deploy-tool/pull/472
* https://github.com/uselagoon/aergia-controller/pull/65
* https://github.com/uselagoon/lagoon-charts/pull/907


shreddedbacon:
https://traefik.io/blog/transition-from-ingress-nginx-to-traefik and https://doc.traefik.io/traefik/v3.5/reference/routing-configuration/kubernetes/ingress-nginx/

~~Update: There is a weird bug with traefik and the way ssl-redirect annotations, a community user has submit a PR to fix waiting for review. There is an image override in the `traefik` branch of lagoon-charts with an image built from that PR which does indeed solve the issue.~~

This only matters if using the ingress-nginx provider in `traefik`

shreddedbacon:
~~New [charts branch](https://github.com/uselagoon/lagoon-charts/tree/ingress-controllers) that can install different controllers.~~

Update: Not using that alternative branch now. We've settled on supporting [Traefik](https://github.com/uselagoon/lagoon-charts/tree/traefik)

PR #4026, opened 2025-11-18
fix: Better Environment Variable Precedence handling for pending deploys

bomoko:
The logic of the current pending deploy has an inaccuracy.

Given the possibility of defining an env var at three different levels
- Org
- Project
- Environment
and given that the _same_ environment variable can (1) appear at every level, but (2) only the most _specific_ instance of the defined variable will be deployed into the environment (so something defined at both Project and Environment levels will see only the environment variable be deployed), this same logic of precedence should apply to what kind of "pending changes" show up for an environment.

This change will deal with this case generally AND with the case where an environment on a higher level might be considered as a pending change, but because the variable at the lower level has been successfully deployed into the environment, it is disregarded.

# General Checklist

# Database Migrations

# Description

# Closing issues

### DIFF
--- a/services/api/src/resources/environment/environment_redeploy.ts
+++ b/services/api/src/resources/environment/environment_redeploy.ts
@@ -54,7 +54,7 @@ const getPendingEnvVarChanges = async(sqlClientPool, envId) => {
         ev.name AS envvar_name,
         ev.updated AS envvar_updated,
         lc.ts as env_last_updated,
-        IF(ev.updated > lc.ts, "deploy", "no-deploy") as must_deploy,
+        IF(ev.updated > lc.ts, 'deploy', 'no-deploy') as must_deploy,
         'Environment' AS envvar_source,
         3 as envvar_priority
       FROM environment e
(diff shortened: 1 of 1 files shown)


services/api/src/resources/environment/environment_redeploy.ts
```diff
@@ -38,67 +38,104 @@ _,
...
+              // First instance of a var to be deployed
+              overrideMap.set(row.envvarName, row);
+          }
+      }
+  })
+
+  const pendingChanges =  Array.from(overrideMap.values()).map((row) => {
+      return {
+        type:environmentPendingChangeTypes.ENVVAR,
+        details: `Variable name: ${row.envvarName} (source: ${row.envvarSource} )`,
```
shreddedbacon:
```suggestion
        details: `Variable name: ${row.envvarName} (source: ${row.envvarSource})`,
```

Random space after the source
```
"details": "Variable name: SSMTP_MAILHUB (source: Organization )",
```

shreddedbacon (changes requested):
The SQL is quite complex to read, but testing locally it does appear to work as intended.

Only one thing is a weird errant space in the output which was there before this pullrequest, but would be good to fix up now.

shreddedbacon (approved):

PR #4027, opened 2025-11-18
feat: change build name for variable only deployments

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Extends deployment type with a new field `buildType` that is an enum, the default is `BUILD`, but also supports `VARIABLES`. This will change depending on if the build has been flagged to be a [variables only deployment](https://github.com/uselagoon/build-deploy-tool/pull/430). The existence of the variables only deployment flag also changes the name of the build to be `lagoon-variables-xyz` instead of `lagoon-build-xyz` to also differentiate in the list of deployments. The `CONTAINER_REGISTRY` scope should be treated the same as `BUILD`, as it is only consumed during a build.

# Usage
An example of triggering a variables only deployment. This sets the build variable `LAGOON_VARIABLES_ONLY` to true.
```
mutation deployEnvironmentLatest{
	deployEnvironmentLatest(input:{
  	environment:{
    	id: 3
      project: {
      	id: 18
      }
    }
    buildVariables: [
    	{name: "LAGOON_VARIABLES_ONLY", value: "true"}
    ]
    returnData: true
  })
}
```

# Limitations
Variables only deployments have some limitations. Only `RUNTIME` and `GLOBAL` scope are considered as being changed (`GLOBAL` because it is considered `RUNTIME` and `BUILD`). However, if the variable being changed is `GLOBAL` or `BUILD` and is required to influence the outcome of built images, or change behaviour of something during a build, then a full deployment would be required. In the event that a `GLOBAL` scope is changed, only the runtime version of that variable will be updated in the environment.

# Requirements

This shouldn't be merged until https://github.com/uselagoon/build-deploy-tool/pull/430 has been evaluated. The two are sort of linked together, merging this without the other just seems silly.

### DIFF
--- a/node-packages/commons/src/util/lagoon.ts
+++ b/node-packages/commons/src/util/lagoon.ts
@@ -4,6 +4,9 @@ import { EnvKeyValue } from '../api';
 export const generateBuildId = () =>
   `lagoon-build-${Math.random().toString(36).substring(7)}`;
 
+export const variableOnlyBuild = () =>
+  `lagoon-variables-${Math.random().toString(36).substring(7)}`;
+
 export const generateTaskName = () =>
   `lagoon-task-${Math.random().toString(36).substring(7)}`;
 
--- a/node-packages/commons/src/util/lagoon.ts
+++ b/node-packages/commons/src/util/lagoon.ts
@@ -4,7 +4,7 @@ import { EnvKeyValue } from '../api';
 export const generateBuildId = () =>
   `lagoon-build-${Math.random().toString(36).substring(7)}`;
 
-export const variableOnlyBuild = () =>
+export const generateVariableOnlyBuildId = () =>
   `lagoon-variables-${Math.random().toString(36).substring(7)}`;
 
 export const generateTaskName = () =>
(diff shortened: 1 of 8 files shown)


node-packages/commons/src/util/lagoon.ts
```diff
@@ -4,6 +4,9 @@ import { EnvKeyValue } from '../api';
 export const generateBuildId = () =>
   `lagoon-build-${Math.random().toString(36).substring(7)}`;
 
+export const variableOnlyBuild = () =>
```
bomoko:
```suggestion
export const generateVariableOnlyBuildId = () =>
```
this would keep the naming of the closely related functions similar.

bomoko (changes requested):
Looks good - I just wonder about the name of the function generating the build id

bomoko:
@shreddedbacon - is this good to look at again?

shreddedbacon:
Yep, just as long as the merge requirements are not ignored, the [other PR](https://github.com/uselagoon/build-deploy-tool/pull/430) still needs review.

bomoko (approved):

PR #4028, opened 2025-11-20
feat: support to delete restored backup files

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Support for being able to delete a restored backup file.

This can be useful if for some reason the restored file is corrupt and you want to trigger a new retrieval.

For local-stack testing, you'll need to use the API directly, deploy an environment and run a backup. Once the backup completes, the UI will show the retrieval option. Retrieve the backup and see the download link (the download link will fail to download in local-stack due to issue described in description in #4017). Get the backup ID and run the delete restore mutation. Refresh the UI after deleting the restore and see the download return to a retrieve button


services/api/src/resources/backup/resolvers.ts
```diff
@@ -18,63 +18,69 @@ import { AuditType } from '@lagoon/commons/dist/types';
...
+      'S3_BAAS_ACCESS_KEY_ID',
+      'XXXXXXXXXXXXXXXXXXXX'
+    );
+    secretAccessKey = getConfigFromEnv(
+      'S3_BAAS_SECRET_ACCESS_KEY',
+      'XXXXXXXXXXXXXXXXXXXX'
+    );
+  }
+
+  return [accessKeyId, secretAccessKey];
```
bomoko: This makes sense to change to an object - the destructuring of it and creation of it can remain the same, it's just about surrounding it with `{}` rather than `[]` - this gets us away from the inherent ambiguity of positional return values with very little effort.

services/api/src/resources/backup/resolvers.ts
```diff
@@ -147,6 +153,71 @@ const getRestoreLocation = async (backupId, restoreLocation, sqlClientPool, user
   return [restoreLocation, restoreSize];
 };
 
+const deleteRestoreLocation = async (backupId, restoreLocation, sqlClientPool) => {
+  let restoreSize = 0;
+  const rows = await query(sqlClientPool, Sql.selectBackupByBackupId(backupId));
```
bomoko, line 167: although you're checking this in the incoming resolver - given that this is a function that is logically distinct from the client function - you should be checking that rows are returned here before making use of them in the next line.

services/api/src/resources/backup/resolvers.ts
```diff
@@ -147,6 +153,71 @@ const getRestoreLocation = async (backupId, restoreLocation, sqlClientPool, user
...
+  // https://{endpoint}/{bucket}/{key}
+  const s3LinkMatch = /([^/]+)\/([^/]+)\/([^/]+)/;
+
+  if (R.test(s3LinkMatch, restoreLocation)) {
+    const s3Parts = R.match(s3LinkMatch, restoreLocation);
+
+    let [accessKeyId, secretAccessKey] = await getRestoreLocationS3Creds(sqlClientPool, project.projectId)
+
+    let awsS3Parts;
+    const awsLinkMatch = /s3\.([^.]+)\.amazonaws\.com\//;
```
bomoko: This seems to be defined in multiple places, suggesting that this would be usefully raised to a module constant.

services/api/src/resources/backup/resolvers.ts
```diff
@@ -147,6 +153,71 @@ const getRestoreLocation = async (backupId, restoreLocation, sqlClientPool, user
...
+      },
+      forcePathStyle: true,
+      signatureVersion: 'v4',
+      endpoint: `${restoreLocationURL.protocol}//${R.prop(1, s3Parts)}`,
+      region: awsS3Parts ? R.prop(1, awsS3Parts) : 'us-east-1'
+    });
+
+    try {
+      const headObjectResponse = await s3Client.send(new HeadObjectCommand({
+        Bucket: R.prop(2, s3Parts),
```
bomoko: What happens here if line 172 evaluates to `false`?

services/api/src/resources/backup/resolvers.ts
```diff
@@ -147,6 +153,71 @@ const getRestoreLocation = async (backupId, restoreLocation, sqlClientPool, user
...
+        Key: R.prop(3, s3Parts)
+      }));
+      restoreSize = headObjectResponse.ContentLength ?? restoreSize;
+
+      // delete the object from the bucket
+      await s3Client.send(new DeleteObjectCommand({
+        Bucket: R.prop(2, s3Parts),
```
bomoko: given these are used in several places, it suggests that it would be good to introduce symbols for bucket and key around line 169 to make this more explicit rather than repeating the rambda prop call.

services/api/src/resources/backup/resolvers.ts
```diff
@@ -147,6 +153,71 @@ const getRestoreLocation = async (backupId, restoreLocation, sqlClientPool, user
...
+        Key: R.prop(3, s3Parts)
+      }));
+
+      await query(
+        sqlClientPool,
+        Sql.deleteRestore({
+          backupId
+        })
+      );
+    } catch(err) {
```
bomoko, line 191: What's the logic behind repeating this query in the catch? Perhaps a comment signposting what's being achieved here would be useful?

services/api/src/resources/backup/resolvers.ts
```diff
@@ -498,6 +569,80 @@ export const updateRestore: ResolverFn = async (
...
+      backupId,
+    }
+  },
+  { sqlClientPool, hasPermission, userActivityLogger, adminScopes }
+) => {
+
+  let rows = await query(sqlClientPool, Sql.selectBackupByBackupId(backupId));
+  if (rows.length == 0) {
+    throw new Error(
+      `Unauthorized: You don't have permission to "delete" on "restore"`
```
bomoko, line 570: is this message to throw the user off, assuming they're trying random backup ids?
Would it make sense to log what's really going on?

bomoko (changes requested):

services/api/src/resources/backup/resolvers.ts
shreddedbacon: > is this message to throw the user off, assuming they're trying random backup ids?

Yes

services/api/src/resources/backup/resolvers.ts
shreddedbacon: I'm not sure I understand this one. If this evaluates to `false`, then it doesnt match an AWS domain
```
    if (R.test(awsLinkMatch, restoreLocation)) {
      awsS3Parts = R.match(awsLinkMatch, restoreLocation);
    }
```

and then this will just default to `us-east-1`
```
region: awsS3Parts ? R.prop(1, awsS3Parts) : 'us-east-1'
```

Unless I'm missing something else?

bomoko: my bad, I was reading this incorrectly

shreddedbacon:
Just rebased in the last force push, the last commit is still the only thing that has changed otherwise.

bomoko (approved):
Happy with changes, looking good.

PR #4029, opened 2025-11-20
refactor: stop internal use of deprecated `bytesUsed` field

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

`bytesUsed` was deprecated in Lagoon v2.18.0. This PR
- Refactors the api to separate the deprecated field into it's own resolvers for easier removal in the future.
- Also fixes an error where the `environment` on `EnvironmentStorage` is always `null`.
- Removes internal use of the field by actions-handler, which is a breaking change.

# Closing issues

n/a


rocketeerbkw:
Tested with local stack, and ready to review now.

shreddedbacon (approved):
Works fine. Requires that people are on a newer version of storage-calculator for the changes in actions-handler, which as long as they're on a current version of lagoon-remote, shouldn't be a problem.

PR #4031, opened 2025-11-25
feat: extend deploymentsByFilter options

shreddedbacon:
# General Checklist

# Database Migrations

# Description

Adds some additional filtering options to the deploymentsByFilter query to be able to limit the number of deployments returned per environment or in total, and filter by environment type.

I set some arbitrary defaults for limits under certain conditions, this is to keep the current view in the `deployments` page in the UI to return the same information it currently does when there are no other changes to the filtering.

These limits are as follows
* If a user requests the `environmentType` filter, but does not provide a `limitPerEnvironment`, the limit is set to 1.
* If a user requests a status of `complete`, `failed`, or `cancelled`, then the `limitPerEnvironment` is set to 5.
In both cases, setting the `limitPerEnvironment` to a higher or lower value will override the default.

These combined with the existing status, start/end date, and deploytarget filters should offer more flexibility for users.

These filters could then be exposed on the all deployments overview page in the UI, allowing users to modify the filters to their liking.

### DIFF
--- a/services/api/src/resources/deployment/resolvers.ts
+++ b/services/api/src/resources/deployment/resolvers.ts
@@ -190,14 +190,14 @@ export const getDeploymentsByFilter: ResolverFn = async (
   if (environmentType && !limitPerEnvironment) {
     // if environment type is set, assume only the latest deployment for the environment
     // is requested depending on the `deploymentStatus` options chosen
-    limitPerEnvironment = 1
+    limitPerEnvironment = 1;
   }
 
-  let limitedStatuses = ["complete", "failed", "cancelled"]
-  if (deploymentStatus.some(element => limitedStatuses.includes(element)) && !limitPerEnvironment) {
+  let limitedStatuses = ["complete", "failed", "cancelled"];
+  if (!limitPerEnvironment && deploymentStatus.some(element => limitedStatuses.includes(element))) {
     // if a user requests `complete`, `failed`, or `cancelled` status builds but does not include a `limitPerEnvironment` limit
     // set the limit to the last 5 deployments to limit the amount of returned data
-    limitPerEnvironment = 5
+    limitPerEnvironment = 5;
   }
   logger.info( Sql.selectDeploymentsByFilter(
     {
@@ -211,7 +211,7 @@ export const getDeploymentsByFilter: ResolverFn = async (
       limitPerEnvironment,
       limit
     }
-  ))
+  ));
   const rows = await query(sqlClientPool, Sql.selectDeploymentsByFilter(
     {
       startDate,
@@ -224,7 +224,7 @@ export const getDeploymentsByFilter: ResolverFn = async (
       limitPerEnvironment,
       limit
     }
-  ))
+  ));
 
   const withK8s = projectHelpers(sqlClientPool).aliasOpenshiftToK8s(rows);
   return withK8s;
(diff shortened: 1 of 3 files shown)


services/api/src/resources/deployment/resolvers.ts
```diff
@@ -178,39 +187,45 @@ export const getDeploymentsByFilter: ResolverFn = async (
...
+  if (environmentType && !limitPerEnvironment) {
+    // if environment type is set, assume only the latest deployment for the environment
+    // is requested depending on the `deploymentStatus` options chosen
+    limitPerEnvironment = 1
   }
 
-  if (endDate) {
-    queryBuilder = queryBuilder.where('deployment.created', '<=', input.endDate);
+  let limitedStatuses = ["complete", "failed", "cancelled"]
+  if (deploymentStatus.some(element => limitedStatuses.includes(element)) && !limitPerEnvironment) {
```
bomoko: ___teensy___ performance note - these conjuncts can be swapped, since we have short circuit evaluation, rather than iterating over deploymentStatus.some, if `!limitPerEnvironment` evaluates to false, we skip a whole set of operations (i.e. swap these two checks around).

services/api/src/resources/deployment/resolvers.ts
```diff
@@ -178,39 +187,45 @@ export const getDeploymentsByFilter: ResolverFn = async (
...
-      queryBuilder = queryBuilder.whereIn('environment.project', userProjectIds);
-  }
-
-  // collect builds for a specific date range
-  if (startDate) {
-    queryBuilder = queryBuilder.where('deployment.created', '>=', input.startDate);
+  if (environmentType && !limitPerEnvironment) {
+    // if environment type is set, assume only the latest deployment for the environment
+    // is requested depending on the `deploymentStatus` options chosen
+    limitPerEnvironment = 1
```
bomoko: as a general point of style, there are a lot of missing semicolons (style isn't to lean on ASI).

services/api/src/resources/deployment/sql.ts
```diff
@@ -152,4 +152,74 @@ export const Sql = {
...
+        }
+        if(deployTargets) {
+          queryBuilder = queryBuilder.whereIn('environment.openshift', deployTargets);
+        }
+        queryBuilder = queryBuilder.whereIn('deployment.status', deploymentStatus);
+        // if includeDeleted is false, exclude deleted environments in the results (default)
+        if (!includeDeleted) {
+          queryBuilder = queryBuilder.where('environment.deleted', '=', '0000-00-00 00:00:00');
+        }
+        queryBuilder = queryBuilder.whereRaw('environment.project IN (SELECT id FROM project)')
```
bomoko: If this is to only include environments where there is a project attached, it might make more sense to simply do an inner join here, rather than a subquery. Other than being slightly more idiomatic/clearer - it also points to the possibility of this being expanded in the future (by org, for instance, with a left join), without too much trouble.

bomoko (changes requested):
This is good - two small comments that I'm happy to take or leave.
I think adding the semicolons is probably important for consistency though (darn golang, teaching us about the joys of no semicolons ...)

services/api/src/resources/deployment/resolvers.ts
shreddedbacon: Yeah, I always forget the bloody semicolons :old-man-yells-at-js:

services/api/src/resources/deployment/sql.ts
shreddedbacon: Ah yeah, nah this is leftover from something else, should have been removed.

bomoko (approved):
:shipit:

Issue #3135, opened 2022-05-11
Tear out environment `hits` data from the Lagoon API


seanhamlin:
This appears to be a vestigial from a time when there was a billing system directly integrated into Lagoon. 

https://github.com/uselagoon/lagoon/blob/main/services/api/src/models/environment.ts#L180-L543

The hits data sends 2 queries to ElasticSearch (or OpenSearch) and then aggregates by hour, and attempts to fill in the gaps. This is a complex operation, and does not work at scale (I was running it over 449 environments using the Lagoon API). Confusing, when it does not work, the API just returns `0` hits, and no error messages. So the data is wrong, and you have no way of knowing if this is the case.

The API is scoped to platform administrators only, so impact will be limited.

## Proposal

Tear out this ElasticSearch querying in an API functionality from Lagoon. I would be a fan of Lagoon storing aggregated data, e.g. hits per day per environment, and a new Lagoon API that could surface that information. But that could come later.

tobybellwood:
Having tight lagoon<>OpenSearch bonds is undesirable, will progress this

PR #4039, opened 2025-12-05
refactor: remove bitbucket and open distro integrations

### LINKED ISSUE #2830: Outdated fields defined for index patterns
The Opendistro Security client hard-codes a selection of fields which are now outdated on non-openshift clusters.

In fact, I don't know if we need to define these at all since the fields should be generated automatically when the index pattern is created? I'm not sure about this though, it probably needs some experimentation.

https://github.com/uselagoon/lagoon/blob/a4c753370d84b51ba036d558bc083880883729dd/services/api/src/resources/group/opendistroSecurity.ts#L101

### LINKED ISSUE #2844: Improve Opensearch compatibility

rocketeerbkw:
# General Checklist

# Database Migrations

# Description

The Bitbucket and Open Distro integrations were deprecated in Lagoon 2.29.0. This PR removes them entirely.

If use of the Bitbucket sync is still desired, it should be simple to modify the removed script and run independently. Source code can be found at https://github.com/uselagoon/lagoon/tree/6c4ff3fb479ddba6b51b43e2c710692a1a333559/services/api/src/bitbucket-sync.

Open Distro integration can similarly be modified and run on-demand. Source code at https://github.com/uselagoon/lagoon/blob/6c4ff3fb479ddba6b51b43e2c710692a1a333559/services/api/src/helpers/sync-groups-opendistro-security.ts

# Closing issues

Closes #2830. Closes #2844.


services/api/src/models/environment.ts
```diff
@@ -264,7 +264,7 @@ export const Environment = (clients: {
           },
         },
       };
-      const legacyResult = await esClient.search(legacyQuery);
+      const legacyResult = await clients.esClient.search(legacyQuery);
```
shreddedbacon, line 267: We probably need to figure out if we deprecate the hits from the API, that would remove all this code that probably isn't used anymore by AIO, and probably other users aren't using it.

rocketeerbkw: https://github.com/uselagoon/lagoon/issues/3135

shreddedbacon: Can we remove this in a future release then? Can we announce its deprecation in this release?

rocketeerbkw: Not everyone is on board with removing `hits` entirely, but maybe it's not a direct Lagoon -> logs query as suggested in https://github.com/uselagoon/lagoon/issues/1696#issuecomment-948395733.

shreddedbacon (approved):

services/api/src/models/environment.ts
shreddedbacon: Ok, we can figure it out separately

PR #4040, opened 2025-12-09
feat: use env vars for all api-db/keycloak-db connection info

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

Makes sure all DB connection info for api-db/keycloak-db can be configured via env vars.

# Closing issues

n/a


rocketeerbkw:
Accidentally made a two way dependency between this and charts so neither test suite will pass. I can fix that.

shreddedbacon:
`ssh-portal-api` and `ssh-token` use `API_DB_USERNAME`, if not provided falls back to the [default](https://github.com/uselagoon/lagoon-ssh-portal/blob/dff1c7b623bc1775b06884f5a5277fe840e27cbe/cmd/ssh-portal-api/serve.go#L28C56-L28C71) `api` username , `api` uses `API_DB_USER`. 

With changes in the charts configured to use an external db with a username that isn't `api`, the ssh-portal and ssh-token services fail to start.

Few different options. Can either update the chart to set `API_DB_USERNAME` in addition to `API_DB_USER`, use `API_DB_USERNAME` in lagoon instead, or some other option?. I think it makes sense to use `API_DB_USERNAME` in lagoon though.

rocketeerbkw:
> `ssh-portal-api` and `ssh-token` use `API_DB_USERNAME`
> 
> Few different options. Can either update the chart to set `API_DB_USERNAME` in addition to `API_DB_USER`, use `API_DB_USERNAME` in lagoon instead, or some other option?. I think it makes sense to use `API_DB_USERNAME` in lagoon though.

Ah, good catch. I used `API_DB_USERNAME` everywhere at first but it broke `api` and I forgot to check the other services. I think changing lagoon-core-api is my preferred option? Since it hasn't been set before it won't be any kind of bc issue and standardizes the env var name.

shreddedbacon:
> I think changing lagoon-core-api is my preferred option? Since it hasn't been set before it won't be any kind of bc issue and standardizes the env var name.

Yeah same

rocketeerbkw:
Tests passing with https://github.com/uselagoon/lagoon-charts/pull/872

shreddedbacon:
Once the charts PR is merged, I'll revert the `CHARTS_TREEISH` override in this PR.

shreddedbacon (approved):

PR #4062, opened 2026-03-16
chore: reduce log verbosity for core ssh service

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

Every time an SSH key is authenticating to the core ssh service, a log like `User lagoon-demo-org-main authorized keys /dev/null is not a regular file` is printed. This PR just makes it so that goes away.

This reduces the amount of logs by ~16%.

# Closing issues

n/a


bomoko (approved):
Changes make sense.

PR #4067, opened 2026-03-30
Fixes expiry and iat logic for legacy tokens

bomoko:
As the expiry logic for the legacy token stands, it doesn't actually protect from someone having a _practically_ long lived token if they set the distance between `exp` and `iat` to be below the current tolerance BUT set them to be arbitrarily far in the future.

Currently we only test 
1. that the time between `exp` and `iat` meets the tolerance
2. that the token hasn't expired (taken care of by the library validation)

What this introduces is, within some tolerance, that the `iat` is before the present moment

### DIFF
--- a/services/api/src/util/auth.ts
+++ b/services/api/src/util/auth.ts
@@ -127,7 +127,7 @@ export const getCredentialsForLegacyToken = async (token: string): Promise<Legac
 
     // First check if the iat is _before_ now, with some tolerance for clock drift
     if (iat > (nowSeconds + clockSkew)) {
-      const msg = `Legacy token (sub:${sub}; iss:${iss}) iat ${(iat)} is issued before ${nowDate}`;
+      const msg = `Legacy token (sub:${sub}; iss:${iss}) iat ${(iat)} is issued before ${nowSeconds}`;
       throw new Error(msg);
     }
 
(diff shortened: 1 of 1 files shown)


services/api/src/util/auth.ts
```diff
@@ -119,7 +119,18 @@ export const getCredentialsForLegacyToken = async (token: string): Promise<Legac
...
   const rejectLegacyExpiry = getConfigFromEnv('LEGACY_EXPIRY_REJECT', 'false') // don't reject intially, just log
+  // Give some tolerance to the calculations around `iat` below
+  const clockSkew = parseInt(getConfigFromEnv('LEGACY_TOKEN_CLOCK_SKEW_TOLERANCE_SECONDS', '60'));
+  const nowDate = Date.now();
+  const nowSeconds = Math.floor(nowDate / 1000);
   if (exp && iat) {
+
+    // First check if the iat is _before_ now, with some tolerance for clock drift
+    if (iat > (nowSeconds + clockSkew)) {
+      const msg = `Legacy token (sub:${sub}; iss:${iss}) iat ${(iat)} is issued before ${nowDate}`;
```
shreddedbacon, line 130:
```suggestion
      const msg = `Legacy token (sub:${sub}; iss:${iss}) iat ${(iat)} is issued before ${nowSeconds}`;
```

Just to make the times shown both be the seconds representation

shreddedbacon (changes requested):
Just a small one. When checking this, the time for `now` was showing milliseconds in the error, this differs from the time for `iat` being shown in seconds.

shreddedbacon (approved):

PR #4072, opened 2026-04-14
feat: support for project cloning

shreddedbacon:
# General Checklist

# Database Migrations

# Description

This feature adds support for cloning a project and environment to a new project. This feature is only available to projects in an organization. Currently only organization admin and owner have the permission to clone projects, similar to how they only have the permission to add projects.

The cloning process optionally supports copying the data (ie, files, databases), group associations, notifications, and project and environment variables.

Supporting chart changes in https://github.com/uselagoon/lagoon-charts/pull/902


services/actions-handler/handler/controller_builds.go
```diff
@@ -132,6 +133,69 @@ func (m *Messenger) handleBuild(ctx context.Context, messageQueue *mq.MessageQue
...
+								deployments{
+									name
+								}
+							}
+						}
+					}
+				}`, message.Meta.Project)
+
+			projectData, err := l.ProcessRaw(ctx, projectRaw, nil)
+			if err != nil {
```
bomoko: should we not be returning nil here?

services/actions-handler/handler/controller_builds.go
```diff
@@ -132,6 +133,69 @@ func (m *Messenger) handleBuild(ctx context.Context, messageQueue *mq.MessageQue
...
+			var project cloneData
+			data, _ := json.Marshal(projectData.(map[string]interface{})["projectByName"])
+			if err := json.Unmarshal(data, &project); err != nil {
+				log.Printf("%sERROR: unable to unmarshal clone data: %v", prefix, err)
+				return nil
+			}
+
+			cloneStatus := ""
+			// handle all clone status', not just complete
+			switch buildStatus {
```
bomoko: Does this mean that in all other cases we're going to just write an empty status?

services/api/src/clients/pubSub.ts
```diff
@@ -64,7 +70,103 @@ const createSubscribe = (events): ResolverFn => async (
...
+          sqlClientPool,
+          orgSql.selectOrganizationByName(orgName),
+        );
+
+        const o = path([0, 'id'], rows);
+        if (!o) {
+          throw new ForbiddenError('Organization not found.');
+        }
+
+        logger.info(`AAAA2 ${JSON.stringify(rows[0])} ${o}`)
```
bomoko: rogue debugging statement?

services/api/src/resources/deployment/sql.ts
```diff
@@ -115,6 +115,13 @@ export const Sql = {
     knex('deployment')
       .where('environment', '=', environment)
       .toString(),
+  selectLatestDeploymentForEnvironment: (environment: number) =>
+    knex('deployment')
+      .where('environment', '=', environment)
+      .orderBy('created', 'desc')
+      .orderBy('id', 'desc')
```
bomoko, line 122: interesting, is there any case where the last id won't be the last created? Does this have any implications for indexing?

services/api/src/resources/project/helpers.ts
```diff
@@ -121,6 +121,66 @@ export const Helpers = (sqlClientPool: Pool) => {
...
+  const hasProjectRestriction = async (restriction: string, projectId: number, adminScopes) => {
+    if (adminScopes.platformOwner) {
+      return
+    }
+    const project = await getProjectById(projectId);
+    if (project.restrictions) {
+      const r = JSON.parse(project.restrictions)
+      if (r.includes(restriction)) {
+        switch (restriction) {
+          case 'no_tasks':
```
bomoko, line 169: Are we happy having these as string literals?

shreddedbacon: i can make them types if it makes more sense

services/api/src/resources/deployment/sql.ts
shreddedbacon: I'll double check what is calling this

services/api/src/clients/pubSub.ts
shreddedbacon: hahah this has me written all over it. Will :nuke: it

services/api/src/resources/deployment/sql.ts
shreddedbacon: Yeah, this is called by the `cloneProject` resolver to verify that the environment is about to copy, that its latest deployment is `completed` to allow the clone process to proceed https://github.com/uselagoon/lagoon/blob/313743d846ed43434dd8f78fb9529e08c2b8e9e2/services/api/src/resources/project/resolvers.ts#L1232-L1236

services/actions-handler/handler/controller_builds.go
shreddedbacon: Ah yeah, this seems a bit funky logic wise when you look at it. However, the code is only hit in a switch on line 126 that only checks for these 3 statuses. So it doesn't actually push an empty status anywhere.
```
	switch buildStatus {
	case "complete", "failed", "cancelled":
```

Could probably do with some refactoring to be like this then?
```
			var cloneStatus string
			// handle all clone status', not just complete
			switch buildStatus { 
			case "complete":
				cloneStatus = "COMPLETE"
				if len(project.Clone.DestinationProject.Deployments) == 1 {
					cloneStatus = "FIRST_DEPLOYMENT_COMPLETE"
				}
			}
			default:
				cloneStatus = strings.ToUpper(buildStatus)
```

services/actions-handler/handler/controller_builds.go
shreddedbacon: i dont think returning nil here should be done, but it should definitely drop out of this particular if statement, as there is code below this function that still needs to be called, even if this function errors.

https://github.com/uselagoon/lagoon/pull/4072/changes/BASE..313743d846ed43434dd8f78fb9529e08c2b8e9e2#diff-aa7b4cea272bc79f24c69914b18417d59087075d9103a281cdf7b39ed620726aR170 there is a return nil here that probably will block remaining code execution too which could be problematic

shreddedbacon:
Gah, the jenkins run is failing at the arm image build+push phase, otherwise all tests are passing.

services/api/src/resources/project/helpers.ts
bomoko: I think this is a refactor we can do at some stage. Obviously consts or types mean stronger support for automated tooling around these things. But for now this suffices, I think.

services/api/src/resources/deployment/sql.ts
bomoko: cool, we can resolve this - was just flagging a potential optimization.

services/actions-handler/handler/controller_builds.go
bomoko: okay, sweet! Resolving

services/actions-handler/handler/controller_builds.go
bomoko: okay - makes sense.

bomoko (approved):

shreddedbacon:
Since the tests are failing on the arm image build phase, this is a known pain point in the jenkins pipeline. The remaining phases after that failure don't run for pull requests anyway, so this is fine to merge.

PR #4076, opened 2026-04-29
Improve mkcert fallback handling in Makefile

tobybellwood:
Enhance the handling of mkcert in the Makefile to ensure a fallback mechanism is in place when mkcert is not available. This change addresses potential issues with certificate generation by checking for mkcert's existence before attempting to use it.


rocketeerbkw (approved):
Works as described.

PR #4091, opened 2026-05-22
Change: Adds project clone permissions

CGoodwin90:
Adds `cloneProject:add` & `cloneProject:cancel` permissions for project cloning + `cancelProjectClone` resolver.


services/api/src/resources/project/resolvers.ts
```diff
@@ -1673,20 +1673,165 @@ export const cancelProjectClone: ResolverFn = async (
...
+      cloneId,
+      organization: organizationId
     }
   },
-  { sqlClientPool, hasPermission, userActivityLogger, models, adminScopes }
+  { sqlClientPool, hasPermission, userActivityLogger, models, adminScopes, keycloakGrant, legacyGrant }
 ) => {
-  // delete cloned environment
+  await hasPermission('organization', 'cloneProject:cancel', {
+    organization: organizationId
```
rocketeerbkw: This `organizationId` is user supplied and can't be trusted. A user can submit their own org instead of the org of the clone. Since there is no verification that the passed org is the same as the org of the clone, this permissions evaluates that "any user which has `cloneProject:cancel` for any organization is allowed to cancel project clones for any other organization."

This can be fixed by checking the org on the source/destination project instead of getting it from user input.

rocketeerbkw (changes requested):

rocketeerbkw (approved):

PR #4105, opened 2026-06-11
fix: race condition when making multiple keycloak API calls

### LINKED ISSUE #3907: Querying for group members sometimes fails
# Describe the bug

Sometimes querying the API to list group members will throw an error:
```
{
  "errors": [
    {
      "message": "keycloakUsers is not iterable",
```

## To Reproduce

1. Create a group with no members
2. Run the query:
    ```
    query groupByName {
      groupByName(name:"memberless-group"){
        id
        name
        memberCount
        members {
          user {
            email
          }
          role
        }
      }
    }
    ```

4. It should fail all the time
5. Comment out _either_ the `memberCount` OR `members` field, and the query no longer fails
     - Here's where it gets weird 

The error also sometimes triggers right after adding a member to a group, then running the query in step 2, but it's difficult to reproduce.

There is some kind of race condition (?) where the call to `keycloakAdminClient.groups.listMembers` in `services/api/src/models/groups.ts` returns `null` instead of an empty array.

## Expected behavior

Listing group members doesn't throw errors.

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

A shared keycloak client is used in the Lagoon API. A race condition exists where an API request can fail or return unexpected data if it is done in the middle of another API request getting an access token.

The fix is to use a new/separate keycloak client for access tokens.

The easiest way to reproduce is to run the query described in #3907, but [disable the group member cache](https://github.com/uselagoon/lagoon/blob/main/services/api/src/resources/group/resolvers.ts#L153) so it is 100% reproducible:

```suggestion
const members = await models.GroupModel.getGroupMembership(group, false);
```

# Closing issues

closes #3907

### DIFF
--- a/services/api/src/clients/keycloak-admin.ts
+++ b/services/api/src/clients/keycloak-admin.ts
@@ -29,6 +29,11 @@ export const getKeycloakAdminClient = async (): Promise<KeycloakAdminClient> => 
     realmName: config.realm,
   });
 
+  const tokenProviderClient = new KeycloakAdminClient({
+    baseUrl: `${config.origin}/auth`,
+    realmName: 'master', // auth against the `master` realm
+  });
+
   /**
    * Use a custom token provider that can automatically refresh expired tokens.
    */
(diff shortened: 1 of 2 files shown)


bomoko (approved):
Yes, this makes sense. While it's weird having a client carrying around another client in its environment, this seems to be a clean way of solving the issue.

PR #4107, opened 2026-06-17
fix: multiple issues with local-api-data-watcher-pusher

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

Fixes a number of issues discovered in `local-api-data-watcher-pusher`:

> mcli: <ERROR> `config` is not a recognized command. Get help using `--help` flag.

There is no `config` command for `mcli`. Maybe it was removed in some version, but all the `mcli cp` calls have been failing silently for awhile. The files were being written to a folder called `local-minio` instead of the minio service. Fixed by using correct `alias` command.

> $ git worktree remove test-mcli-error
> warning: could not open directory 'local-dev/api-data-watcher-pusher/.mcli/': Permission denied
> error: failed to delete 'test-mcli-error': Directory not empty

This is an annoyance for local development. Running `mcli alias set` creates a `/home/.mcli` folder with root permisisons. Since docker-compose mounts `/home` to the host it requires `sudo` to clean up the folder on the host. Fix is to run and mount scripts out of `/app` instead of `/home`.

> [WARN  tini (7)] Tini is not running as PID 1 and isn't registered as a child subreaper.
> Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
> To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.

Just a warning, but easy to fix by running `tini` as the entrypoint only, not in the command.

> "errors":[{"message":"Error: Provided harbor configuration is not valid: Error: all branches, excluding pullrequests: latestPushed must be a number"

There is still one error left, will address in #4108.

# Closing issues

n/a

### DIFF
--- a/docker-compose.local-dev.yaml
+++ b/docker-compose.local-dev.yaml
@@ -45,4 +45,4 @@ services:
       - "./local-dev/keycloak:/lagoon/keycloak"
   local-api-data-watcher-pusher:
     volumes:
-      - ./local-dev/api-data-watcher-pusher:/home
+      - ./local-dev/api-data-watcher-pusher:/app
--- a/local-dev/api-data-watcher-pusher/Dockerfile
+++ b/local-dev/api-data-watcher-pusher/Dockerfile
@@ -1,5 +1,8 @@
 FROM ${UPSTREAM_REPO:-uselagoon}/commons:${UPSTREAM_TAG:-latest}
 
+RUN mkdir -p /app \
+    && fix-permissions /app
+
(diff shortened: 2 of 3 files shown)


bomoko (approved):

shreddedbacon:
Running a local-stack after these changes, where you may have had one before with an existing `local-dev/api-data-watcher-pusher/.mcli` directory can result in a local-stack startup failure unless you delete that directory.

PR #4109, opened 2026-06-18
Change: Restricts cloning for projects with deployTarget Configuration

CGoodwin90:
Attempting to clone a project with deployTarget Configurations fails, this adds a simple check to not allow cloning of these projects. We may look at adding support for cloning projects with deployTarget Configurations in the future.


rocketeerbkw (approved):
Makes sense.

PR #4129, opened 2026-07-20
feat: log api-db/keycloak-db slow queries to stdout

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

The default for Lagoon db images is to log queries to `/var/log` which is not persistent and does not prevent endless growth.

This PR configures the slow query log (when enabled) to print to `stdout` so that they can be ingested by normal container log methods.

# Closing issues

n/a

### DIFF
--- a/services/api-db/slow-query-stdout-entrypoint.bash
+++ b/services/api-db/slow-query-stdout-entrypoint.bash
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+if [ -n "$MARIADB_LOG_SLOW" ]; then
+  # Log slow queries to stdout
+  ln -sf /proc/1/fd/1 /var/log/mariadb-slow.log
+fi
--- a/services/api-db/Dockerfile
+++ b/services/api-db/Dockerfile
@@ -7,8 +7,11 @@ ENV LAGOON_VERSION=$LAGOON_VERSION
 
 USER root
 
+RUN /bin/fix-permissions /var/log/
+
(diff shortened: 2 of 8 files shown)


services/api-db/slow-query-stdout-entrypoint.bash
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+if [ -n "$MARIADB_LOG_SLOW" ]; then
```
shreddedbacon, line 5: Rather than have a `MARIADB_` and `MYSQL_` prefix on the variable, why not use `APIDB_SLOW_LOG` and `KEYCLOAKDB_SLOW_LOG`

rocketeerbkw: These are the same vars that turn on slow logging in the base images

https://github.com/uselagoon/lagoon-images/blob/e2a9611d12d190461d176523a272c7556a1387ae/images/mariadb/entrypoints/100-mariadb-logging.bash#L5

shreddedbacon: Yeah, my only thing would be there is no way to turn it on for one or the other without doing some crafty chart changes with supporting logic for whether the chart is installing with mysql/mariadb and then keycloakdb/apidb

at least with isolation the chart logic is a bit simpler and the db type wouldn't matter.

shreddedbacon: I just saw where you were going to use this, directly with env vars in the values. I guess it works in that case.

PR #3979, opened 2025-09-08
feat/refactor: rewrite webhook-handler

shreddedbacon:
# General Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for inclusion in changelog

# Database Migrations

- [ ] If your PR contains a database migation, it **MUST** be the latest in date order alphabetically

# Description

This is a substantial rewrite of the `webhook-handler` which consolidates the `webhook-handler` and `webhooks2tasks` services. Additionally, the API will now call out to the `api-sidecar-handler` service to handle triggering deployments and environment removals, this allows for shared code between the `webhook-handler` and `api-sidecar-handler`.

One way this is done is by moving from individual go services with their own go.mod files, into a mono repo setup. There are some remaining services that will need to be converted in a follow up pullrequest though. I left them out of this to try to reduce the amount of changes.

Builds off fork of [drone/go-scm](https://github.com/amazeeio/go-scm/tree/pr-commit-hashes), which we can try to get merged into upstream.

Requires https://github.com/uselagoon/lagoon-charts/pull/858

PR #4134, opened 2026-07-23
fix: send notifications for webhooks like previously

shreddedbacon:
# General Checklist

# Database Migrations

# Description

When #3979 was introduced, some notifications that would be sent when webhooks were received were missed. This adds them back, and adds missing checks for new supported git provider types to existing notifications.

Additionally, this removes an image that was being set in Microsoft Teams notifications as this appears to not actually be supported the way it was implemented and results in this error when Teams processes it.
```
'Post_card_in_a_chat_or_channel_1' failed: Invalid image URI
```


bomoko (approved):

bomoko (approved):
same as before. All good.

PR #4140, opened 2026-07-29
refactor/fix: webhook 200s with fixed envtype checks

shreddedbacon:
# General Checklist

# Database Migrations

- [ ] If your PR contains a database migration, it **MUST** be the latest in date order alphabetically

# Description

Some git hosting providers will disable webhooks if they receive too many errors. This changes the error code to 200s for all webhooks.

This is done because some webhooks would previously return a 400 indicating a failure to deploy. Previous webhook-handler in Lagoon also responded with 200s for all webhooks, so this is returning to that previous behaviour.

Additionally fixes an environment type check when deploying


bomoko (approved):
Makes sense to support previous behaviour.

shreddedbacon:
Just rebased to bring inline with main

bomoko (approved):

PR #4141, opened 2026-07-30
Serve pages as Markdown and publish llms.txt

bonus414:
Publishes each docs page as plain Markdown alongside the HTML, plus an
llms.txt index, so AI tools can read the docs directly instead of scraping.

Two small changes: the plugin in docs/requirements.txt (unpinned, matching
the other entries) and enabled in mkdocs.yml.

Checked with a strict build both ways — the native build that publishes the
site, and the container check in CI. The Japanese site still builds and
needs no translation changes, since this adds nothing to the nav.

CI's container step will fail until the plugin lands in the mkdocs-material
image. The native build, which is what publishes, passes now.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -5,3 +5,4 @@ mkdocs-git-revision-date-localized-plugin
 mkdocs-awesome-pages-plugin
 mkdocs-markdownextradata-plugin
 mkdocs-static-i18n
+mkdocs-llms-source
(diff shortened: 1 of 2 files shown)


bomoko:
@bonus414 - changes build locally but the test isn't working (see above). 
See https://github.com/uselagoon/lagoon/actions/runs/30575260373/job/90981818563?pr=4141

bonus414:
That run predated the image fix — `mkdocs-llms-source` landed in the image via amazeeio/mkdocs-material#166 and a fresh `latest` published last night. I re-ran the job and it's green now, so this should be ready for another look.

bomoko (approved):
:shipit:

PR #4161, opened 2026-08-31
Chore: upgrade to keycloak 26.7.2

bomoko:
# General Checklist

# Database Migrations

- [ ] If your PR contains a database migration, it **MUST** be the latest in date order alphabetically

# Description

Standard keycloak upgrade


shreddedbacon (approved):
Spun up in `local-stack` and did some spot user testing using the different local-stack accounts with the varying roles.

The upgrade path works in CICD too, so no concern around upgrade from 26.6 to 26.7

I thought there was something in KC that was deprecated and due for removal in 26.7.0 with FGAP, but it doesn't appear to have been removed yet otherwise we probably would see some auth failures in local-stack when testing the various accounts.

Happy to ✔️ this. I wasn't able to spot anything immediately problematic

PR #4162, opened 2026-08-31
fix: updating a retention policy always applies it to all projects

rocketeerbkw:
# General Checklist

# Database Migrations

n/a

# Description

Updating a retention policy always calls `postRetentionPolicyUpdateHook` which treats the policy like a global one. This results in the policy being applied to all projects, regardless of it's links or overrides.

The fix is to first check if it is linked at all, and to use the correct post update hook for global vs org/project scoped links.

# Closing issues

n/a


shreddedbacon (approved):
Ran locally and can see that updating a policy defined against a project does not try to update as if it was a global policy.

